PR #150, opened 2014-05-17
Move to Mocha

arthurschreiber:
As promised a long time ago, I started to work on converting all of Mochas Unit and Integration tests from `nodeunit` over to `mocha`.

Here's what I have so far, but it's still pretty much WIP, but I think this nicely show cases the benefits of the move.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
   "devDependencies": {
     "async": "0.2.10",
     "coffee-script": "1.7.1",
-    "nodeunit": "0.8.4"
+    "nodeunit": "0.8.4",
+    "mocha": "^1.18.2"
   },
   "scripts": {
     "test": "scripts/tests-unit",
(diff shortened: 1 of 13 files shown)


arthurschreiber:
@bretcope Is there still interest in this?

bretcope:
I think moving to mocha would be a good thing since it's more commonly used and understood by node developers. Whether or not it happens depends entirely on whether you want to do it. No one will complain about making the switch... but no one is likely to help either.

Hadis-Fard:
@arthurschreiber what's the roadmap on this one? what are your thoughts about writing  new tests in Mocha for any new changes and converting the nodeunit tests in batches over the time? meanwhile we can run both in parallel, and when we have moved to Mocha completely we delete the nodeunit from CI.

arthurschreiber:
Closing this in favor of #746.

PR #182, opened 2014-09-08
add support for fReadOnlyIntent and automatic re-routing

rossipedia:
add support for fReadOnlyIntent and automatic re-routing

part of #183


closedstack:
I couldn't find any doc on how I can change connection string to use MultiSubnetFailover and ReadOnlyIntent. Would appreciate any assistance

bretcope:
@closedstack it's just a connection option. http://pekim.github.io/tedious/api-connection.html Look for "readOnlyIntent".

Issue #223, opened 2014-12-29
Redirect fails on different server name with Login7Packet


christopheranderson:
TDS supports interpretting the server name from the username string ("me@servername.net") but when you redirect a TDS connection, on the reconnect, it can fail to login using this method if the server name provided in the username string doesn't match the server name you were redirected to. The means I found to fix this was to pass the server name along in the Login7Packet. 

i.e.
```coffee
sendLogin7Packet: ->
    loginData =
      domain: @config.domain
      userName: @config.userName
      password: @config.password
      database: @config.options.database
      serverName: @config.options.server
      appName: @config.options.appName
      packetSize: @config.options.packetSize
      tdsVersion: @config.options.tdsVersion
      initDbFatal: not @config.options.fallbackToDefaultDb
      readOnlyIntent: @config.options.readOnlyIntent
```

This worked because serverName was already passed along in login7-payload.coffee, just not initialized in connection.coffee.

This bug is related to #222, preventing redirects in some scenarios.

sandorfr:
Similar issue here.

It appears that sql azure (v12) now uses routing to redirect the connection to the appropriate server. 
The first authentication is doing ok but then after the routing occurs it complains about not having the servername. 

In fact when it is redirected it appears it should use the redirected servername instead of the original one. 

In addition to your workaround I had to edit reconnect from Rerouting to change the server so it matches the redirection data.
```
this.config.options.server = this.routingData.server 
```

This does not look clean but it works/

christopheranderson:
Correct, this issue will be an issue for most people using SQL Azure v12 from within Azure as they redirect all Azure connections in this manner.

christopheranderson:
@sandorfr, you can give my fork a try. Should work for you.

My original suggested fix had a typo in, but I've validated my current fix and I believe it isn't too hacky. :)

sandorfr:
Hi,

I tested it and works well for both remote and internal sql azure v12 connections.

Thanks :)

christopheranderson:
Wonderful. Glad to hear it!

bretcope:
Fixed via https://github.com/pekim/tedious/pull/224

georgievrado:
We are experiencing the same issue. We've upgraded to latest mssql module which comes with tedious 1.9.0. We've been only able to make this work if we set the server and port to match the one we get redirected to.

This does not work:
``` "db": {
        "user": "user@azure.server",
        "password": "password",
        "server": "azure.server",
        "database": "db",
[...]

Issue #222, opened 2014-12-29
TLS Negotiation bug causes redirected connections to fail on redirect


christopheranderson:
When you are redirected on your initial connection, if you've already negotiated SSL once, it won't wait for SSL negotiation to complete before trying to process (in my case) the Login7 message, where it will crash. I fixed this on my local driver by resetting the tlsNegotiationComplete state - see below:
```coffee
message: ->
          if  @tlsNegotiationComplete
            @tlsNegotiationComplete = false
            @transitionTo(@STATE.SENT_LOGIN7_WITH_STANDARD_LOGIN)
          else
```

This may not be the best way for this to work. You may want to reset this when starting the TLS Negotiation, for instance:

```coffee
enter: ->
          @tlsNegotiationComplete = false
```

FYI - There is another issue #223 related to redirect failing when there is a different server url you're redirected to.

sandorfr:
I believe I'm facing the same issue as you do with the latest sql azure version. After the redirection occurs it tries to login again and fails with message "Unrecognized token 20 at offset 1".

sandorfr:
I tried your second fix and it appears to fix this issue. I am now struggling with a server name issue...

christopheranderson:
Confirmed, "Unrecognized token 20 at offset 1" is the error I saw being created for this bug. Should have included that in my original message. :) #223 addresses the server name issue (with a suggested, tentatively working on SQL Azure v12 fix).

bretcope:
Can someone make a pull request to address these issues?

christopheranderson:
Roger. I'll give it a shot tonight/tomorrow.

bretcope:
Fixed via https://github.com/pekim/tedious/pull/224

PR #224, opened 2015-01-07
Changes To Fix Rerouting on TLS Encrypted Connections  

christopheranderson:
Needed to make the following changes to support (in connection.coffee unless otherwise stated)
- New event needed for rerouting called as such "rerouting"
- Replaced previous "end" event on rerouting with aforementioned new event, "rerouting"
- Initialize tlsNegotiationComplete as false on start of SENT_TLSSSLNEGOTIATION
- Add server to serverName property in Login7Packet in the event the servername differs from the redirected hostname address (eg connect to serverA.net -> redirected to serverB.net -> server at serverB.net expects serverA's name for the server)
- Fixed up the integration/connection-test.coffee test to account for redirection on an encrypted server requiring 2 separate "encrypt" events, resulting in 3 additional successes in addition to the original expectation of 5 success

Changes not included in this commit
- Need to update the Connection api webpage to document the new rerouting event

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -74,7 +74,7 @@ class Connection extends EventEmitter
     REROUTING:
       name: 'ReRouting'
       enter: ->
-        @cleanupConnection()
+        @cleanupConnection(true)
       events:
         message: ->
         socketError: (error) ->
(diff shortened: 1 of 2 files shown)


christopheranderson:
FYI - Changes related to issues #222 and #223

bretcope:
Published as `1.8.1`. Thank you.

PR #227, opened 2015-01-13
Fix VarBinary handling with strings.

arthurschreiber:
When storing a String into a VarBinary column, tedious currently incorrectly encodes that string as UCS-2. This is a bit weird, because that means that

```
request.addParameter('test', TYPES.VarBinary, "Some String")
```

and

```
request.addParameter('test', TYPES.VarBinary, new Buffer("Some String"))
```

will cause different binary data to be stored in the database.

---

I'm not really sure whether that is a bug or I'm just doing something wrong.

To clarify: I'd expect a VarBinary column to store exactly what I throw at it, and in case of a String I'd expect it to store the same binary representation I'd get when calling `new Buffer("Some String")`. But that's not what is happening. By default, Tedious currently encodes this into `UCS-2`, which I find super weird and super awkward to work with.

Another Note: This is probably a breaking change.

### DIFF
--- a/src/tracking-buffer/writable-tracking-buffer.coffee
+++ b/src/tracking-buffer/writable-tracking-buffer.coffee
@@ -175,7 +175,7 @@ class WritableTrackingBuffer
     @writeUInt16LE(value.length)
     @writeString(value, encoding)
   
-  writeUsVarbyte: (value, encoding = @encoding) ->
+  writeUsVarbyte: (value, encoding = "utf8") ->
     if Buffer.isBuffer value
       length = value.length
     else
--- a/test/integration/parameterised-statements-test.coffee
+++ b/test/integration/parameterised-statements-test.coffee
@@ -100,6 +100,12 @@ exports.intZero = (test) ->
 exports.intNull = (test) ->
   execSql(test, TYPES.Int, null)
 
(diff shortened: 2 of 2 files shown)


bretcope:
I don't think I agree with this change.

1. It's a breaking change, as you say, and I don't see a compelling reason to introduce it.
2. Encoding as UCS2 should be faster since that's what JavaScript strings are to begin with.
3. UCS2 is what SQL Server uses for unicode support with nchar and nvarchar, so people should be familiar with it when working with SQL Server.
4. It's _really_ subjective what the right behavior should be to begin with.

If anything, what I think we should do is deprecate strings support for binary types. Output a message to the console saying something like:

> Using a string value for binary parameter types is deprecated. This functionality will be removed in the next major release. Please use a Buffer value instead.

arthurschreiber:
@bretcope That's probably a better idea. Users then have to decide on their own what encoding they want to use for their Buffers, and it'll remove all the confusion.

Sweet. Is there any objection against using `depd` (https://github.com/dougwilson/nodejs-depd) for outputting deprecation notices?

bretcope:
That's fine with me.

arthurschreiber:
Closing this as it's totally out of date. 😄

Issue #226, opened 2015-01-12
Cannot easily parse error number from error object on sql request


ashelley:
Hello,

I'm looking for a way to programatically deal with database errors.  For instance I would like to easily detect when a unique key violation occurs.  

Currently on this line:

https://github.com/pekim/tedious/blob/master/src/connection.coffee#L312

```
@request.error = RequestError token.message, 'EREQUEST'
```

we seem to lose token.number (in this case it would be 2627);

if the code was like this:

```
@request.error = RequestError token.message, 'EREQUEST'
@request.error.number = token.number;
```

it would offer the ability to detect this error more easily in code instead of having to regex the error message.

It could be that there is already a good way to deal with this by adding the event listener to to errorMessage on the connection and doing this:

```
connection.on('errorMessage', function(token) {
	connection.request.error.token = token;
});
```

however because the errorMessage event is emitted before setting the error here:

https://github.com/pekim/tedious/blob/master/src/connection.coffee#L309

```
    @tokenStreamParser.on('errorMessage', (token) =>
      @emit('errorMessage', token)
      if @loggedIn
        if @request
          @request.error = RequestError token.message, 'EREQUEST'
      else
        @loginError = ConnectionError token.message, 'ELOGIN'
    )
```

The above strategy of using the eventListenner on the connection to preserve the actual error code can't work.

Is there a good way for me to implement using the actual token.number for error handling in this driver?

Thanks,

-Adam

bretcope:
I think you're right. Why don't you put together a pull request and re-submit.

ashelley:
Sounds good.  I will look at this closer and see if I can create a comprehensive patch to look at for discussion.  Thank you for the feedback.

PR #233, opened 2015-01-20
Add error.number property and some simple tests for its correctness.

lee-houghton:
I've had a stab at a possible patch for #226 based on ashelley's fix, with some tests, since it's very relevant to my interests.

I wasn't sure about coding conventions - I don't usually add parentheses around function calls in coffeescript, especially when one of them is a callback, but not all the code seemed to agree on whether to do so or not.

Critique gratefully accepted!

### DIFF
--- a/test/integration/errors-test.coffee
+++ b/test/integration/errors-test.coffee
@@ -26,7 +26,7 @@ exports.uniqueConstraint = (test) ->
   drop table #testUnique;
   """
 
-  test.expect(2)
+  test.expect(3)
   execSql test, sql, (err) ->
     test.ok(err instanceof Error)
     test.strictEqual(err.number, 2627)
@@ -38,7 +38,7 @@ exports.ansiNullDefaults = (test) ->
   drop table #testAnsiNullDefault;
   """
 
-  test.expect(2)
+  test.expect(3)
   execSql test, sql, (err) ->
     test.ok(err instanceof Error)
     test.strictEqual(err.number, 515)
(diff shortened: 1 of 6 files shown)


arthurschreiber:
Actually, can you extend this to copy down all "interesting" properties from the error message token into the error object?

Having `state`, `class`, `serverName`, `procName`, `lineNumber` also available might be very helpful.

lee-houghton:
I've updated my patch to include the extra properties you mentioned. I wasn't sure how best to test them, but figured that `procName` and `lineNumber` could be tested via executing a temporary stored procedure that always fails, but error `state` values seem to refer to internal SQL Server code that might change between versions so I didn't test for any particular values. I wasn't sure if default severity levels are guaranteed either.

test/integration/errors-test.coffee
```diff
@@ -0,0 +1,113 @@
...
+
+execSql = (test, sql, requestCallback) ->
+  connection = new Connection(config)
+
+  request = new Request sql, ->
+    requestCallback.apply this, arguments
+    connection.close()
+
+  connection.on 'connect', (err) ->
+    if (err)
```
arthurschreiber: You should use test.ifErr instead of logging to the console.

lee-houghton:
Good point, thanks, I didn't see that. I based these tests off one of the other integration tests, so I've changed/added `test.ifError` to the connect handler in those too. I also realised that I actually could quite easily test `err.state` and `err.class` by simply using `raiserror` instead of generating a divide by zero error.

arthurschreiber:
LGTM! :+1: I'm on a trip till the end of next week, so I won't be able to get this merged until then. I'll get this merged once I'm back.=

arthurschreiber:
Perfect. All my :heart: goes to you for being so patient.

Issue #243, opened 2015-02-12
Crypto Deprecation Warning with Node v0.11.14+


borwahs:
As of Node v.0.11.14, it looks like the `crypto.createCredentials()` call was deprecated in favor of `tls`.

Testing with a simple connection, a warning is thrown:

    $ node test.js
       createCredentials() is deprecated, use tls.createSecureContext instead
       [ { '' : 1 } ]

Here is the code I used to create the warning. I was using [node-mssql](https://github.com/patriksimek/node-mssql) to test.

    var sql = require('mssql');

    var config = {
        driver: 'tedious',
        user: 'test',
        password: 'test',
        server: 'test',
        database: 'test',

        options: {
            encrypt: true
        }
    }

    var connection = new sql.Connection(config, function(err) {
        var request = new sql.Request(connection);
        request.query('select 1 as number', function(err, recordset) {
            console.dir(recordset);
        });
    });

The `crypto` function call is found in the `connection` class under the `initiateTlsSslHandshake` function.

Link to line of code: https://github.com/pekim/tedious/blob/85cac9f5c130d7a2876571fbdeb74a5b60257805/src/connection.coffee#L651

The `tls.createSecureContext` function wasn't added until Node v0.11.14:

* http://nodejs.org/docs/v0.11.13/api/tls.html#tls_tls_createsecurecontext_details
* http://nodejs.org/docs/v0.11.14/api/tls.html#tls_tls_createsecurecontext_details

Note the documentation changes for `crypto.createCredentials()`:

* Node v0.11.13: http://nodejs.org/docs/v0.11.13/api/crypto.html#crypto_crypto_createcredentials_details
* Node v0.11.14: http://nodejs.org/docs/v0.11.14/api/crypto.html#crypto_crypto_createcredentials_details
* Node v0.12.0: http://nodejs.org/api/crypto.html#crypto_crypto_createcredentials_details

PR #244, opened 2015-02-12
Fix deprecation warning for #243.

borwahs:
This addresses the issue described in #243.

It uses the `tls.createSecureContext` function instead of `crypto.createCredentials`.

Note - I wasn't sure how deprecations should be handled. The `tls.createSecureContext` function wasn't added until Node v0.11.14:

* http://nodejs.org/docs/v0.11.13/api/tls.html#tls_tls_createsecurecontext_details
* http://nodejs.org/docs/v0.11.14/api/tls.html#tls_tls_createsecurecontext_details

My first pull request, so let me know if I missed anything.

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -648,7 +648,10 @@ class Connection extends EventEmitter
       payload.toString '  '
 
   initiateTlsSslHandshake: ->
-    credentials = tls.createSecureContext(@config.options.cryptoCredentialsDetails)
+    credentials = if tls.createSecureContext
+      tls.createSecureContext(@config.options.cryptoCredentialsDetails)
+    else
+      crypto.createCredentials(@config.options.cryptoCredentialsDetails)
     @securePair = tls.createSecurePair(credentials)
 
     @securePair.on('secure', =>
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -648,7 +648,7 @@ class Connection extends EventEmitter
       payload.toString '  '
 
   initiateTlsSslHandshake: ->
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Hey @borwahs!

Thanks for the PR. As you can see from the appveyor build results, this does break compatibility with Node 0.10 (there's another flaky test that sometimes fails, but let's ignore that for now).

The easiest fix to support 0.10 as well as 0.11+ would be to change this to:

```js
credentials = if tls.createSecureContext
  tls.createSecureContext(@config.options.cryptoCredentialsDetails)
else
  crypto.createCredentials(@config.options.cryptoCredentialsDetails)
```

borwahs:
Hey @arthurschreiber!

Thanks for the feedback and direction on how to support the different Node versions. I didn't think to check if the object exists by checking for null.

I committed the change you suggested. 

Thanks!!

arthurschreiber:
Cool, thank you very much. I'll get that merged once the build has finished.

borwahs:
@arthurschreiber I was watching the build and just curious if you were referring to the `datatypes-in-results-test` that is flaky and fails?

I found the error in build 32 for v0.10 where I broke it.

arthurschreiber:
@borwahs You're right, that is the test that is flaky. It tries to send a large query to sqlserver, and because of the slow speed of free AppVeyor VMs, it sometimes fails.

There is an actual performance problem inside tedious, and I'm sure that test will start being less flaky once I have converted tedious internals to use streams.

borwahs:
@arthurschreiber Awesome. Thanks for your help with the code and logistics!

Issue #202, opened 2014-10-08
Integrated type checks


patriksimek:
There are situations when you accidentally pass a value that doesn't match specified data type, for example when you pass `999` to `TinyInt`. There are no type chcecks in Tedious so the application throws an `Out of bounds` error in Tedious core which also cause application to crash. Because of that, I suggest to add type checks.

This idea is based on https://github.com/patriksimek/node-mssql/issues/84 but I belive this feature should be implemented right in the Tedious core, not mssql.

Before I start working on this, I would like to discuss the feature design. There are two ways of how this can be done.

### Synchronous errors

```javascript
try {
	request.addParameter("name", TYPES.TinyInt, 999);
} catch (err) {
	err instanceof TypeError
	err.code === "ETYPE"
	err.message === "Value does not match the type of parameter 'name'."
}
```

### Asynchronous errors

```javascript
request = new Request('select @name', function(err) {
	err instanceof TypeError
	err.code === "ETYPE"
	err.message === "Value does not match the type of parameter 'name'."
});
request.addParameter("name", TYPES.TinyInt, 999);
connection.execSql(request);
```

I preffer the second one, but I would like you to tell me what suits your needs the best. I also count with an option to disable type checks.

arthurschreiber:
Passing a wrong/out-of-bounds value to `request.addParameter` should cause an `Error` to be thrown right there, not later when the request is executed. It is a programmer error and happens when executing a fully synchronous function, making this an async error does not seem "correct" to me.

bretcope:
It's a matter of mixing error handling strategies. While we certainly _could_ give an earlier error in the form of an exception, it means there is more than one place to check for an error. There is _always_ going to be a callback which receives an error argument, and users are well accustomed to checking it.  There is almost never going to be a try/catch around it because that's not the error handling strategy in most node.js libraries which involve async.

I don't see any harm in passing the error to the callback. I do see a lot of harm in throwing because it's going to bypass the error handling which users already have in place and crash their program instead. I would strongly object to that approach.

arthurschreiber:
@bretcope I understand what you're aiming for.

But let me play Devil's advocate here. :wink:

That's not even how core Node.JS / io.js code works. If there is a check that can be made e.g. before an asynchronous function is executed and this check shows a programmer error (not an operational error), even node.js internal libraries will throw an exception. E.g. see here: https://github.com/iojs/io.js/blob/v1.x/lib/_http_client.js#L46-L57

[...]

PR #245, opened 2015-02-12
Integrated type checks

patriksimek:
I finally found some time to finish integrated type validation. It's a basic set of validators that should prevent users seeing errors like `index out of range` or similar. It also adds basic type coercion, so `1` and `"1"` are both valid values for data type `Int`.

```javascript
request = new Request('select @name', function(err) {
    err instanceof RequestError
    err.code === "EPARAM"
    err.message === "Validation failed for parameter 'name'. Value must be between 0 and 255."
});
request.addParameter("name", TYPES.TinyInt, 999);
connection.execSql(request);
```

Type validation applies on `execSql`, `execute` and `callProcedure` methods.

Feature is based on #202 discussion.

I believe this can be merged without a problem, but I wanted to be absolutely sure I didn't forgot about anything so I opened this PR.

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -732,6 +732,11 @@ set xact_abort #{xact_abort}"""
 
   execSql: (request) ->
     request.transformIntoExecuteSqlRpc()
+    if request.error?
+      return process.nextTick =>
+        @debug.log request.error.message
+        request.callback request.error
+
     @makeRequest(request, TYPE.RPC_REQUEST, new RpcRequestPayload(request, @currentTransactionDescriptor(), @config.options))
   
   newBulkLoad: (table, callback) ->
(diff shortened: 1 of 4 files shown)


arthurschreiber:
This does a lot more than just type validation, it also does type coercion. I don't know how to feel about that.

patriksimek:
It adds type coercion for two cases - `string` to `number` and `string` to `date`. Both are very useful and should not cause a problem. Btw type coercion already exists in Tedious, this is just a small expansion.

patriksimek:
I'm wrong, as I can see from the code that `string` to `number` already exsists in Tedious. So this only adds `string` to `date` type coercion.

bretcope:
I think the string to date coercion is quite useful. As far as string to number, that rubs me a bit wrong as well, but since it already exists, I suppose it doesn't matter.

I wonder if in the future we should consider having a sort of strict-mode setting. Might be more maintenance than it's worth. May also be a candidate for a breaking change in 2.0 someday.

PR #251, opened 2015-02-23
Drop old node versions

arthurschreiber:
The last released version of 0.8 was in July 2014, and the last released version of 0.6 was somewhere in 2012. I think we can safely drop support for those.


bretcope:
+1

PR #225, opened 2015-01-09
[WIP] Stream based token parser

arthurschreiber:
This replaces the token parser inside tedious with a stream based version based on the [dissolve](https://github.com/deoxxa/dissolve) module.

This is pretty much a big ugly hack at the moment, but it passes all tedious unit and integration tests. (Also, the application I'm working on passes all it's tests with the new parser). Beware, here be dragons.

The idea behind these changes is to make the token parsing code more performant. Right now, if a token is split over multiple tds messages, parsing is retried each time a new message is received from scratch. Obviously, this results in incredibly bad performance.

The `dissolve` module gives us a way to do chunk based parsing of tokens, without having to throw away the parsing results even in the case that tokens span multiple packages.

Here's some performance numbers (see `benchmarks/benchmarks.coffee`):

### Old Parser

```
$ node_modules/.bin/coffee benchmarks/query.coffee
Benchmarking 'nvarchar (small)'
nvarchar (small) x 68.06 ops/sec ±0.32% (22 runs sampled)
Memory: 5.4609375
Benchmarking 'nvarchar (large)'
nvarchar (large) x 1.84 ops/sec ±1.45% (14 runs sampled)
Memory: 19.265625
Benchmarking 'varbinary (small)'
varbinary (small) x 66.09 ops/sec ±0.14% (14 runs sampled)
Memory: 3.6171875
Benchmarking 'varbinary (large)'
varbinary (large) x 1.89 ops/sec ±1.64% (14 runs sampled)
Memory: 242.21484375
Benchmarking 'varbinary (huge)'
varbinary (huge) x 0.02 ops/sec ±9.94% (5 runs sampled)
Memory: 324.49609375
Done!
```

### New Parser

```
$ node_modules/.bin/coffee benchmarks/query.coffee
Benchmarking 'nvarchar (small)'
nvarchar (small) x 69.67 ops/sec ±0.50% (29 runs sampled)
Memory: 7.828125
Benchmarking 'nvarchar (large)'
nvarchar (large) x 7.01 ops/sec ±1.82% (37 runs sampled)
Memory: 31.6328125
Benchmarking 'varbinary (small)'
varbinary (small) x 65.99 ops/sec ±0.19% (13 runs sampled)
Memory: 6.296875
Benchmarking 'varbinary (large)'
varbinary (large) x 8.09 ops/sec ±2.41% (41 runs sampled)
Memory: 51.69140625
Benchmarking 'varbinary (huge)'
[...]

PR #252, opened 2015-02-23
Fix AppVeyor CI Flakiness

arthurschreiber:
AppVeyor builds sometimes fail because requests fail to complete in the default request time of 15 seconds, so let's increase the timeout value a bit.


arthurschreiber:
I'll merge this for now. The request timeout should be changed once #225 gets merged, as that will improve performance quite a bit.

PR #253, opened 2015-02-24
fix daylight saving issue for non utc dates

lchngr:
In some cases the date written and read from database are not the same. This only happens if the connection is established with options.useUTC = false. My node and sql server run both in the CET/CEST timezone.

Example: 
Input date "2015-06-18T22:00:00Z" -> 2015-06-19 00:00:00 CEST
Output date "2015-06-17T22:00:00Z" -> 2015-06-18 00:00:00 CEST

This was caused because the EPOCH_DATE used to calculate the days since 1900-01-01 did not consider the daylight savings difference. The calculated difference was 42171.958333333336, which was floored down to 42171. When taking the DST into account the difference would be 42172 days.

fixed datatypes:
datetime
datetime2
smalldatetime
date

### DIFF
--- a/src/data-type.coffee
+++ b/src/data-type.coffee
@@ -2,7 +2,7 @@ guidParser = require('./guid-parser')
 NULL = (1 << 16) - 1
 EPOCH_DATE = new Date(1900, 0, 1)
 UTC_EPOCH_DATE = new Date(Date.UTC(1900, 0, 1))
-YEAR_ONE = new Date(2000, 0, -730118).getTime()
+YEAR_ONE = new Date(2000, 0, -730118)
 UTC_YEAR_ONE = Date.UTC(2000, 0, -730118)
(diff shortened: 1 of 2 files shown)


arthurschreiber:
Hey @lchngr, many thanks for your contribution.

Would you be so kind to provide a unit and/or integration test that shows that this fix is working correctly and to ensure we don't reintroduce this in the future?

lchngr:
Hi @arthurschreiber 
I think I'll be able to provide some unit tests, but for the integration tests I think I'll need some guidance, how to do it correctly. This environment is pretty new for me.

jstans:
I just spent a while tracking down the source of this bug and was just about to make a pull request until I saw this sat at the top :)

When can we expect a version bump with this fix? I'm using the node-mssql library in my project which is dependent on this.

PR #254, opened 2015-02-25
Publish npm modules through travis

arthurschreiber:
This extends the travis config to automatically publish a new `tedious` package after a git tag was pushed and the travis build for the tag passed.

---

This probably needs some discussion. //cc @bretcope @pekim @patriksimek 

The idea here is to use automation wherever possible, and to align the workflow between all collaborators. E.g. the 1.10.0 release was a bit messy, we (still) don't have a git tag for it, and the tests actually failed on travis, because the package.json contained the `^` symbol which does not work on Node < `0.10`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,12 @@ node_js:
   - "1.8"
   - "2.0"
   - "2.3"
+
+deploy:
+  provider: npm
+  email: "schreiber.arthur@googlemail.com"
+  api_key:
+    secure: bsf8+yRHbphPCaP6w0Fqg66zOEyBss/XGpKqSiESrBI4cbIPAKKvUL/SFzibKqRV6mtYYJGmwRSaS4okhwSeCvi0hnHqAMBQQs863eatoUNbTTrGfH5EB97jaqkcdpCCvtup2Uqu+GcHXRgW1HsQ51jAntfY5DAazVsuRCTNNHI=
+  on:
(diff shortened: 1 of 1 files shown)


patriksimek:
Sorry for not tagging the repo, I thought GH for Mac synchronizes the tags automatically, but I was wrong. I just did that manually.

About automatic npm publishing - does it increase the version automatically? I wasn't able to figure it out from official docs.

arthurschreiber:
It takes whatever you have in `package.json`, I suppose. So you prepare all changes, push those to the repo, wait for the build to be green, tag the release, and a package will be generated and pushed to npm.

patriksimek:
So, +1 for me.

bretcope:
This seems reasonable. We should still use `npm version` to do the version bumps and commit tagging (though you still have to remember to push the tags). But having Travis do the actual packaging seems like a good idea.

Long term it would be nice if there was an automated way to publish to an npm feed every time anything is pushed to master. Maybe a `tedious-unstable` package. That way it would be easier for people to try out changes before publishing to the primary feed.

arthurschreiber:
Some issues here. Apparently, travis does run the deployment hook for each build that is run. We're running two builds right now (`0.10` and `0.11`). It's possible to only run a build if some preconditions are met, but it's still not perfect. Not sure if we want to proceed here or drop this... :disappointed:

Issue #262, opened 2015-04-23
Changes port number to 11024 in Azure


koistya:
After connection is established the original connection string (`config`) is modified, `config.options.port` is set to `11024` thus all subsequent connections using this `config` var fail with a timeout.

This only happens when code is executed in Azure Web App. Wasn't able to reproduce in locally a local environment, connecting to the exact same database.

Perhaps it's related to re-route which may work differently when connection is established from Azure datacenter vs from outside network.

arthurschreiber:
We shouldn't modify options objects when passed in, I've seen the same thing happen as well.

alexanderpreston:
Hi Guys, I have a production application in Azure that is down due to the port number being overridden to 11085. After the override all subsequent connections fail with a timeout.

Is there a quick work around?

Was it caused by the recent Azure SQL Server V12 upgrade?

koistya:
I'm using this workaround for now:

```js
let config = () => Object.assign({}, {/* connection string */});
let connection = new sql.Connection(config());
```

lee-houghton:
If you're getting the error all the time, perhaps you could do some `Object.observe()` magic to figure out where in the tedious code is overwriting the passed-in config? (Unless the Tedious guys already know where it's happening, in which case ignore me.)

koistya:
@lee-houghton I have tried your solution, it prints:

```
config.options.port overwritten:
Trace
    at D:\home\site\wwwroot\server.js:51:13
    at runMicrotasksCallback (node.js:337:7)
    at process._tickCallback (node.js:355:11)
```

bretcope:
As far as I can tell, the only place we would reassign the config port is if we receive a `ENVCHANGE:Routing` token from the server. That's probably still the wrong thing to do since all future connection attempts will go there instead of the originally provided details. However, what is provided back from the server should be a valid re-route.

Not sure if this is what's happening here. It looks like the Tedious implementation of the re-route is correct, but it's hard to know for sure. Perhaps @christopheranderson would be able to help track this down to assess whether the port change is due to an ENVCHANGE token.

gleeds:
I'm seeing the exact same thing.  I upgrade my Azure database to v12, and now only the first connection works in my application running on an Azure Website.  When I run the application locally connected to the production Azure database, it works fine.

gleeds:
[...]

PR #269, opened 2015-05-27
Fix for issue #262 changes port number

gleeds:
Actual issue was when a connection is redirected to a different server
name as well as port number, only the change to the port number was
being saved.  Pushing the redirection servername into the options
object fixes the problem.

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -82,7 +82,7 @@ class Connection extends EventEmitter
         connectTimeout: ->
           @transitionTo(@STATE.FINAL)
         reconnect: ->
-          @config.server = @routingData.server
+          @config.options.redirectionServer = @routingData.server
           @config.options.port = @routingData.port
           @transitionTo(@STATE.CONNECTING)
 
@@ -471,7 +471,7 @@ class Connection extends EventEmitter
       @connectOnPort(@config.options.port)
     else
       instanceLookup(
-        @config.server
+        @config.options.redirectionServer or @config.server
         @config.options.instanceName
         (message, port) =>
           if @state == @STATE.FINAL
@@ -486,7 +486,7 @@ class Connection extends EventEmitter
     @socket = new Socket({})
 
     connectOpts =
-      host: @config.server
+      host: @config.options.redirectionServer or @config.server
       port: port
 
     if @config.options.localAddress
@@ -517,7 +517,7 @@ class Connection extends EventEmitter
       @requestTimer = setTimeout(@requestTimeout, @config.options.requestTimeout)
 
   connectTimeout: =>
-    message = "Failed to connect to #{@config.server}:#{@config.options.port} in #{@config.options.connectTimeout}ms"
+    message = "Failed to connect to #{@config.options.redirectionServer or @config.server}:#{@config.options.port} in #{@config.options.connectTimeout}ms"
 
     @debug.log(message)
     @emit('connect', ConnectionError(message, 'ETIMEOUT'))
@@ -561,7 +561,7 @@ class Connection extends EventEmitter
 
   socketError: (error) =>
     if @state == @STATE.CONNECTING
-      message = "Failed to connect to #{@config.server}:#{@config.options.port} - #{error.message}"
+      message = "Failed to connect to #{@config.options.redirectionServer or @config.server}:#{@config.options.port} - #{error.message}"
       @debug.log(message)
       @emit('connect', ConnectionError(message, 'ESOCKET'))
     else
@@ -573,7 +573,7 @@ class Connection extends EventEmitter
   socketConnect: =>
     @socket.setKeepAlive(true, KEEP_ALIVE_INITIAL_DELAY)
     @closed = false
-    @debug.log("connected to #{@config.server}:#{@config.options.port}")
+    @debug.log("connected to #{@config.options.redirectionServer or @config.server}:#{@config.options.port}")
     @dispatchEvent('socketConnect')
 
   socketEnd: =>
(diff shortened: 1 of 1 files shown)


bretcope:
I think we need to go a little bit further with this fix. I think we should have a `redirectionPort` or similar property as well, and I think we should make sure the redirect server and port only last for one redirect (`connect()` should clear them out after using them). This way, you always attempt to connect to the server/port which you specified in the config. This is good, because you might be connecting to an availability group which redirects you to a specific node. If that node drops out, you really want to connect to the AG again and have it tell you where to go rather than constantly trying to reach the dead node.

gleeds:
I see.  That would mean that the issue here was opposite of what I was thinking, i.e., that the problem was that redirection port was being persisted between connects when it shouldn't have as opposed to the server name change not being persisted.  On the plus side I think that's an easier problem to address.  I can make that change if you'd like.

bretcope:
To be clear, I'm suggesting that _neither_ should be persisted beyond a single reconnect. If you're interested in working on it, that would be great, because I certainly don't have time.

gleeds:
Yes, that is my understanding.  I've got an updated fix that I'm trying to test now in an environment I provisioned to test this bug.

bretcope:
Sorry, I didn't notice you had pushed an update to the PR. I'll take a look at it.

SyRenity:
@bretcope 
Any chance to get this PR included in the next release?

tysjiang:
also waiting for this PR, just started noticing the issue connections to azure sql gets messed up intermittently

arthurschreiber:
I'll look into this this weekend.

bretcope:
@SyRenity @tysjiang have you confirmed that these changes fix the issues in your environment?

SyRenity:
@bretcope On development and staging yes (when installed manually), need either NPM or suggestion for installing CoffeeScript modules via npm github.

SyRenity:
I did noticed a new type of errors recently, not sure how they related to Tedious though.
Often, in a long running process, the Azure SQL starts dropping connections with:


name: 'ConnectionError',
  message: 'Database \'XXXX\' on server \'XXXX\' is not currently available.  Please retry the co
nnection later.  If the problem persists, contact customer support, and provide them the session tracing I
D of \'{82694C35-1E39-XXXX-XXXX-XXXXX}\'.',
  code: 'ELOGIN'

This happens less often, but still requires a node restart.
I contacted MS for support, however I think this should be checked in context of this PR too.

SyRenity:
@arthurschreiber Any news?

gleeds:
@SyRenity When you get that connection error, is the sever name the one from your configuration, or is it a redirection server?  Something like this: `12HEXDIGITS.tr6.southcentralus1-a.worker.database.windows.net`

SyRenity:
@gleeds No, it's a different name like a8ff6bcXXXX (X are numerical), and also without the region and MS domains.

gleeds:
Hmm, googling around I see a lot of forum posts about that error to the effect of "Azure Database are ephemeral and you need to have retry logic in place."  I don't have enough experience with the intricacies Azure to say if that is the case here or not.  Did MS Support give you any information?

https://social.msdn.microsoft.com/Forums/azure/en-US/b386f02d-3782-432d-9135-e5bf4f4e8bf9/database-xxx-on-server-yyy-is-not-currently-available
http://stackoverflow.com/questions/23193503/azure-sqlexception-database-on-server-is-not-currently-available
https://social.technet.microsoft.com/Forums/en-US/a1d4be73-2fd9-4524-a312-deecf07c6908/azure-sql-connection-errors?forum=ssdsgetstarted

SyRenity:
@gleeds Not yet, still working on it with them.

SyRenity:
@gleeds Got a reply from MS:

1.       The login failure for your database happened for 1 minute 30 seconds because of ‘High availability’ failover for the replicas that we keep in the backend.

2.       The logins were failed from XXXXXX to XXXXXXX   UTC (~1:30 min in total)

In such situations we recommend to do a retry which should be successful. Please let me know if you have any questions/concerns regarding the same.

Unfortunately Tedious seems to stay in this status, until the Node is restarted.

As said I haven't seen any reconnection logic in Tedious, what would be a good solution here?

arthurschreiber:
I still think this does not actually do the right thing. Tedious should never ever modify the config object that is passed into the constructor, ever.

We should store away the passed in config and do a deep clone of it, and reset the config on every connection attempt to what we stored away.

SyRenity:
Got the error again, even with the pull request merged inside:
Thu Jul 09 2015 12:38:40 GMT+0000 (UTC) error {"name":"RequestError","message":"Requests can only be made in the LoggedIn state, not the Final state","code":"EINVALIDSTATE"}

It certainly appears that due to the transient nature of Azure, Tedious is unable to coop with such configuration changes.

Any advice how to handle it, other then plain exiting node.js once this error encountered?

gleeds:
@arthurschreiber , do you think that Tedious should do the copy before `defaultConfig` runs and sets up all of the default values?  `@config` also get's modified later to set the negotiated `tdsVersion`.

gleeds:
@arthurschreiber or @bretcope, if connection.coffee should be using a deep clone of the configuration as Aurthur sudgested, what do you think of this approach,  first, replacing the usage of `@config` throughout connection.coffee with `@currentConfig`, everywhere except for the constructor, then turning defaultConfig into a function that handles the cloning of the canonical `@config` provided by the user to the active `@currentConfig` used by the class for connections.  defaultConfig would look something like this:

```coffeescript
defaultConfig: ->
    @currentConfig.options = {}
    @currentConfig.options.textsize = @config.options.textsize || DEFAULT_TEXTSIZE
    @currentConfig.options.connectTimeout = @config.options.connectTimeout || DEFAULT_CONNECT_TIMEOUT
    @currentConfig.options.requestTimeout = @config.options.requestTimeout || DEFAULT_CLIENT_REQUEST_TIMEOUT
    @currentConfig.options.cancelTimeout  = @config.options.cancelTimeout || DEFAULT_CANCEL_TIMEOUT
#and so on...
```

If the initial values are ever needed later, then we could go back to `@config`.  Alternatively we could just rename `@config` in the constructor to something else for a more terse commit.

bretcope:
It might be easier and less error prone to rename `@config` to `@initialConfig` than to update all of the usages and make sure people don't accidentally add usages in the future.

The real trick is likely in knowing when the config should be reset, and when it shouldn't.

arthurschreiber:
Fixed by #309 and #310. Will be released in `1.11.5` and `1.12.3`.

PR #273, opened 2015-06-12
Do not pass additional arguments to transaction continuation.

arthurschreiber:
This fixes a tiny issue where the continuation that was passed to the `transaction` method
would be called with one more argument than expected. This argument was always `undefined`,
and this bug was only affecting the `arguments.length` value.

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -831,7 +831,7 @@ set xact_abort #{xact_abort}"""
 
     txDone = (err, done) =>
       args = []
-      args.push(arguments[i]) for i in [2..arguments.length]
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Would anyone mind if I push a new release with this fix in place?

bretcope:
Looks good.

arthurschreiber:
Appveyor build is failing due to timeouts. Everything else looks fine.

arthurschreiber:
I merged this to the wrong branch. :disappointed: I'll get that fixed and merge it to the master branch.

PR #278, opened 2015-06-18
data-type validation should validate 'Invalid Date'

praneetkinnser:
The current data-type.js validation for datetime, does not account for 'Invalid Date' validation. As a result, the driver throws an uncaught exception.

Example: new Date('16:24') will pass the validation function but when reaches writable-tracking-buffer.js, just blows up. Such request, if within a transaction, leaves the transaction unresolved till the app exits.

### DIFF
--- a/src/data-type.coffee
+++ b/src/data-type.coffee
@@ -125,11 +125,9 @@ TYPE =
         buffer.writeUInt8(0)
     validate: (value) ->
       if not value? then return null
-      if `value == "Invalid Date"` then return new TypeError "Invalid date."
-      if value instanceof Date then return value
-      value = Date.parse value
+      if not (value instanceof Date) then value = Date.parse value
       if isNaN value then return new TypeError "Invalid date."
-      value
+      return value
   0x3B:
     type: 'FLT4'
     name: 'Real'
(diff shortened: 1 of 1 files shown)


bretcope:
I haven't checked for sure, but it doesn't look to me like this will actually perform the correct check on a real `Date` value since it's checking against a string. I also don't love that it introduces a second code path for checking invalid date in the same function.

I think a better way would be something like this:

```coffeescript
     validate: (value) ->
       if not value? then return null
       if not (value instanceof Date) value = Date.parse value
       if isNaN value then return new TypeError "Invalid date."
       return value
```

praneetkinnser:
So, it does an equality check when an Invalid Date is sent (instance type is still Date in such cases). However, I do agree with your point of having two code paths for returning error. I have made those changes in the second commit. 

```cs
validate: (value) ->
      if not value? then return null
      if not (value instanceof Date) then value = Date.parse value
      if isNaN value then return new TypeError "Invalid date."
      return value
```
```js
function validate(value) {

    if (value == null) {
        return null;
    }

    if (!(value instanceof Date)) {
        value = Date.parse(value);
    }

    if (isNaN(value)) {
        return new TypeError("Invalid date.");
    }
    return value;
}

console.log(validate(null));    //null
console.log(validate('5'));     //988693200000
console.log(validate(''));      //TypeError
console.log(validate('abc'));   //TypeError

console.log(validate('2015/12/14 11:16:34'));   //1450113394000
console.log(validate('2015/12/14'));    //1450072800000
console.log(validate('2015-12-14'));    //1450051200000
console.log(validate('11:16:34'));      //TypeError

console.log(validate(new Date('5')));   //Tue May 01 2001 00:00:00 GMT-0500 (CDT)
console.log(validate(new Date('abc'))); //TypeError
console.log(validate(new Date('16:34')));   //TypeError
console.log(validate(new Date('11:16:34')));    //TypeError
console.log(validate(new Date('2015/12/14 11:16:34'))); //Mon Dec 14 2015 11:16:34 GMT-0600 (CST)
```

bretcope:
That looks reasonable to me.

PR #285, opened 2015-07-29
Stream Parser using Generators

arthurschreiber:
### TL;DR

Make Tedious fast and use less memory using the magic of generators.

### Stream Parser using Generators

This is an alternative to #225.

Instead of using the `dissolve` module to do the parsing, I'm using ES2015 Generators instead. This makes the code **a lot** more readable and the necessary changes to convert the existing code to stream parser **a lot** less intrusive (check the diff to see how "simple" the actual changes are).

As ES2015 Generators are only available by default with io.js, I'm using babel as an additional transpilation step to convert the generators to ES5 compatible code.

This introduces a few new runtime dependencies (`babel-runtime`, `bl` and `readable-stream`) and a new development dependency (`babel`).

---

Benchmarks:

Transpiled generators are generally fast. Here's a benchmark from running the code that is generated by compiling the coffeescript code into JS and transpiling that using babel:

```
$ node node_modules/coffee-script/bin/coffee benchmarks/query.coffee
Benchmarking 'nvarchar (small)'
nvarchar (small) x 71.71 ops/sec ±0.89% (36 runs sampled)
Memory: 3.5
Benchmarking 'nvarchar (large)'
nvarchar (large) x 9.86 ops/sec ±1.94% (49 runs sampled)
Memory: 21.16015625
Benchmarking 'varbinary (small)'
varbinary (small) x 66.51 ops/sec ±0.23% (15 runs sampled)
Memory: 6.5234375
Benchmarking 'varbinary (4)'
varbinary (4) x 66.66 ops/sec ±0.28% (16 runs sampled)
Memory: 2.59375
Benchmarking 'varbinary (large)'
varbinary (large) x 22.08 ops/sec ±2.33% (53 runs sampled)
Memory: 47.87109375
Benchmarking 'varbinary (huge)'
varbinary (huge) x 2.58 ops/sec ±1.47% (17 runs sampled)
Memory: 60.8828125
Done!
```

This is faster than the change that I proposed in #225, while being less intrusive and a lot easier to maintain in the long term.

---

Regarding the inclusion of `babel`: I'd love to convert Tedious from CoffeeScript to ES2015. CoffeeScript was a welcome improvement over plain old javascript, but nowadays we have ES2015 giving us most of the sugar that CoffeeScript provided while being natively supported by io.js (and it's easily transpiled using `babel` for older Node.JS versions).

This is something that we maintainers probably need to discuss seperately from this issue.

---

Any thoughts?

### DIFF
--- a/benchmarks/benchmarks.coffee
+++ b/benchmarks/benchmarks.coffee
@@ -1,0 +1,122 @@
+{Request, TYPES} = require "../src/tedious"
+
+module.exports =
+
+  "nvarchar (small)":
+    setup: (connection, cb) ->
+      request = new Request "CREATE TABLE #benchmark ([value] nvarchar(max))", (err) ->
+        return cb(err) if err
+
+        request = new Request "INSERT INTO #benchmark ([value]) VALUES (@value)", cb
+        request.addParameter("value", TYPES.NVarChar, "asdf")
+        connection.execSql(request)
+
+      connection.execSqlBatch(request)
+
+    exec: (connection, cb) ->
+      request = new Request "SELECT * FROM #benchmark", cb
+      connection.execSql(request)
+
+    teardown: (connection, cb) ->
+      request = new Request "DROP TABLE #benchmark", cb
+      connection.execSqlBatch(request)
+
+  "nvarchar (large)":
+    setup: (connection, cb) ->
+      request = new Request "CREATE TABLE #benchmark ([value] nvarchar(max))", (err) ->
+        return cb(err) if err
+
+        request = new Request "INSERT INTO #benchmark ([value]) VALUES (@value)", cb
+        request.addParameter("value", TYPES.NVarChar, new Array(5 * 1024 * 1024).join("x"))
+        connection.execSql(request)
+
+      connection.execSqlBatch(request)
+
+    exec: (connection, cb) ->
+      request = new Request "SELECT * FROM #benchmark", cb
+      connection.execSql(request)
+
+    teardown: (connection, cb) ->
+      request = new Request "DROP TABLE #benchmark", cb
+      connection.execSqlBatch(request)
+
+  "varbinary (small)":
+    setup: (connection, cb) ->
+      request = new Request "CREATE TABLE #benchmark ([value] varbinary(max))", (err) ->
+        return cb(err) if err
+
+        request = new Request "INSERT INTO #benchmark ([value]) VALUES (@value)", cb
+        request.addParameter("value", TYPES.VarBinary, new Buffer("asdf"))
+        connection.execSql(request)
+
+      connection.execSqlBatch(request)
+
+    exec: (connection, cb) ->
+      request = new Request "SELECT * FROM #benchmark", cb
+      connection.execSql(request)
+
+    teardown: (connection, cb) ->
+      request = new Request "DROP TABLE #benchmark", cb
+      connection.execSqlBatch(request)
+
+  "varbinary (4)":
+    setup: (connection, cb) ->
+      request = new Request "CREATE TABLE #benchmark ([value] varbinary(4))", (err) ->
+        return cb(err) if err
+
+        request = new Request "INSERT INTO #benchmark ([value]) VALUES (@value)", cb
+        request.addParameter("value", TYPES.VarBinary, new Buffer("asdf"))
+        connection.execSql(request)
+
+      connection.execSqlBatch(request)
+
+    exec: (connection, cb) ->
+      request = new Request "SELECT * FROM #benchmark", cb
+      connection.execSql(request)
+
+    teardown: (connection, cb) ->
+      request = new Request "DROP TABLE #benchmark", cb
+      connection.execSqlBatch(request)
+
+
+  "varbinary (large)":
+    setup: (connection, cb) ->
+      request = new Request "CREATE TABLE #benchmark ([value] varbinary(max))", (err) ->
+        return cb(err) if err
+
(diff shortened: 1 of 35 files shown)


arthurschreiber:
I still need to figure out how to chain this all together (e.g. for running the unit tests). `npm prepublish` will already create fully functional output files.

arthurschreiber:
By the way, I also executed the benchmarks against `node-sqlserver`, and here's the numbers I get:

```
Benchmarking 'nvarchar (small)'
nvarchar (small) x 67.30 ops/sec ±0.19% (20 runs sampled)
Memory: 2.0859375
Benchmarking 'nvarchar (large)'
nvarchar (large) x 19.90 ops/sec ±4.60% (52 runs sampled)
Memory: 15.390625
Benchmarking 'varbinary (small)'
varbinary (small) x 65.53 ops/sec ±0.22% (12 runs sampled)
Memory: 0.47265625
Benchmarking 'varbinary (4)'
varbinary (4) x 65.48 ops/sec ±0.47% (11 runs sampled)
Memory: 1.13671875
Benchmarking 'varbinary (large)'
varbinary (large) x 11.84 ops/sec ±2.97% (58 runs sampled)
Memory: 2.01171875
Benchmarking 'varbinary (huge)'
varbinary (huge) x 1.23 ops/sec ±2.42% (11 runs sampled)
Memory: 119.41796875
Done!
```

So with this PR, tedious will be both faster and use less memory when handling binary data than the official, binary Microsoft TDS driver for Node. :laughing:

arthurschreiber:
I'm not quite sure why the build is passing on Travis, as everything should be failing right now.

bretcope:
I love the idea of using generators. It also gives the opportunity to completely revamp the Tedious API, which is, as the name implies, rather tedious to use. I always thought the API should look something like what I did with [Tedium](https://github.com/bretcope/tedium). It may be time to start thinking about what Tedious 2.0 looks like.

Having a transpiled ES5 version is fine, but generators are the way of the future, and pushing people in that direction is the right way to go. I would prefer to see the reverse situation from now, where the generators version _is_ the base library, and if you need an old-style library, that can be provided via a wrapper library.

Unfortunately I'm no longer actively using Tedious, and my time for contributing is practically non-existent, but I'm always happy to give feedback.

arthurschreiber:
@bretcope well, the change here is rather basic and is limited to the internals of Tedious.

Regarding "the future": I think generators are only a stopgap solution, "async/await" using Promises is the way to go. But that should not block us from using generators right now to improve. :smiley:

bretcope:
I agree that they're a stopgap, but transitioning from generators to async/await is pretty straight forward compared to transitioning from callbacks. Basically replacing `function *` with `async function` and `yield *` with `await`. It's the same principle (functions which can be resumed). The difference is primarily whether you need a library as the promise runner vs having it built into the language.

I haven't seen any implementation of async/await yet, so it's hard to know how far off those are. I think if there were a desire to redo the api (much needed, in my opinion), then generators would be the way to go for now.

I'm also ___HUGELY___ in favor of ditching coffeescript. I've always found it frustrating, difficult to read, and I think it limits the number of people interested in contributing. But, like I said, I'm not particularly active in writing code for Tedious anymore, so I don't have much stake in the decision.

arthurschreiber:
The build is passing now (on both Travis as well as Appveyor).

@patriksimek @bretcope @pekim I'd like to go ahead and merge this, and then maybe start iterating towards moving Tedious off of CoffeeScript and onto idiomatic "ES6/ES7/ES2015/ESnext/whatever else it's called today" using babel.

The source will still need transpilation to be able to execute on older versions of iojs and Node.js, but it will be "plain" JavaScript. :smile: 

I'll also continue working on moving the tests to mocha (I really can't take this nodeunit nonsense anymore).

I'll try to do this conversion in steps, which means we'll have both ES* code and CoffeeScript code living together until the migration is done.

If you got any concerns, please let me know, Feedback is very welcome.

rossipedia:
I know I don't do much on this these days, but I'm definitely :+1: for bringing tedious into the modern JS age

pekim:
It all sounds good to me.

My biggest regret with tedious was using coffeescript. And I never did like the backtracking of packet token parsing.

Great job Arthur.

arthurschreiber:
@pekim Thanks.

Okay, so here's my plan for now.

* I'll test these changes with `node-mssql`. If everything is green there, I'll merge this as-is and do a release for `1.12`. The `node-mssql` test suite written by @patriksimek was able to highlight issues with #225, so if it does not complain about the changes here, I feel confident to release this. Also, this does not break any of the ~2000 test cases that the application I'm working on has.

* I'll start converting things to idiomatic next-gen javascript. I'll start out with the token parsing code that is using generators so I don't have to pipe the coffeescript generated output through babel anymore. As a part of this, I'll also be converting the test suite over to use mocha and chai.

Everything will be done step-by-step, and after each major milestone, I want to release a new tedious version, on a regular basis. I don't want to end in a situation where development on the old code base becomes stale because a "rewrite" is in the works.

Thanks everyone.

rossipedia:
Great work @arthurschreiber!

arthurschreiber:
I just ran this against `node-mssql`, and all tests are passing. I'll release `1.11.3` with the changes that were made on `master` so far, and then I'll push another release containing the generator changes.

arthurschreiber:
Okay, I finally got releases via Travis working. Apparently, you can't release to npm using npm 1.x, but the node 0.10 build job on travis still uses that npm version. Changing our builds to always update to the latest npm fixed that issue. Sorry for the resulting noise.

I'll go ahead and merge this and release it as `1.12.0`.

PR #290, opened 2015-08-01
Convert token parsing functions to JavaScript

arthurschreiber:
This PR is meant as a first step to start converting `tedious` from CoffeeScript to next-gen JavaScript.

i started with the files in the `src/token/` folder, as that allows me to remove the stopgap solution that I put in place that compiled CoffeeScript output using `babel`.


arthurschreiber:
I just pushed the latest batch of changes. All files inside `src` are now converted to next-gen javascript code.

It's not perfect (as it's based on the code generated by CoffeeScript), but I tried to make use of many ES6 features like classes, const, let, and so on.

bretcope:
Not sure what your plans were, but I would suggest squashing this into one commit before merging.

arthurschreiber:
Yup, that's a good idea. I was mainly waiting for anyone to take a quick look and tell me this is not totally crazy before I merge this. :smile:

Issue #307, opened 2015-08-28
tedious + azure v12 connectionError/ELOGIN


spanditcaa:
I am having an issue with tedious 1.11.2 and SQL Azure v12 on iojs 2.5.  The error message we get back is different from what is being reported in #272 - 
````
name: ConnectionError
message: Database 'ABC' on server 'aaa1aaaaa' is not currently available.  Please retry the connection later.  If the problem persists, contact customer support, and provide them the session tracing ID of 'CDF3AEE3-4E70-4DAF-85BA-311AE76BEB29'.
code: ELOGIN
````
In case there are any microsoft folks around, additional tracing IDs we've seen include:
````
C6B4392B-08FE-48DC-B058-E2D8A4D52D16
D0CEA175-3F84-44CB-B02E-793CDE938E88
118A46D8-03B8-4409-9849-7F3B3C3425CE
````
It has happened several times a day since our v12 upgrade.  When this does occur, I notice that all instances of our API appear to lose connection simultaneously.  As with #272 if we restart the affected instances everything reconnects and works normally (for 6-12 hours).

Thanks

spanditcaa:
Working with Microsoft, we've believe this is the same port/token issue and should resolve with https://github.com/pekim/tedious/pull/269.

arthurschreiber:
Hey Sanjay, this is great, thanks. I'll take a look at this asap. I'm currently holding off releasing a new version of tedious because I want to get #303 solved beforehand.

Would a new version in the 1.11.x release line with this fix be good enough for you for now?

arthurschreiber:
Fixed by #309 and #310. Will be released in `1.11.5` and `1.12.3`.

PR #309, opened 2015-09-03
Don't overwrite configuration with routing data.

### LINKED ISSUE #258: Note about Azure SQL DB v12 Support in 1.8.1 or higher
A couple of people in Azure have included me on incidents where folks were trying to use old tedious drivers with their v12 platform and it gets escalated to their engineering team. Is it a good idea to put a note under ["SQL Server Support"](http://pekim.github.io/tedious/index.html) to alert people to this?

### LINKED ISSUE #262: Changes port number to 11024 in Azure

arthurschreiber:
This fixes #262, #307 and #269. I also believe it fixes #258.

It's a bit simpler than the fix proposed in #269, and passes all unit and integration tests.

I can push a beta package to npm with this fix in place, if anyone is interested in testing this. I'll release `1.11.5` including this fix and I'll port the fix to 1.12 and the `master` branch once someone confirms this is working for them.

//cc @gleeds @bretcope @tysjiang @SyRenity @spanditcaa PTAL

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -82,8 +82,6 @@ class Connection extends EventEmitter
         connectTimeout: ->
           @transitionTo(@STATE.FINAL)
         reconnect: ->
-          @config.server = @routingData.server
-          @config.options.port = @routingData.port
           @transitionTo(@STATE.CONNECTING)
 
     SENT_TLSSSLNEGOTIATION:
@@ -486,8 +484,12 @@ class Connection extends EventEmitter
     @socket = new Socket({})
 
     connectOpts =
-      host: @config.server
-      port: port
+      host: @routingData?.server || @config.server
+      port: @routingData?.port || port
+
+    # Unset routing information so we reuse the initial config on the next
+    # connection attempt again.
+    @routingData = undefined
 
     if @config.options.localAddress
       connectOpts.localAddress = @config.options.localAddress


arthurschreiber:
This should also fix #307.

spanditcaa:
Thanks @arthurschreiber we would be happy to help test.

arthurschreiber:
@spanditcaa Let me go ahead and merge this. I'll push a beta release in a few minutes.

arthurschreiber:
@spanditcaa I just pushed `tedious@1.11.5-beta.2-g2780aac`. Please let me know if that version works for you.

gleeds:
@arthurschreiber I'll test this in my environment too.

gleeds:
This fails for me with the following error:
```
{ [ConnectionError: Database 'MYDATABASENAME' on server '' is not currently available.  Please retry the connection later.  If the problem persists, contact customer support, and provide them the session tracing ID of '{02AC65E1-3E6B-44D8-91A3-D6F977303E1F}'.]
  name: 'ConnectionError',
  message: 'Database \'MYDATABASENAME\' on server \'\' is not currently available.  Please retry the connection later.  If the problem persists, contact customer support, and provide them the session tracing ID of \'{02AC65E1-3E6B-44D8-91A3-D6F977303E1F}\'.',
  code: 'ELOGIN' }
```

I'm pretty sure the problem is that the `serverName` property in `sendLogin7Packet` needs to match the name of the server you are being redirected to, and not the initial server from your configuration.

arthurschreiber:
@gleeds This is what @christopheranderson also described in https://github.com/pekim/tedious/issues/223, right? Let me take a look into what's going on.

Issue #272, opened 2015-06-11
Tedious peredically stops working with Azure v12 DB


SyRenity:
After we upgraded to v12, we often get the following errors:
```
{"RequestError\",\"message\":\"Requests can only be made in the LoggedIn state, not the Final 
state\",\"code\":\"EINVALIDSTATE\"}
```
The only way to resolve them is to restart the Node.JS process.

Is there a way to cause Tedious to reconnect to Azure DB, other then existing Node?
I've looked through documentation, however didn't find any reconnection parameters.

We use version 1.11.0 of Tedious.

bretcope:
Can you see if the changes in [this pull request](https://github.com/pekim/tedious/pull/269) solves your issue?

SyRenity:
@bretcope Haven't encountered it on development machine with the patch, however I need to test it on staging to be sure. Any idea what would be the best way to ensure the prebublish runs after npm install from github?

SyRenity:
@bretcope Any idea?

SyRenity:
Unfortunately the pull request has not helped, we need to stop the Node and restart it via watchdog for the time being

PR #310, opened 2015-09-03
Send correct servername in Login7 after rerouting

arthurschreiber:
In addition to #309.

I can push another beta package, just need some other pair of eyes looking over this.

//cc @gleeds @bretcope @tysjiang @SyRenity @spanditcaa PTAL

### DIFF
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -616,7 +616,7 @@ class Connection extends EventEmitter
       userName: @config.userName
       password: @config.password
       database: @config.options.database
-      serverName: @config.server
+      serverName: @routingData?.server || @config.server
       appName: @config.options.appName
       packetSize: @config.options.packetSize
       tdsVersion: @config.options.tdsVersion
--- a/src/connection.coffee
+++ b/src/connection.coffee
@@ -487,10 +487,6 @@ class Connection extends EventEmitter
       host: @routingData?.server || @config.server
       port: @routingData?.port || port
 
-    # Unset routing information so we reuse the initial config on the next
-    # connection attempt again.
-    @routingData = undefined
-
     if @config.options.localAddress
       connectOpts.localAddress = @config.options.localAddress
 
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Appveyor build for 0.10 failed because of some unrelated issue. I'll merge this and tag a new beta release.

spanditcaa:
will update and let you know what we find.

arthurschreiber:
Published as `tedious@1.11.5-beta.5-gb7b6e4d`.

gleeds:
I retested with this version, it seems like it has fixed the issue for me (#262).  I personally never ran into the #272, so I can't confirm if that is resolved.

spanditcaa:
We're working with some Microsoft folks who were able to trigger a failover/reroute for us.  The driver restored connection as desired.

spanditcaa:
However - this was from a host running outside the Azure datacenter.  It seems we aren't able to connect from a host running in Azure.  We're troubleshooting this now.  As I understand, there is a different connection path from internal to Azure datacenters.

gleeds:
@spanditcaa I can only get the initial redirect when connecting from a host inside Azure, when I connect from a host outside of azure I never get redirected to another server (maybe because the other server isn't publicly routable?)

Another thing that keeps tripping me up is that I have tedious twice in the app I'm trying to test on (because I'm dumb), both in the root, and as a part of the mssql npm library, and I keep accidentally updating only one or the other.  Also the Kudu deployment thing seems to sporadically run npm update when I deploy.  You might double check your package.json for tedious on the deployed server to make sure it really is the right one.

spanditcaa:
@gleeds @arthurschreiber  

We had what seems to be an unrelated deployment issue.  I ran a clean deployment in Azure and everything is working normally.  We also had another failover/reroute triggered and the hosts picked up the new servers and reconnected in about 200ms.

arthurschreiber:
That sounds great. I'm really glad the fix is working for you.

I'll need to polish the changes a bit and create some unit tests for this, and then I'll push new 1.11 and 1.12 releases containing the fix.

I won't get around to doing this today, but I hope I can find some time this weekend.

arthurschreiber:
Everyone listening in on this, I released `1.11.5` and `1.12.3` containing this fix.

spanditcaa:
Thanks @arthurschreiber.  I think 1.11.5 may need an npm publish.  I only see 1.12.3 added there.

arthurschreiber:
@spanditcaa It should be available now. There's some issue with pushing to the npm registry using older  versions of the npm client, which caused the push from Travis to fail. I pushed the version manually for now.

PR #322, opened 2015-10-09
Binary data truncated on insert

bjouhier:
Binary data is truncated on insert. Only the first byte of the buffer is written, the following ones are set to 0.

This is a regression from 1.8.0. I fixed it by restoring the `declaration()` function from 1.8.0.

I've added a unit test. It should probably be merged into one of the existing unit tests but I did not know how to best do it.

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -954,8 +954,18 @@ export const TYPE = {
     dataLengthLength: 2,
     maximumLength: 8000,
 
-    declaration: function() {
-      return 'binary';
+    declaration: function(parameter) {
+      var length;
+      if (parameter.length) {
+        length = parameter.length;
+      } else if (parameter.value != null) {
+        length = parameter.value.length || 1;
+      } else if (parameter.value === null && !parameter.output) {
+        length = 1;
+      } else {
+        length = this.maximumLength;
+      }
+      return "binary(" + length + ")";
     },
 
     resolveLength: function(parameter) {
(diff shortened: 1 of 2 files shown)


pprice:
Would be awesome to merge this, I'm facing this bug right now.

arthurschreiber:
@bjouhier Would you mind converting the test case from CoffeScript back into JS? Also, instead of creating and dropping a table, could you please use a temporary table in the test case?

pprice:
If any one stumbles across this PR looking for the issue, here is the work around I'm using for relatively short binary data. 

```javascript
var sql = 'INSERT INTO [Table] (BinaryColumn) VALUES (CONVERT(varbinary(max), @binary, 2))'
var request = new Request(sql, function(err, rows) { 
   //... 
});

request.addParameter('binary', TYPES.VarChar, myBuffer.toString('hex')); 
```

Essentially serializing the buffer into a hex string, and having TSQL convert it back to varbinary.

bjouhier:
@arthurschreiber I'll do the changes.

arthurschreiber:
I merged this and will fix up the remaining issues in a follow up PR. Thanks for your contribution! :heart:

bjouhier:
Thanks Arthur. I did not find the time this week and I'm glad you merged it as is.

arthurschreiber:
@bjouhier No worries. I didn't find the time to put your fix into a release, but I'll try to do that on sunday.

PR #325, opened 2015-10-16
call callback for request on socket error

ashelley:
When a socket error occurs the request callback is never called.  This patch calls the request callback when the connection state is set to SENT_CLIENT_REQUEST.  A simple test has been added.  This patch probably needs more scrutiny and testing.

If you are using tedious-connection-pool this patch will not have any effect as discussed here:

https://github.com/pekim/tedious-connection-pool/issues/23

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1055,7 +1055,10 @@ Connection.prototype.STATE = {
   SENT_CLIENT_REQUEST: {
     name: 'SentClientRequest',
     events: {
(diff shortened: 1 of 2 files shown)


arthurschreiber:
This is looking good. I'll incorporate this change plus a few fixes and push a new release early next week. Thanks again! :heart:

arthurschreiber:
Closed in favor of #330.

PR #336, opened 2015-10-28
Fix issues with DateTime2 parameters

### LINKED ISSUE #335: Data type 0x2A has an invalid data length or metadata length
I have bad luck bumping tedious version from 1.12.7 to 1.13.0. I keep getting the following error:
```
Error: The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Parameter 4 ("@createdAt"): Data type 0x2A has an invalid data length or metadata length.
```

Query with params:
```
insert into app.Certificates (clockSkew, createdAt) values (@clockSkew, @createdAt)
[{"name":"clockSkew","value":331155,"type":"int"},{"name":"createdAt","value":"2015-10-27T18:30:26.260Z","type":"datetime2"}]
```

My database is hosted on Azure.

arthurschreiber:
This fixes #335.

I accidentally introduced a typo when converting this part of the code from CoffeeScript to Javascript. Unfortunately, there were no test cases for the `DateTime2` parameter type, so the issue was not found through our automated test cases.

This fixes the typo and adds test cases to ensure we don't introduce a similar issue again.

//cc @pawelrychlik @sentient Can you take a look? If this looks good to you, I'll push a new release with this fix included.

### DIFF
--- a/src/rpcrequest-payload.js
+++ b/src/rpcrequest-payload.js
@@ -72,7 +72,7 @@ export default class RpcRequestPayload {
       if (type.hasScale) {
         if (parameter.scale) {
           param.scale = parameter.scale;
-        } else if (type.resolvePrecision) {
+        } else if (type.resolveScale) {
           param.scale = type.resolveScale(parameter);
         }
       }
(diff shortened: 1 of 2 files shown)


sentient:
This fixed my issue with DateTime2. 

Thank you for the quick fix

Issue #183, opened 2014-09-08
Support SQL Server Availability Groups


rossipedia:
At Stack Exchange we need to support [SQL Server AlwaysOn Availability Groups](http://msdn.microsoft.com/en-us/library/ff877884.aspx), and bringing this support to Tedious would make it a much more viable solution for those in similar situations. There's definitely quite a few parts to getting the full feature-set implemented, so I'd like to keep this PR open as kind of a running discussion/integration during the implementation process.

The first set of changes involves the [`fReadOnlyIntent`](http://msdn.microsoft.com/en-us/library/dd304019.aspx) bit. I found that Tedious had the wrong bits set for the various `TypeFlags`. Fixing that allowed setting the `fReadOnlyIntent` bit and getting the proper `ENVCHANGE` token (type 20) back from SQL Server after logging in. After that it was a matter of fixing up the state machine to properly transition and reconnect to the server/port indicated in the `ENVCHANGE` token.

Hopefully (meaning I haven't tested this yet), this will have the added benefit of gracefully handling a manual fail-over as well.

The next step is to handle the situation where a DNS lookup for the SQL Server host name returns multiple IP addresses, which should be attempted in order until a connection is made. I'm pretty sure I'll be able to re-use the `REROUTING` state for that. We're going to use Microsoft's native driver as the reference implementation for how connection attempts should be performed. See [SQL Server Native Client Support for High Availability, Disaster Recovery](http://msdn.microsoft.com/en-us/library/gg471494.aspx).

chrislukkk:
We are a software company also uses SQL Server AG Listener as well as its [High Availability and Disaster Recovery](https://msdn.microsoft.com/en-us/library/gg471494.aspx) feature, we found this feature is not completely implemented in Tedious(specificlly the **MultiSubnetFailover** connection property), so we want to add it to Tedious to support that. 

The Logic will be using [dns.lookup](https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback) to get all the ip addresses of the Listener's DNS name, and connect to multiple ip addresses almost at same time and found the first one correctly connected.

[...]

PR #346, opened 2015-12-02
Multi subnet failover using dns resolve

chrislukkk:
Implement multi subnet failover #183 using dns resolve to resolve multiple ips from one host.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -400,13 +400,20 @@ export default class Connection extends EventEmitter {
         this.socket.connect(connectOpts);
     } else {
       //look up both ipv4 and ipv6 addresses
-      this.resolveIPaddresses(connectOpts.host, (addresses) => {
+      this.resolveIPaddresses(connectOpts.host, (resolve4Err, resolve6Err, addresses) => {
+        var message = resolve4Err ? ("Error in IPV4 address resolution: " + resolve4Err.message) : null; 
+        message += resolve6Err ? ("\nError in IPV6 address resolution: " + resolve6Err.message) : null;
+        if(message)
+        {
+          this.debug.log(message);
+        }
+        
         if (addresses && addresses.length > 1) {
           this.multiConnectOnPort(addresses, connectOpts.port, (opts) => {
             this.socket.connect(opts || connectOpts);
           });
         } else {
-          this.socket.connect(connectOpts);
+          return this.emit('connect', ConnectionError(message, 'ERESOLVEIPADDR'));
         }
       });  
     }
@@ -426,19 +433,23 @@ export default class Connection extends EventEmitter {
   resolveIPaddresses(host, callback) {
     let ipv4Addresses = undefined;
     let ipv6Addresses = undefined;
+    let resolve4Err = undefined;
+    let resolve6Err = undefined;
     //resolve IPV4
     dns.resolve4(host, (err, addresses) => {
+      resolve4Err = err;
       ipv4Addresses = addresses || new Array();
       if (ipv6Addresses) {
-        callback(ipv4Addresses.concat(ipv6Addresses));
+        callback(resolve4Err, resolve6Err, ipv4Addresses.concat(ipv6Addresses));
       }
     });
     
     //resolve IPV6
     dns.resolve6(host, (err, addresses) => {
+      resolve6Err = err;
       ipv6Addresses = addresses || new Array();
       if (ipv4Addresses) {
-        callback(ipv4Addresses.concat(ipv6Addresses));
+        callback(resolve4Err, resolve6Err, ipv4Addresses.concat(ipv6Addresses));
       }
     });
   }
--- a/src/connection.js
+++ b/src/connection.js
@@ -399,18 +399,15 @@ export default class Connection extends EventEmitter {
     }
     else {
       //look up both ipv4 and ipv6 addresses
-      Dns.lookup(connectOpts.host, { all : true }, (err, addresses) => {
-        if(typeof addresses !== 'undefined' && addresses.length > 1)
-        {
+      this.resolveIPaddresses(connectOpts.host, (addresses) => {
+        if(typeof addresses !== 'undefined' && addresses.length > 1) {
           this.multiConnectOnPort(addresses, connectOpts.port, (opts) => {
             this.socket.connect(opts || connectOpts);
           });
-        }
-        else
-        {
+        } else {
           this.socket.connect(connectOpts);
-        } 
-      });
+        }
+      })  
     }
 
     this.socket.on('error', this.socketError);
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Hey @chrislukkk,

first of all, thank you for the work on this. Multi Subnet failover is a great feature to have as part of tedious. :+1: 

There is a few things that need to be fixed up before this will be ready to be merged:

* From what I understand, the default behavior for a TDS connection when connecting to a hostname is that the hostname is resolved via DNS and for each returned record, a connection attempt is performed. These attempts are performed serially.

* If `multiSubnetFailover` is true, these connections should happen in parallel, the way you implemented here.

* I think we should ignore the `multiSubnetFailover` option in case that the `host` that is being connected to is an ip address, otherwise the code will try to lookup an IP address in DNS, which won't work.

Do you feel up to the task of extending this PR to implement these missing functionality?

src/connection.js
```diff
@@ -12,6 +12,7 @@ import RpcRequestPayload from './rpcrequest-payload';
 import SqlBatchPayload from './sqlbatch-payload';
 import MessageIO from './message-io';
 import { Socket } from 'net';
+import Dns from 'dns';
```
arthurschreiber: I'd prefer `dns` over `Dns` here.

src/connection.js
```diff
@@ -392,7 +393,23 @@ export default class Connection extends EventEmitter {
...
     }
-    this.socket.connect(connectOpts);
+    
+    if (!multiSubnetFailover) {
+        this.socket.connect(connectOpts);
+    }
+    else {
+      //look up both ipv4 and ipv6 addresses
+      this.resolveIPaddresses(connectOpts.host, (addresses) => {
+        if(typeof addresses !== 'undefined' && addresses.length > 1) {
```
arthurschreiber: The existing code has a single space between `if` and `(`.

src/connection.js
```diff
@@ -392,7 +393,23 @@ export default class Connection extends EventEmitter {
     if (this.config.options.localAddress) {
       connectOpts.localAddress = this.config.options.localAddress;
     }
-    this.socket.connect(connectOpts);
+    
+    if (!multiSubnetFailover) {
+        this.socket.connect(connectOpts);
+    }
+    else {
```
arthurschreiber: Can you put the closing `}` and the `else {` on the same line, as in other areas of the existing code base?

src/connection.js
arthurschreiber: No need for a `typeof` here: `if (addresses && addresses.length)`

src/connection.js
```diff
@@ -405,6 +422,62 @@ export default class Connection extends EventEmitter {
     return this.messageIo.on('secure', this.emit.bind(this, 'secure'));
   }
 
+  resolveIPaddresses(host, callback) {
+    let ipv4Addresses = undefined;
+    let ipv6Addresses = undefined;
+    //resolve IPV4
+    Dns.resolve4(host, (err, addresses) => {
```
arthurschreiber: You need to handle `err` by passing it back to `callback` and add correct error handling to wherever `resolveIPaddresses` is called from.

src/connection.js
```diff
@@ -405,6 +422,62 @@ export default class Connection extends EventEmitter {
...
+      ipv4Addresses = addresses || new Array();
+      if (ipv6Addresses) {
+        callback(ipv4Addresses.concat(ipv6Addresses));
+      }
+    });
+    
+    //resolve IPV6
+    Dns.resolve6(host, (err, addresses) => {
```
arthurschreiber: Same as above.

src/connection.js
```diff
@@ -405,6 +422,62 @@ export default class Connection extends EventEmitter {
...
+      ipv6Addresses = addresses || new Array();
+      if (ipv4Addresses) {
+        callback(ipv4Addresses.concat(ipv6Addresses));
+      }
+    });
+  }
+ 
+  multiConnectOnPort(addresses, port, callback) {
```
arthurschreiber: This function feels a bit lost. It's called `multiConnectOnPort` and I'd expect it to establish the connection, but that's not what it does. The way it interacts with `connectOnPort ` also does not feel right to me. :disappointed:

src/connection.js
tinyjoker: I want to keep check of ```addresses.length >1``` here. If there is only one ip address for a host, we should fall back to regular connect logic

chrislukkk: I want to keep check of ```addresses.length >1``` here. If there is only one ip address for a host, we should fall back to regular connect logic

arthurschreiber: Ah, right. :+1:

src/connection.js
chrislukkk: fixed with new commit

src/connection.js
chrislukkk: fixed with new commit

src/connection.js
chrislukkk: fixed with new commit

arthurschreiber:
Thanks @chrislukkk for the fix-ups.

src/connection.js
chrislukkk: The ```multiConnectOnPort ``` function only figures out which IP address resolved from dns can be connected, and invoke callback in ```connectOnPort```, which calls ```socket.connect``` using connection opts and establish the connection.

I could try to move the logic of 'establish connection using ```this.socket.connect```' from ```connectOnPort``` to ```multiConnectOnPort ```. Do you think it will make the implementation clearer? :smile:

arthurschreiber: Mmh, I'm not sure. We can also leave it as-is for now to get this merged faster and then reiterate on it, if that makes sense.

chrislukkk:
Hi @arthurschreiber, 

Thanks a lot for the thorough code review!

Regarding to the things you mentioned:

a. You are correct that TDS client by default connects *sequentially* if there are multiple ```ip``` addresses bind to an ```host```, and ```multisubnetfailover``` make the workflow in *parallel*.

To support **sequential** connection, however, will change the default connect behavior of tedious, because tedious by default  use ```socket.connect```(in [net.js](https://github.com/nodejs/node/blob/master/lib/net.js)), which internally calls ```dns.lookup``` to find the first found IP record and connect to it(not finding all IP record and connect sequentially). To correct that we will use need to use same logic of ```multiconnect``` for connection regardless whether ```multisubnetfailover``` is set. We were worried that will be too much for this change.

b. I will add the logic to check if the ```host``` is an IP address. There is an npm package called ip which does the ip format check already. Is it ok to use other npm package or should I just copy the regex over?

src/connection.js
chrislukkk: I'm good with that. Thanks!

src/connection.js
chrislukkk: fixed in 3755e4932477a09cc0513a7512c7e74932ede578

arthurschreiber:
Hey @chrislukkk 

Thanks for all the fixups. As you already found out, there is an `isIP` method in the core `net` module. This is looking pretty good, I'll probably merge this as-is ASAP and then fix up some smaller issues in some later PRs. Thank you so much for this! :heart:

chrislukkk:
Hi @arthurschreiber, is there any update one the process of this PR? Thank you!

arthurschreiber:
Hey @chrislukkk 

I'm really sorry this has been dragged on so long. I'll see if I can carve out a bit of free time to review this thoroughly and get this change integrated. :+1:

arthurschreiber:
Hey @chrislukkk,

I'm closing this in favor of #362. Please take a look and let me know what you think!

PR #350, opened 2015-12-19
Add Variant data type support

patriksimek:


src/value-parser.js
```diff
@@ -317,6 +315,33 @@ export default function(parser, metaData, options, callback) {
 
         case 'UDT':
           return readMaxBinary(parser, callback);
+        
+        case 'Variant':
+          let valueMetaData = metaData.valueMetaData = {};
+          Object.defineProperty(valueMetaData, 'isVariantValue', {value: true});
```
arthurschreiber: Is there a specific reason you're using `Object.defineProperty` here?

patriksimek: I need to tell `readDataLength` function that I'm parsing a value of variant data type so I already know it's length. `Object.defineProperty` is used to disable enumeration of this property so it is not inspectable in column metadata. It is not necessary but I preffer to hide those internal pointers from end user.

PR #354, opened 2016-01-09
Connect callback

moander:


tvrprasad:
What's the purpose of this callback? Just curious.

moander:
The Request class takes a callback in the constructor but the Connection does not.

tvrprasad:
I looked at the Request and it doesn't seem like the callback is ever invoked.
I ran a test with some logging and I don't see if invoked.

I think that's dead code that needs to be removed.

tvrprasad:
Closing the PR as the callback parameter added is unused.

PR #362, opened 2016-02-09
Multi Subnet Failover

arthurschreiber:
This is based on the work done by @chrislukkk over in #346.

This PR extracts the whole logic for how a connection is established into a separate `Connector` class. It also changes the way how connections are established to fully follow the Microsoft specification.
- Connections via IP-Address are established as before.
- Connections via hostname with `multiSubnetFailover: false` are established by looking up all IP Addresses from DNS and trying to connect serially.
- Connections via hostname with `multiSubnetFailover: true` are established by looking up all IP Addresses from DNS and opening a socket to all of them in parallel. The socket that connects first will be used, all other sockets will be destroyed.

@chrislukkk Thanks for your hard work on the initial implementation of this. Please take a look. I'll add extensive unit tests ASAP, but this should be working just fine.


mitant:
Any news on the status of this?  I tried it with a service I'm developing to connect to my load-balanced mssql database and it continues to occasionally time out.  I might try chrislukk's branch

arthurschreiber:
@mitant Do you have any more information on when these timeouts happen?

takotuesday:
Whats the status of this PR? The inability to set `multisubnetfailover=true` is a deal breaker. Any update would be greatly appreaciated. Cheers!

arthurschreiber:
@takotuesday Would you mind taking this code and going on a test drive with it? @chrislukkk hasn't reported back.

mitant:
Guys, the problem is not that you can't set "multisubnetfailover=true", it's that the functionality doesn't appear to exist in this driver package to behave in the way that multisubnetfailover=true is supposed to indicate.

mitant:
@arthurschreiber The behavior that I observed in my testing was that it would behave the same as if the functionality wasn't implemented--that is if I pointed the connection at the "load balancing" host it would occasionally time out.  If I point at the db instance behind the load balancing host, it will work flawlessly, but then it's not load balanced anymore.

Issue #396 has a similar issue but for some reason they can't connect at all.  I personally was able to connect and perform operations for some amount of time, which later would timeout.

jblaketc:
I'm on a project that we had to connect to a cluster DB. I'm using Sequelize as the connector wrapper, and it wasn't working until I took this code and manually put it in the node_modules/tedious folder. It's been in there for about a month, and we're just about out of UAT with no connection issues so far. So as far as my experience goes, this is working good. It solved the issue I was having immediately. 

The project will go to Production in a few weeks so it'll get stress-tested then if that's what y'all are looking for.

arthurschreiber:
@jblaketc Thank you so much for reporting back and letting me know that this is working for you!

@mitant Did you use a checkout of this Pull Request in your testing? As long as this PR is not merged, `tedious` does not support `multiSubnetFailover`. With this PR, the functionality should be implemented according to Microsofts specification.

mitant:
I'm getting some good results.  Earlier in the day pre-pull I was getting db timeouts on startup of my loopback project (which depends on tedious).  I haven't observed any timeouts recently.  As I do more dev work today I'll make a note if I run into any more issues

p.s. to anyone watching this thread, I used the following to load this pr
npm install git+https://github.com/tediousjs/tedious.git#pull/362/head

Hobbit71:
I have tested this and so far looks promising. After some initial issues with connection, i can reliably run my app tests and pass. 

p.s. Thanks for the load link @mitant

arthurschreiber:
Thanks everyone who tested this! ❤️

It looks like this is stable enough to be merged into `master`, but there's a consistent build failure with Node 0.10 on AppVeyor, and I'm not sure what is causing this. Once I figure that out, I'll get this merged. 👍

tvrprasad:
@arthurschreiber I tried running this on my local machine against Node 0.10 and am able to repro. I see the error below. In binary-insert-test.js, test.done() is being called from 'end' callback. Is it possible that this PR is somehow causing the callback to not be invoked? I tried adding callbacks for 'close' and 'error' but that didn't fix it. I'll look at this more later but sharing in case there is something here that may be obvious to you.

For some reason this part of the output is not there in on the failure link.


binary-insert-test:
FAILURES: Undone tests (or their setups/teardowns):
- insertBinary

To fix this, make sure all tests call test.done()

arthurschreiber:
@tvrprasad Thanks for checking this out! I've not been able to reproduce this so far, but the output you pasted at least gives me some sort of direction! 😄

I'll try to take a deeper look into this after the weekend - hopefully we can find out what's wrong here so this can be merged and released soon!

tvrprasad:
@arthurschreiber Just for FYI, I was able to repro with Node version v0.10.46, running on Windows version 10.0.10586.

tvrprasad:
@arthurschreiber I debugged this more and I think I got this figured out.

The issue is at line 12 in Connector.js.
'req' returned by dns.lookup() is undefined when domain === 'localhost' and 'req.oncomplete' throws an exception.
Test config for the integrations tests sets domain as localhost and thus all the tests are failing.

Adding the below snippet of code got the tests to pass for me. This assumes localhost has both IPv4 and IPv6 loopback addresses. We could make this more robust by either:
1) Return only the IPv4 address or
2) Detect if IPv6 loopback address is enabled on the machine or
3) Get the hostname and pass that to dns.lookup(), which I verified works.

    if (domain.toLowerCase() === 'localhost') {
      const results = [];
      results.push({ address: '127.0.0.1', family: 4 });
      results.push({ address: '::1', family: 6 });

      callback(null, results);

      return undefined;
    }

arthurschreiber:
@tvrprasad Oh, wow, thanks for the looking into this. 👍 

I fixed the underlying bug, and CI seems to be happy now. I couldn't find the root of the issue as my development machine is running OS X and the bug was only happening on Windows 7 and later.

If you can take a look over the fix 👀, that'd be great. If everything looks good, this can be merged soon! 😄

tvrprasad:
@arthurschreiber Made a pass on your changes. Looks good to me!
Glad you found a cleaner fix than my hack :)

tvrprasad:
@arthurschreiber You were going to merge this change.  Is there anything blocking? Let me know if I can help.

Thanks,
Prasad

arthurschreiber:
There's two things that blocked me from merging this:

1. I haven't written any unit or integrations tests at all for this, so I feel very uncomfortable merging this as-is. Writing integration tests is a pain because it requires running a local DNS server, while writing unit-tests is hard because I don't want to expose the internals (but then they're also not exposed to the unit-test code).
1. I want to extract the code related to the `lookupAll` function inside the `connector.js` file into a separate npm package, as I don't think it's something that tedious should be responsible for.

I started working on number 2, but I think number 1 here is the more urgent issue. Maybe you can find some time to work on this?

tvrprasad:
I can spend some time on 1. One idea for unit testing:

1) Write a wrapper class for dns.lookup method, say dnsLookupWrapper, that has one method lookup.
2) In production initialize the class with with 'dns', in test code initialize with 'dnsMock'.
3) Have Connector class take dnsLookupWrapper as a parameter.
4) Have Connector class pas dnsLookupWrapper as a parameter to lookupAll method invocation.

Would that work for unit testing? If you have other ideas, I'd love to hear and help implement. I'm a JavaScript newbie.

tvrprasad:
@arthurschreiber You mentioned local dns server for integration testing. What exactly did you have in mind for the test? It's possible to add/remove DNS entries locally by just editing the hosts file. Worked as expected with dns.lookup() on my windows box.

arthurschreiber:
> 1) Write a wrapper class for dns.lookup method, say dnsLookupWrapper, that has one method lookup.
> 2) In production initialize the class with with 'dns', in test code initialize with 'dnsMock'.
> 3) Have Connector class take dnsLookupWrapper as a parameter.
> 4) Have Connector class pas dnsLookupWrapper as a parameter to lookupAll method invocation.

I'd have gone with something like the `sinon` npm module to stub/mock the calls to the (still not existing, see 2) `dns-lookup-all` module instead of going down the road of dependency injection explicit mock classes.

> You mentioned local dns server for integration testing. What exactly did you have in mind for the test?

We'd need a local DNS that returns entries where only one of the entries resolves to a reachable server.

tvrprasad:
Cool. I'll checkout out sinon to see how it works. For now I'll do the work to mock the calls to dns.lookup to learn how that works. Once you have the dns-lookup-all module ready, I can make it work with that. And btw, you're making dns-lookup-all into a module, not a package, right? Separate package would seem heavy weight.

For DNS, we can edit hosts file on the machine to fake DNS addresses (c:\windows\system32\drivers\etc\hosts on Windows and /etc/hosts on Linux). I tested this on Windows, it works for everything except 'localhost' and <simple localhost name>. It works for everything else including FDQN for host machine.

Can we merge this with unit tests only and track the integration test in a separate issue? There may be still be some messy details to work even if we go with the hosts file solution. If you agree I'll open a separate issue to track the integration test.

arthurschreiber:
> For DNS, we can edit hosts file on the machine to fake DNS addresses (c:\windows\system32\drivers\etc\hosts on Windows and /etc/hosts on Linux). I tested this on Windows, it works for everything except 'localhost' and . It works for everything else including FDQN for host machine.

This only allows specifying a single IP address per hostname, but not a combination of multiple IP addresses for a single hostname (which is all what the multisubnet failover option is about). I might be wrong, but some quick googling confirms this.

tvrprasad:
@arthurschreiber I've tried this on Windows and it does work for multiple ip addresses. I can try on Linux later today or over the weekend.

hosts file entries:
168.61.167.220 a.b.com
168.61.167.221 a.b.com
2607:f8b0:400a:800::200e a.b.com

dns.lookup output:
a.b.com addresses:  [ { address: '168.61.167.220', family: 4 },
  { address: '168.61.167.221', family: 4 },
  { address: '2607:f8b0:400a:800::200e', family: 6 } ]

tvrprasad:
Sent a PR for unit tests - https://github.com/tediousjs/tedious/pull/444
Thanks.

takotuesday:
@arthurschreiber whats the status of this PR? Waiting to use this feature in production :smile:

tvrprasad:
@arthurschreiber 
pop()
Can you make a pass and let me know if you have feedback or if I can merge this? Thanks.

takotuesday:
@arthurschreiber any update?

arthurschreiber:
@tvrprasad Thank you so much for the super quick turnaround time on reviewing my latest changes. ❤️ Your reviews and contributions are highly appreciated!

As you've seen, I added some more sophisticated testcases and refactored some of the code for better testability. I hope you're not sad I didn't merge the test cases you proposed in your other PR.

I pushed some fixes for some of the issues you've already mentioned. I'll give it another pass tomorrow, so if you have any more questions or things to point out, please let me know. 😁

tvrprasad:
@arthurschreiber I'm glad to see this feature move forward. If the test cases from my PR helped jump start your tests, I'll consider that a bonus :-)

I like your refactoring and the use of mitm 👍

arthurschreiber:
Ok, I think this is ready now (yay!). 🎉

This might need some more tweaking in future versions - we might have to take another look into how the connection timeouts are handled in light of parallel and sequential connections, but I'd rather ship this in it's current state and fix it up later if we find any weird interactions.

@tvrprasad Would you care to take another look?

src/connector.js
```diff
@@ -0,0 +1,150 @@
...
+
+      cb(null, socket);
+    };
+
+    socket.on('error', onError);
+    socket.on('connect', onConnect);
+  }
+
+  executeForHostname(cb) {
+    lookupAll(this.options.host, 4, (err, addresses) => {
```
tvrprasad: Any reason for supporting only IPv4?

arthurschreiber: Ah, you're right - there's no specific reason why we only support IPv4.

What we should have is a global option on whether you want to use IPv4 *or* IPv6. Otherwise, if your servers support both, we'll try connecting to each one of them twice.

tvrprasad: > we'll try connecting to each one of them twice.

Not sure what you mean there.

From https://nodejs.org/api/dns.html#dns_dns_lookup_hostname_options_callback:

_family <Number> - The record family. If present, must be the integer 4 or 6. If not provided, both IP v4 and v6 addresses are accepted._

That suggests, there is nothing special to do other than to pass on the user intent to lookupAll.

arthurschreiber: Right. What I mean is that we should default to IPv4, and allow users to specify IPv6 if they want to use it.

If we don't have a default, the DNS lookup will potentially return 2 entries (one with IPv4 and one with IPv6) for each server. When we do serial or parallel connections, we would connect to each server twice - which is weird and unexpected.

I can't imagine there are users with a setup where some servers in their HA setup speak only IPv4 and others speak only IPv6, because that would mean their servers could not even talk to each other.

arthurschreiber: Also, maybe I'm wrong and the DNS name that is registered will never have 2 different IPs (IPv4/IPv6) for the same physical host? Could you check with the SQLServer team, I could not find anything specific in the documentation.

tvrprasad: I confirmed with someone in the team, you can have IPv4/IPv6 addresses for a DNS name resolving to the same physical host.

I'm still wondering though if the default should be to try both IPv4 and IPv6. It does waste some resources by potentially attempting two connections to the same server but that takes the burden off the app developer to have to know which IP versions the server supports.

Let me know if you agree (or not) with my proposal on the default and I'll put together a PR accordingly.

tvrprasad: I would say, no, we don't need options to allow users to specify one specific IP version.

tvrprasad: Sent a small PR with changes. Pick it up if it makes sense :)
https://github.com/tediousjs/tedious/pull/481

arthurschreiber: @tvrprasad I'm wondering - can we just drop all of the sequential connection handling and just do parallel connections, always?  Do you see any disadvantages to that?

I'm asking because with the JDBC driver, `multiSubnetFailover=true` is highly suggested and should never have any negative effects.

Also, I think we should look into the instance listener connection code, I don't think that was updated for support of parallel connections.

tvrprasad: > I'm wondering - can we just drop all of the sequential connection handling and just do parallel connections, always? Do you see any disadvantages to that?

Parallel connections seem to have cause a few issues before with services getting overwhelmed on startup when they try to make a ton of parallel connections.

Oe strategy that has been adopted is to first try just the the first IP address returned by DNS lookup, and only if that fails, then attempt the parallel connections Now you have to slice the connection timeout between the first attempt and the second attempt. This slicing of the timeout between the first attempt and the rest, caused a different problem, in that it increased the likely of connection timeout. The team is now investigating yet another tweak, which is to do the slicing for local databases but not for Azure databases (or something close to that).

Summary is that, it can get complicated :) Sequential as the default and Parallel as opt-in sounds to me like a reasonable place to start for this feature.

> Also, I think we should look into the instance listener connection code, I don't think that was updated for support of parallel connections.

What's "instance listener connection code"? Is this required to merge this PR?

gshively11:
@arthurschreiber Great work on this branch. We are just rolling out AGs in production and I did some testing with this branch and a slightly modified [mssql](https://github.com/patriksimek/node-mssql) (to pass the new options to tedious). Everything looked good.

Do you have an idea of when this branch will hit master and if it'll be included in release 1.15.0?  And are there any missing pieces yet to be implemented?  Anything you would want help with?  Getting our node apps working with AGs is near the top of my TODO list, so I'm happy to provide any assistance needed.

takotuesday:
@arthurschreiber anything I can do to help expedite this? Thanks!

arthurschreiber:
@takotuesday I just pushed another change that would need review from @tvrprasad, then we should be good to go and release this as 1.15.0.

@tvrprasad Would you mind taking another look? I added `multiSubnetFailover` support when connecting via an instance name instead of specifying a specific port. It only supports UDPv4 now, do you think we need support for UDPv6 as well? The change is in https://github.com/tediousjs/tedious/pull/362/commits/2a7462d15e83f67cc9ef061ecf70f522e8837ea4.

src/instance-lookup.js
```diff
@@ -10,19 +11,41 @@ const MYSTERY_HEADER_LENGTH = 3;
...
-    }
+    clearTimeout(timer);
+    socket.close();
+
     message = message.toString('ascii', MYSTERY_HEADER_LENGTH);
     const port = parseBrowserResponse(message, instanceName);
-    socket.close();
+
     if (port) {
       return callback(undefined, port);
```
tvrprasad: You're firing socket.send for all the IP addresses returned at line 86.  You could end up invoking this multiple times, right?

src/instance-lookup.js
```diff
@@ -52,7 +73,23 @@ function instanceLookup(server, instanceName, callback, timeout, retries) {
       socket = dgram.createSocket('udp4');
       socket.on('error', onError);
       socket.on('message', onMessage);
-      socket.send(request, 0, request.length, SQL_SERVER_BROWSER_PORT, server);
+
+      if (multiSubnetFailover) {
```
tvrprasad: MultiSubnetFailover is handled differently here vs in Connector. In connector, we call lookupAll for both cases and then fire requests in parallel or in sequence depending on whether multiSubnetFailover succeeds or not. Whereas here, we only try one IP address for multiSubnetFailover == false. I think we should keep these consistent.

Also we're doing lookupAll twice for instance-lookup code path. Once in instanceLookup() method and then again in connectOnPort(). We should only need to do this once I think. One way to do that would to invoke lookupAll in connect() method for all cases and then pass addresses as a parameter for connectOnPort() method as well as instance-lookup().

One other suggestion for the instance-lookup case would be to use the same IP address that returned the port, in connectOnPort. I'm not sure if it's mandated but seems safe to do.

tvrprasad:
@arthurschreiber I think we should support UDPv6 to stay consistent with the non-instance name based code path. Also, I added a couple of other review comments.

Thanks for the updates!

src/instance-lookup.js
tvrprasad: > One other suggestion for the instance-lookup case would be to use the same IP address that returned the port, in connectOnPort. I'm not sure if it's mandated but seems safe to do.

Turns out, this is actually the wrong thing to do. What you're doing is correct, it's valid for the IP address that responded with the port to not be hosting a SQL Server instance.

arthurschreiber: > Also we're doing lookupAll twice for instance-lookup code path. Once in instanceLookup() method and then again in connectOnPort(). We should only need to do this once I think. One way to do that would to invoke lookupAll in connect() method for all cases and then pass addresses as a parameter for connectOnPort() method as well as instance-lookup().

I'd expect that the server browser response contains IP-addresses, not DNS host names, so the connection would be done without another DNS lookup.

tvrprasad: That sounds reasonable.

tvrprasad:
Sent a PR to address remaining issues with MSF. https://github.com/tediousjs/tedious/pull/502

tvrprasad:
@arthurschreiber The sender tests are failing in 0.10 and 0.12. I'll take a look.

takotuesday:
This is awwwwwwwwwwweessoommmmmeee, thank you @arthurschreiber!

arthurschreiber:
Releasing this as 1.15.0 now.

arthurschreiber:
Also, thank you to everyone who helped getting this over the finish line 🏁 !

tvrprasad:
This is great! Thanks :-)

amihanov:
Hi @arthurschreiber,
I'm working on a project and I need Sequelize to be able to use tedious with multiSubnetFailover to true. However, I can't find a way to do this. Is there a "clean" way to change tedious's behavior to have multiSubnetFailover property enabled by default ??
Thanks a bunch!

renaudholcombe:
@amihanov, you should be able to just pass the MultiSubnetFailover setting as a dialect option when initializing Sequelize.

amihanov:
@renaudholcombe, Yes!! thanks. it worked great. Here's the code (in case some can use it):

`var Sequelize = require ('sequelize');

var config = {
  adapter: 'mssql',
  host: 'myHost',
  user: 'myUser',
  password: 'myPassword',
  port: 1433,
  database: 'myDatabase',
  dialectOptions: {
    multiSubnetFailover: true
  },  
  pool: {
    max: 5,
    min: 0,
    idle: 10000
  },
  maxConcurrentQueries: 100,
  logging: false
};

var sequelize = new Sequelize(config.database, config.user, config.password, {
  host: config.host,
  port: config.port,
  dialect: config.adapter,
  dialectOptions: config.dialectOptions,
  pool: config.pool,
  maxConcurrentQueries: config.maxConcurrentQueries,
  logging: config.logging
});

sequelize
  .authenticate()
  .then(() => {
    console.log('Connection has been established successfully.');
  })
  .catch(err => {
    console.error('Unable to connect to the database:', err);
  });

`

amihanov:
@renaudholcombe, Yes!! thanks. it worked great. Here's the code (in case some can use it):

```javascript
var Sequelize = require ('sequelize');

var config = {
  adapter: 'mssql',
  host: 'myHost',
  user: 'myUser',
  password: 'myPassword',
  port: 1433,
  database: 'myDatabase',
  dialectOptions: {
    multiSubnetFailover: true
  },  
  pool: {
    max: 5,
    min: 0,
    idle: 10000
  },
  maxConcurrentQueries: 100,
  logging: false
};

var sequelize = new Sequelize(config.database, config.user, config.password, {
  host: config.host,
  port: config.port,
  dialect: config.adapter,
  dialectOptions: config.dialectOptions,
  pool: config.pool,
  maxConcurrentQueries: config.maxConcurrentQueries,
  logging: config.logging
});

sequelize
  .authenticate()
  .then(() => {
    console.log('Connection has been established successfully.');
  })
  .catch(err => {
    console.error('Unable to connect to the database:', err);
  });
 ```

PR #366, opened 2016-02-16
Document Request event 'prepared'

jordanbtucker:
In my testing, I had to wait for the `prepared` event before `Connection.execute` would work properly. I discovered this undocumented event by looking at the test cases.


arthurschreiber:
LGTM! ❤️ Many thanks for your contribution!

Issue #397, opened 2016-06-24
Unable to verify certificate/ca from server are valid


SimonHooker:
With encrypted=true I expect connection to a server with a self-signed certificate to fail due to the CA not being trusted.  Connecting though silently accepts the certificates.

Digging through the tls source this looks to be functionality supported by `tls.tlsSocket()` but not by `tls.createSecurePair()`

`tls.createSecurePair()` is deprecated as of node 6.x ( previously reported in https://github.com/pekim/tedious/issues/135 ) and so this might be good time to look at migrating over to using tls.tlsSocket as it will allow for connections to reject if unauthorised.

This may be considered a duplicate of https://github.com/pekim/tedious/issues/282

SimonHooker:
Have started work on what will hopefully work to allow this functionality - just needs a bit more work and hopefully some tests around it

https://github.com/SimonHooker/tedious/commit/d691dac2d54c244e93529ebe54fdc00fec0351b6

SimonHooker:
Have submitted a pull request

https://github.com/pekim/tedious/pull/399

Issue #135, opened 2014-03-10
createSecurePair() is deprecated, use TLSSocket instead


patriksimek:

bretcope:
This doesn't actually appear to be deprecated. http://nodejs.org/api/tls.html#tls_tls_createsecurepair_context_isserver_requestcert_rejectunauthorized

There is a note which says:

> NOTE: cleartext has the same APIs as [tls.TLSSocket](http://nodejs.org/api/tls.html#tls_class_tls_tlssocket)

But I don't see anything about deprecation.

Upon further digging, it looks like it was once deprecated, but the deprecation notice was [removed](https://github.com/joyent/node/blob/b3fcc245fb25539909ef1d5eaa01dbf92e168633/ChangeLog#L145) in 0.11.15.

SimonHooker:
This looks like it has since been deprecated according to the  
https://nodejs.org/api/tls.html#tls_tls_createsecurepair_context_isserver_requestcert_rejectunauthorized_options

Referenced in this commit in node 6.0 ( so not LTS yet ) - the favoured tls.tlsSocket is present in 4.x
https://github.com/nodejs/node/pull/6063

And referenced in these meeting minutes from May 2016
https://github.com/nodejs/node/blob/7f1111b1bb3f1af777e062cd7ca62dd0dc0fc8e0/doc/ctc-meetings/2016-05-04.md

PR #399, opened 2016-06-27
Adding options.trustServerCertificate #397

SimonHooker:
Adding a new optional option trustServerCertificate ( boolean, default: true )

If true, behavior is unchanged from the current (1.14) behavior in that encrypted connections are supported but the certificate is not verified in any way.

If false, the ssl.verifyError() method is called following connection throwing an error if it fails to verify i.e. due to being self-signed.  If a CA or certificate is specified within the cryptoCredentialsDetails then this allows users to verify against an expected self-signed certificate.

The node tls methods specify that a connection is always permitted when it comes to client connections, and that rejections should be handled in javascript ( https://github.com/nodejs/node/blob/master/src/node_crypto.cc )

The name for this argument is taken from the Microsoft JDBC driver which has an optional argument trustServerCertificate with the same behavior.  In the JDBC driver the default value is "false" howerver I chose a default of "true" to preserve current default functionality within tedious.


arthurschreiber:
`tls.createSecurePair` takes an argument that is called `rejectUnauthorized`, which is basically doing the same certificate verification you added into the `secure` callback.

What do you think about using the Node.JS core provided functionality here?

SimonHooker:
I did start down that route with a view to using rejectUnauthorized however that is a server-side flag only.  It gets passed through as argument 4 to `new Connection()`

If you follow that through to `node_crypto.cc` ( https://github.com/nodejs/node/blob/master/src/node_crypto.cc ) it comes into play lines 2845:2860, however if `!is_server` it always returns a verify mode of `SSL_VERIFY_NONE` with a comment indicating that `reject_unauthorised is ignored for clients` along with line 2863 containing the comment `// Always allow a connection. We'll reject in javascript.`

So yes I could repurpose the cryptoCredentialsDetails.rejectUnauthorised parameter without harming the functioning of TLS but felt it more usable to instead mirror the method already in place on another SQL Server driver.  If you would prefer I change this functionality to piggy-back on the rejectUnauthorised parameter of course.

arthurschreiber:
@SimonHooker You're totally right! ⚡

One thing that bugs me is `throw`ing the error. Looking at the implementation of the same functionality in Node.JS core, they pass on the error to `socket.destroy`: https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L1048-L1086

Is there any other logic in there we might want to steal? 😄

SimonHooker:
Hi Arthur,

That is so, however that is in the TLSSocket implementation and at present Tedious uses createSecurePair instead from `_tls_legacy.js`.  There is already a ticket for this to be swapped over to TLSSocket as createSecurePair is deprecated ( https://github.com/pekim/tedious/issues/135 ) and so perhaps at this time that would be the time to re-evaluate this implementation as undoubtedly this entire block of code will need revising.

Just as a heads-up, I took this solution in part from the TLSSocket implementation and partially from node-mysql's verification behaviour, though that also uses the TLSSocket implementation.

SimonHooker:
Ah great, just noticed that you have converted the coffee file to js and I've been editing the coffee file! Argh!  Right, back to square one on that one then.

SimonHooker:
There, have reimplemented in your `message-io.js` ( maybe `message-io.coffee` should be deleted at this point as it seems to be superceded by `message-io.js`? I didn't do that under this ticket though )

Have added in a destroy() to clean up the securePair however the destroy of tls.SecurePair does not accept any arguments and so cannot be implemented in the same way as the more modern tlsSocket - so have left the error being emitted.

But I now have a test failure to look into.

src/message-io.js
```diff
@@ -85,18 +85,29 @@ module.exports = class MessageIO extends EventEmitter {
...
-    });
+    this.securePair.on('secure', (_this) => {
+      return function() {
+        const cipher = _this.securePair.cleartext.getCipher();
+
+        if (!trustServerCertificate) {
+          const verifyError = this.ssl.verifyError();
+          if (verifyError) {
+            _this.securePair.destroy();
+            throw verifyError;
```
arthurschreiber: What about passting the `verifyError` into the `destroy` call just above? That should emit an `error` event on the connection instead of crashing the process. 😄

src/message-io.js
```diff
@@ -85,18 +85,29 @@ module.exports = class MessageIO extends EventEmitter {
...
     this.securePair = tls.createSecurePair(credentials);
     this.tlsNegotiationComplete = false;
 
-    this.securePair.on('secure', () => {
-      const cipher = this.securePair.cleartext.getCipher();
-      this.debug.log('TLS negotiated (' + cipher.name + ', ' + cipher.version + ')');
-      this.emit('secure', this.securePair.cleartext);
-      this.encryptAllFutureTraffic();
-    });
+    this.securePair.on('secure', (_this) => {
```
arthurschreiber: why `_this`?

src/message-io.js
SimonHooker: It would be nice, but look at lines 824:240 of https://github.com/nodejs/node/blob/master/lib/_tls_legacy.js where the `SecurePair.prototype.destroy` is.

Unfortunately it has no arguments so does not pass it through.  Once this gets changed to `TLSSocket` instead of `createSecurePair` then the error can be proxied in through the destroy as recommended.

Will try again on the emit front.

arthurschreiber: We can't change to TLSSocket, unfortunately. We have to open a non-TLS connection first and upgrade to TLS later in the process. 😞

arthurschreiber: You're right about `.destroy` on the `SecurePair`. But we probably also want to destroy the socket (`this.socket`) after destroying the `SecurePair`. You should be able to pass the error there.

src/message-io.js
SimonHooker: this.ssl comes from the callback function of this.securePair.on( 'secure' )

To get access to `this.ssl.verifyError` I needed access to the `this` scope from the `secure` event so I needed to change away from an arrow function but still needed access to the parent `this` scope to access the other methods such as `securePair` and `encryptAllFutureTraffic` as they are not defined in the "this" scope of the 'secure' event.

src/message-io.js
SimonHooker: Good to know, but what then happens when/if createSecurePair is removed - it sounds like that may be happening in Node 6 or so.  Surely we need to find a way to migrate to TLSSocket in the long term? (this is more a discussion for #135 though I suppose)

SimonHooker: Good point on the this.socket, I totally forgot about that - will sort that out

src/message-io.js
```diff
@@ -85,18 +85,30 @@ module.exports = class MessageIO extends EventEmitter {
...
+    this.securePair.on('secure', ((_this) => {
+      return function() {
+        const cipher = _this.securePair.cleartext.getCipher();
+
+        if (!trustServerCertificate) {
+          const verifyError = this.ssl.verifyError();
+          if (verifyError) {
+            _this.securePair.destroy();
+            _this.socket.destroy();
+            _this.emit('error', verifyError);
```
arthurschreiber: I think you'll need to `return` here, otherwise this will first emit the `error` event and then right after it will emit the `secure` event.

src/message-io.js
arthurschreiber: Won't the `this` inside the `secure` callback be equal to `this.securePair` outside?

src/message-io.js
SimonHooker: I thought that too at first, however the error emit appears to essentially throw.  I could of course add a blank return in, but it seemed to not serve any use overall.

src/message-io.js
SimonHooker: It will not, due to the fact that the arrow function `( (_this)=>{ return function() { } } )( this )` ends up returning `function(){}` with `_this` accessible, where `_this` is the outer `this` due to being passed in as argument.

That more traditional function then has the `this` scope from the `secure` event ( arrow functions do not have this, they do not have their own this scope )

To paraphrase from mozilla, `An arrow function expression ... does not bind it's own this, arguments, super, or new.target` - however it is the functions own `this` that I need bound.

src/message-io.js
arthurschreiber: The `error` event only throws if there is no listener attached to it, otherwise it will not throw. These semantics are not very clear. 😞

SimonHooker: No problem - have added the return in.  I feel happier with it being there anyway :)

src/message-io.js
SimonHooker: Have revised this block a bit to improve readability / understandability

SimonHooker:
Is there anything more I need to do to this prior to getting it merged in?  Rather require this functionality to allow me to get some clients up and running!

arthurschreiber:
@SimonHooker I'm in the process of setting up a dev environment again for working on tedious. I'll try to get this merged in as soon as possible! 😄

SimonHooker:
Hi there, sorry to nag on this but any idea if it'll take much longer?

Failing that, any advise as to how I can get npm to use my fork of tedious instead of this version?  I tried just installing SimonHooker/tedious (having merged the branch into my master) however it appears to not compile the /lib for some reason, any ideas what step(s) I've likely missed?

arthurschreiber:
Hey Simon,

Sorry this is taking so long. I've been in contact with the Node.js core
team to understand how to switch to the tls.connect API, which would keep
us from reimplement this functionality. I'll have something ready in the
next few days.
Simon Hooker <notifications@github.com> schrieb am Di., 12. Juli 2016 um
12:08:

SimonHooker:
Hi @arthurschreiber  

Just as a heads up, for an idea as to how to switch to tls.connect down the line it could be worth looking at libraries like https://github.com/mysqljs/mysql which, while not MSSQL, uses tls.connect for DB connection so could provide a good pointer.

In the shorter term it would be good to get this fix as-is in place, it can always be ported to using tls.connect at a later time but having the parameter as-is would provide parity with JDBC connectors without affecting server-side functionality - IIRC this functionality as-is is not necessarily within the tls.connect API either as there are flags within crypto that prevent it being paid attention to unless the instance is flagged as a server.

I've just now got a couple of clients instead of just 1 that are blocked by my inability to verify their SSL connections, which this pull request would resolve!

SimonHooker:
Hi there,

It's been almost 3 weeks!  What is preventing this being mergable?

SimonHooker:
Sorry hit the wrong button there.

It's been 3 more weeks - 4 weeks since the last change!  What is preventing this from being merged?

SimonHooker:
Closed this PR as it is superceded by #419

PR #403, opened 2016-07-04
Fix issue: sql.Time.

sant123:
Now when a parameter is of type "sql.Time" and the value is a string, it will try to parse it; instead of returning a TypeError('Invalid time');

https://github.com/patriksimek/node-mssql/issues/320

http://stackoverflow.com/questions/33098711/mssql-nodejs-module-throws-invalid-time-in-execution-of-a-stored-procedure

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1305,6 +1305,10 @@ const TYPE = module.exports.TYPE = {
       if (value instanceof Date) {
         return value;
       }
+      value = Date.parse(value);
+      if (!isNaN(value)){
+        return value;
+      }
       var timespan = (function(input) {
         var regex = /^([0-9]{1}|(?:0[0-9]|1[0-9]|2[0-3])+):([0-5]?[0-9])(?::([0-5]?[0-9])(?:\.(\d{1,9}))?)?$/;
         var result = (regex.exec(input) || []).slice(1, 5);
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1306,7 +1306,7 @@ const TYPE = module.exports.TYPE = {
         return value;
       }
       value = Date.parse(value);
-      if (!isNaN(value)){
+      if (!isNaN(value)) {
         return value;
       }
       var timespan = (function(input) {
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1305,10 +1305,26 @@ const TYPE = module.exports.TYPE = {
       if (value instanceof Date) {
         return value;
       }
-      value = Date.parse(value);
-      if (isNaN(value)) {
+      var timespan = (function(input) {
+        var regex = /^([0-9]{1}|(?:0[0-9]|1[0-9]|2[0-3])+):([0-5]?[0-9])(?::([0-5]?[0-9])(?:\.(\d{1,9}))?)?$/;
+        var result = (regex.exec(input) || []).slice(1, 5);
+
+        if (result[0] == null) {
+          return null;
+        }
+
+        return {
+          hours: result[0] || 0,
+          minutes: result[1] || 0,
+          seconds: result[2] || 0,
+          milliseconds: result[3] || 0
+        };
+      })(value);
+
+      if (timespan == null) {
         return new TypeError('Invalid time.');
       }
+      value = new Date(0, 0, 0, timespan.hours, timespan.minutes, timespan.seconds, timespan.milliseconds);
       return value;
     }
   },
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1305,9 +1305,9 @@ const TYPE = module.exports.TYPE = {
       if (value instanceof Date) {
         return value;
       }
-      value = Date.parse(value);
-      if (!isNaN(value)) {
-        return value;
+      var dateValue = Date.parse(value);
+      if (!isNaN(dateValue)) {
+        return dateValue;
       }
       var timespan = (function(input) {
         var regex = /^([0-9]{1}|(?:0[0-9]|1[0-9]|2[0-3])+):([0-5]?[0-9])(?::([0-5]?[0-9])(?:\.(\d{1,9}))?)?$/;


sant123:
@arthurschreiber, could you review this pull request?

Thanks :)

src/data-type.js
```diff
@@ -1305,10 +1305,30 @@ const TYPE = module.exports.TYPE = {
       if (value instanceof Date) {
         return value;
       }
-      value = Date.parse(value);
-      if (isNaN(value)) {
+      var dateValue = Date.parse(value);
```
arthurschreiber: We're trying to move to ES2015, so we prefer `let` and `const` over `var`.

src/data-type.js
```diff
@@ -1305,10 +1305,30 @@ const TYPE = module.exports.TYPE = {
...
+      if (!isNaN(dateValue)) {
+        return dateValue;
+      }
+      var timespan = (function(input) {
```
arthurschreiber: Is there any specific reason why this is in a nested function? Nesting functions like this might have an adverse impact on performance. I think the logic in here could easily be part of the enclosing function.

sant123:
What I'm trying to do is that everyone can enter a time in string format like:

```javascript
request.input('input_parameter', sql.Time, "14:42:10");
```

Currently, if some one writes the code above; it will get an exception of this type:

```error: RequestError: Validation failed for parameter 'input_parameter'. Invalid time.```

Or this error in the Error object:

```javascript
{ [RequestError: Validation failed for parameter 'input_parameter'. Invalid time.] name: 'RequestError', message: 'Validation failed for parameter \'input_parameter\'. Invalid time.', code: 'EPARAM', number: 'EPARAM', lineNumber: undefined, state: undefined, class: undefined, serverName: undefined, procName: undefined, precedingErrors: [] }
```

The only way to avoid this error is doing this:

```javascript
request.input('input_parameter', "14:42:10");
```

As you see, we need to remove the sql type from the method in order to work.

The regex in the PR validates if the input string match with a valid 'timespan'. In this case it can accept:

``` hh:mm ```
``` hh:mm:ss ```
``` hh:mm:ss.mmm ```

According with SQL Server timespan type, milliseconds can only have 9 numbers of length for a valid timespan.

File: src/data-type.js (hunk 1 above)
Comment (sant123) [inline src/data-type.js]: Got it!

File: src/data-type.js (hunk 2 above)
Comment (sant123) [inline src/data-type.js]: No, there is not specific reason :sweat_smile: I just wrote it and exists as you see. But I can change that ;)

**Comment by arthurschreiber** [comment]
> What I'm trying to do is that everyone can enter a time in string format like:

I'm not sure whether I like this or not. Generally, I'm not a big fan of silent type conversions.

@tvrprasad @SaloniSonpal What do you think?

**Comment by sant123** [comment]
Well, the same happens in a string date @arthurschreiber, Javascript converts from a string to a datetime, in this case a string to a time.

**Comment by tvrprasad** [comment]
> Well, the same happens in a string date @arthurschreiber, Javascript converts from a string to a datetime, in this case a string to a time.

@sant123 The difference I see is that there is no custom parsing there. It's just using JavaScript native parser.

> @tvrprasad @SaloniSonpal What do you think?

@arthurschreiber I'm not a fan of the change either.

I think it's a question of programming model. Specifically whether you specify parameter values in native JavaScript or in SQL format or both. All the types currently support JavaScript native format only, I don't see any custom parsing in data-type.js currently. That makes this change inconsistent with the current model.

I think we should keep the API simple and only support JavaScript native way of specifying parameter values. We could make it explicit in the documentation.

My 2c :-)

**Comment by sant123** [comment]
Well, the main idea of this implementation was solve the stack overflow question and the mssql issue, currently JavaScript does not support timespan, but sql does... And here is when sql accepts timespan strings. In the method, we guarantee the input string be a real timespan, that's why type parameter is there... Isn't it? If none type is specified, then the sql server will try to solve it, if it fails an exception is raised. I think this is more a validation step, because as i said JavaScript does not support timespan.

**Comment by tvrprasad** [comment]
@sant123 The stack overflow developer could unblock by converting the string to Date object or prepend with "2000-01-01 " or something like that to make it a valid Date string.

I can see how it would be a convenience to do that for the developer, especially if he/she is importing the data from a different source. It still seems inconsistent that the 'value' can be either Date or a string that's essentially something arbitrary.

You can achieve what you're doing without custom parsing. I'll add a review comment on the code.

All things considered, I suppose there is a small overall benefit in allowing the user to be able to specify "hh:mm:ss". Especially if we can get rid of the custom parsing, please see my suggestion in code review comment.

@arthurschreiber may have more to say.

File: src/data-type.js [hunk 3]
```diff
@@ -1305,10 +1305,30 @@ const TYPE = module.exports.TYPE = {
       if (value instanceof Date) {
         return value;
       }
-      value = Date.parse(value);
-      if (isNaN(value)) {
+      var dateValue = Date.parse(value);
+      if (!isNaN(dateValue)) {
```
tvrprasad: I think you can replace with a simpler version that does not require custom parsing:

```javascript
if (!isNaN(dateValue)) {
  return dateValue;
}
dateValue = Date.parse("2000-01-01 " + value);
if (isNaN(dateValue)) {
  return new TypeError('Invalid date or timespan.');
}

return dateValue;
```

arthurschreiber:
So, I've been thinking a bit more about this:

* JavaScript does not have a `Time` class, only `Date`, which is actually more like a `DateTime`. Using a `Date` object when specifying a time feels a little weird, but it's how things work now.

* If we want to support converting from `String`s that contain a "valid" time, we should figure out what exactly we want to support. https://msdn.microsoft.com/en-us/library/bb677243.aspx contains a list of supported formats by other clients, maybe we want to make sure we support the same formats?

I like @tvrprasad's suggestion of (ab)using native `Date.parse` to parse the time portion. But we need to get this right, so tests are really important for this.

sant123:
@arthurschreiber, @tvrprasad I just reviewed the msdn link and the main format for parsing the time is ```hh:mm:ss[.nnnnnnn]```, however Javascript will only take the first 3 numbers from the milliseconds field like ```hh:mm:ss[.nnn]```. I like the @tvrprasad for parsing the string time.

tvrprasad:
Looks like there is this nanosecondDelta property you can set on the 'value' parameter you pass to addParameter() when it's of Date type. Search for nanosecondDelta  in https://github.com/tediousjs/tedious/blob/master/src/data-type.js for how it's consumed.

We'll have three ways to specify Time now:
Date type: Allows nano second delta with nanosecondDelta property.
INT type: Millisecond granularity only.
"hh:mm:ss.nnn": Millisecond granularity only.

I think we should be consistent across all forms of the 'value' parameter by way of supporting nano second granularity. I suppose that can be a different PR as 'Time' is not the only TYPE that has this inconsistency currently.

src/data-type.js
sant123: Mmmm it seems this validation will keep these inputs:

* "10:"
* "10:54:"

¿Is there a problem with it?

src/data-type.js
```diff
@@ -1305,10 +1305,30 @@ const TYPE = module.exports.TYPE = {
...
+        return dateValue;
+      }
+      var timespan = (function(input) {
+        var regex = /^([0-9]{1}|(?:0[0-9]|1[0-9]|2[0-3])+):([0-5]?[0-9])(?::([0-5]?[0-9])(?:\.(\d{1,9}))?)?$/;
+        var result = (regex.exec(input) || []).slice(1, 5);
```
jdanyow: Skipping the `slice` here would eliminate an array allocation.

Changing `regex.exec(input) || []` to `regex.exec(input) || [0,0,0,0,0]` would simplify the code below because `|| 0` wouldn't be needed.

PR #405, opened 2016-07-11
Avoid implicit clipping of valid BigInt strings when serializing

evocateur:
While using the `mssql` package, I found if I sent valid BigInt strings as sproc parameters they were getting mangled into incorrect values. For example, a parameter such as `'9223372036854775807'` would become `'-9223372036854775808'` when inserted into the database. It all boils down to an unguarded implicit casting of a String into a Number (when calling `Math.abs(num)`) that clips the valid BigInt string into a smaller number.

``` javascript
Math.abs('9223372036854775807')
// 9223372036854776000
```

This pull request is largely an update of the adapted node-int64 code to match the current implementation that respects the difference between `Number` and `String` parameters to the `numberToInt64LE()` helper function.

I switched to using [`biginteger`](https://github.com/silentmatt/javascript-biginteger) because it accepts a base when calling `toString()`, which makes porting the updated node-int64 algorithm relatively straightforward.


arthurschreiber:
Wow, first of all, thank you for this PR and the thorough implementation (including tests, yay!). ✨ 

I'll try to take a more in-depth look in the next days and get this merged asap.

evocateur:
Thanks, glad I could help!

src/tracking-buffer/bigint.js
```diff
@@ -66,15 +68,27 @@ function convertLEBytesToString(buffer) {
 module.exports.numberToInt64LE = numberToInt64LE;
 function numberToInt64LE(num) {
   // adapted from https://github.com/broofa/node-int64
-  const negate = num < 0;
-  let hi = Math.abs(num);
-  let lo = hi % 0x100000000;
-  hi = (hi / 0x100000000) | 0;
+  let hi, lo;
```
tvrprasad: This would only help for the case where the large integer value is specified as a string. For the cases, where the value is specified as a number, like below, the value is corrupted before execution gets here.
```JavaScript
commandBigInt.addParameter('Id', TYPES.BigInt, 9007199254740999);
```

https://github.com/tediousjs/tedious/pull/504 handles this by adding range validation with an error messages that suggests specifying the parameter as VarChar.

We have two options:

- Take this change and tweak the validation in https://github.com/tediousjs/tedious/pull/504 to only enforce when the specified value is a number. Any existing code using BigInt with large numbers will magically start working correctly.
- Discard this PR and just stick with blanket validation in https://github.com/tediousjs/tedious/pull/504 . This keeps the code and the story simple and alerts any existing code using large numbers with BigInt that there was data corruption.

I'm leaning towards the latter option.

evocateur:
Any existing code attempting to represent 64-bit integers in JavaScript as a Number is already broken, imo. Telling someone to change their DB schema to workaround a single language's constraint in a multi-language environment is a non-starter, to say nothing of the weirdness of an auto-incrementing index being stored as a string.

So anyway, I obviously prefer the former option: warn when attempting to use a clipped Number, but otherwise allow those who are using the string representation to have full parity with their database's capabilities.

This PR has been basically ignored for almost a year now, so I'm not holding my breath. I've already forked it internally, so by all means do whatever you want. I appreciate the effort in creating this helpful library; I wish the best for current and future maintainers.

tvrprasad:
> This PR has been basically ignored for almost a year now

Not anymore :-) We'd love for you to not have to work off of a fork, and get any other fixes you made in your fork into the main repository.

> Any existing code attempting to represent 64-bit integers in JavaScript as a Number is already broken, imo.

It's the same thing for code attempting to represent 64-bit integers in JavaScript as a string also, right? Without your patch I mean.

> Telling someone to change their DB schema to workaround a single language's constraint in a multi-language environment is a non-starter

Hmm, this should not require a change to DB schema, just a code change on the client I believe. The below code works for me, where Id is bigint in the DB schema. SQL Server seems to handle this just fine. Similar query in SSMS also works for me.

// JavaScript
```JavaScript
command.addParameter('Id', TYPES.VarChar, '9007199254740999');
```
// SSMS
<code>
INSERT INTO BigIntInNodeTest (Id, IsAtMax) VALUES ('9007199254740999', 0);
</code>

It's true that it's a little odd to specify VarChar in client code when the column is bigint in DB Schema. But then it's also unintuitive to specify a string with TYPES.BigInt in client code.

IanChokS:
Hi @evocateur,

I know it's been a while, but any chance you could update this PR to the latest master branch? That way we can move forward with this. 

Thanks!

evocateur:
@IanChokS Sorry, I don't have time for this anymore.

PR #407, opened 2016-07-18
Update documentation changelog and remove CoffeeScript notice

Nedlinin:
The changelog was pretty out of date so I took a quick stab at updating it.

I also removed the index notice about the source of Tedious being in CoffeeScript as it has since been converted to ES2016.

Lastly to fix a Github Pages build/Jekyll error I converted from trying to use pygments (unsupported on Github Pages) to rouge.

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,7 +18,6 @@ title: Overview
 </h3>
 <p>
   Tedious is pure Javascript, as are its dependencies.
-  (Well, the source is <a href="http://jashkenas.github.com/coffee-script/">CoffeeScript</a>.)
   So it should run without change on any platform where Node is supported.
 </p>
 
--- a/_config.yml
+++ b/_config.yml
@@ -1,2 +1,2 @@
 destination: site
-pygments: true
+highlighter: rouge
--- a/changelog.html
+++ b/changelog.html
@@ -5,6 +5,149 @@ title: ChangeLog
 
 <h1>Change Log</h1>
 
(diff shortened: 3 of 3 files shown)


index.html
```diff
@@ -18,7 +18,6 @@
 </h3>
 <p>
   Tedious is pure Javascript, as are its dependencies.
-  (Well, the source is <a href="http://jashkenas.github.com/coffee-script/">CoffeeScript</a>.)
   So it should run without change on any platform where Node is supported.
```
rossipedia: Only suggestion I'd have is mention the minimum Node version required (since Tedious is using ES2015+)

Nedlinin: I was actually planning to do so... But, I'm not actually familiar enough with the build process of Tedious.

I believe you guys are using Babel under the hood to convert to ES5?  If so, the package,json is probably accurate and Node v0.10+ is probably suitable.  But, I haven't actually run any tests manually or automated to double check the accuracy of that.

arthurschreiber: Yes, you're right, we use babel and transpile when publishing a new release, so even users of Node 0.10 are able to use tedious. 👍

arthurschreiber:
This is absolutely awesome. I have started maintaining release notes on the GH release pages and forgot to also update the documentation website. I am hoping we can somehow move the whole documentation to be somehow auto-generated from the source code instead of maintaining it separately, but that is a bigger task outside of the scope of this PR!

PR #408, opened 2016-07-18
Updated README to point to the new organization URLs as required.

Nedlinin:
The README had links that now 404.  Updated them to the new URLs.  

Also updated the build badge for dependency status (david-dm) as it could not find the project under pekim anymore.

AppVeyor/Travis on the other hand only find it under pekim and not tediousjs so I left that one pointing to the old one...


arthurschreiber:
Awesome! ❤️

PR #409, opened 2016-07-19
Fixing broken links

cressie176:
Noticed some of the links in the README still point to 'pekim' rather than tediousjs, including the link to "more documentation" which currently 404s. I've fixed the ones I could, but left some of the badges that broke with the tediousjs org


arthurschreiber:
Many thanks for your contribution! ❤️ Unfortunately, #408 contained the same changes and was opened earlier, so I merged that one instead. Looking forward to more contributions from you in the future! 👍

Issue #414, opened 2016-08-01
Windows authentication requires domain to be upper case and won't accept FQDN


tvrprasad:
This is related to https://github.com/tediousjs/tedious/issues/246. Teasing this out as a separate issue for easy tracking.

Domain name in connection config only works with upper case.

// Works
domain: REDMOND
domain: REDMOND.CORP.MICROSOFT.COM

// Does not work
domain: redmond
domain: redmond.corp.microsoft.com

@arobson - FDQN seems to work for me.

Proposed Fix:
=========
Convert domain to upper case.

tvrprasad:
Fix merged to master.

Issue #246, opened 2015-02-12
Windows Authentication issue to SQL 2014


ryperl:
New to tedious and just trying to sort this out of getting a windows auth connection to SQL Server 2014- 

Tedious package version: 1.9.0
node version 12

// config setup 
var config = {
	userName: "User",
	password: "MyPass",
	server: "MyServer",
	domain: "MyDomian",
	options: {
		database: "myDb"
	},
	debug: {
		packet: true,
		data: true,
		payload: true,
		token: true,
		log: true
	}
};

I have verified that I am able to connect to my 1433 port, and for testing this is a SQL server running locally.

Stack Trace: 
D:\Development\Node\node-sql\index.js:25
                throw err;
                      ^
ConnectionError: Login failed. 
The login is from an untrusted domain and cannot be used with 
Windows authentication.
    at Parser.<anonymous> 
(D:\Development\Node\node-sql\node_modules\tedious\lib\connection.js:447:37)
    at Parser.emit (events.js:107:17)
    at Parser.nextToken 
(D:\Development\Node\node-sql\node_modules\tedious\lib\token\token-stream-parser.js:91:18)
    at Parser.addBuffer 
(D:\Development\Node\node-sql\node_modules\tedious\lib\token\token-stream-parser.js:68:17)
    at Connection.sendDataToTokenStreamParser 
(D:\Development\Node\node-sql\node_modules\tedious\lib\connection.js:879:35)
    at Connection.STATE.SENT_NTLM_RESPONSE.events.data 
(D:\Development\Node\node-sql\node_modules\tedious\lib\connection.js:226:23)
    at Connection.dispatchEvent 
(D:\Development\Node\node-sql\node_modules\tedious\lib\connection.js:742:59)
    at MessageIO.<anonymous> 
(D:\Development\Node\node-sql\node_modules\tedious\lib\connection.js:670:22)
    at MessageIO.emit (events.js:107:17)
    at MessageIO.eventData 
(D:\Development\Node\node-sql\node_modules\tedious\lib\message-io.js:56:12)

Any push in the right direction appreciated.

patriksimek:
Try this.

```javascript
var config = {
    userName: "User",
    password: "MyPass",
    server: "MyServer",
    domain: "MyDomian",
    options: {
        database: "myDb",
        encrypt: true
    }
};
```

arobson:
[...]

PR #417, opened 2016-08-03
NTLM authentication requires domain to be upper case 

tvrprasad:
This addresses the issue:
https://github.com/tediousjs/tedious/issues/414

The issues speaks to FDQN in addition to the pull request. But I couldn't repro the FDQN failure. It seems to work for me just fine.

@arthurschreiber @arobson  - This is a follow-up to our discussion on https://github.com/tediousjs/tedious/issues/246.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -41,6 +41,12 @@ class Connection extends EventEmitter {
     this.config = config;
 
     if (typeof (config.domain) === 'string') {
+      // Assigning 'config' passed in directly to 'this.config' will have
+      // 'config' and 'this.config' referring to the same object. Now modifying
+      // this.config.domain would also modify config.domain which would be an
+      // unexpected side effect on the input parameter. JSON serializing and
+      // deserializing creates a new object thus protecting the input parameter
+      // from side-effects.
       this.config = JSON.parse(JSON.stringify(config));
       this.config.domain = this.config.domain.toUpperCase();
     }
--- a/src/connection.js
+++ b/src/connection.js
@@ -41,13 +41,6 @@ class Connection extends EventEmitter {
     this.config = config;
 
     if (typeof (config.domain) === 'string') {
-      // Assigning 'config' passed in directly to 'this.config' will have
-      // 'config' and 'this.config' referring to the same object. Now modifying
-      // this.config.domain would also modify config.domain which would be an
-      // unexpected side effect on the input parameter. JSON serializing and
-      // deserializing creates a new object thus protecting the input parameter
-      // from side-effects.
-      this.config = JSON.parse(JSON.stringify(config));
       this.config.domain = this.config.domain.toUpperCase();
     }
 
--- a/src/connection.js
+++ b/src/connection.js
@@ -39,6 +39,12 @@ class Connection extends EventEmitter {
     super();
 
     this.config = config;
+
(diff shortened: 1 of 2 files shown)


tvrprasad:
@arthurschreiber @arobson  - Please take a look.

src/connection.js
```diff
@@ -39,6 +39,12 @@ class Connection extends EventEmitter {
     super();
 
     this.config = config;
+
+    if (typeof (config.domain) === 'string') {
+      this.config = JSON.parse(JSON.stringify(config));
```
arthurschreiber: I don't think this line here is quite right - what's the reason for serializing and deserializing the config here?

arthurschreiber:
One small nitpick here, other than that, I think this looks good! 👍

src/connection.js
tvrprasad: I did that protect the 'config' parameter being passed in. Assigning config directly to this.config and then modifying this.config.domain would essentially modify config.domain which I figured would be an unintended side-effect.

tvrprasad: I'll and send an update to the pull request with comments added... once I figure how :)

tvrprasad:
@arthurschreiber  Pushed a change with a comment on why I did the JSON serialize/de-serialize. Please take a look.

tvrprasad:
login7-payload-test seems to be failing on Node v6 on Travis CI. Failure is not related to my change as I only changed comments in the latest update.

I couldn't repro it with and without my changes on my Windows box with Node v6 as well as older versions.

arthurschreiber:
@tvrprasad Thanks for the explanation! 👍 

I agree that the current tedious behavior here is wrong, and we should not modify the object that got passed in. Unfortunately, I don't agree with the implementation that you proposed, as it's a "hack" in my eyes - it abuses JSON serialization to create a copy of the config object.

If you remove that change from the PR, I'll merge it. I'd rather fix the "not modifying the passed in config" issue in a separate PR. How does that sound?

arthurschreiber:
> login7-payload-test seems to be failing on Node v6 on Travis CI. Failure is not related to my change as I only changed comments in the latest update.
> 
> I couldn't repro it with and without my changes on my Windows box with Node v6 as well as older versions.

As mentioned in #432, this is because Node.JS 6.4.0 shipped with a bug. Once Node.js 6.5.0 gets released, this will be fixed. I'll pin the AppVeyor builds to 6.3.x for the time being.

tvrprasad:
@arthurschreiber Updated PR to not do serialize/de-serialize.

Opened issue https://github.com/tediousjs/tedious/issues/435 to protect config parameter being passed in a non-hacky manner. I don't know at the moment what the non-hacky way to do that may be. Please add comments to the new issue if you have suggestions. Thanks.

arthurschreiber:
Thanks! ❤️ Could you also merge the lastest changes from master?

tvrprasad:
@arthurschreiber Done - merged to latest from master.

tvrprasad:
@arthurschreiber Anything blocking from getting this committed to master? Thanks.

tvrprasad:
@arthurschreiber Can you merge this change? Keen to see my first commit :)

arthurschreiber:
Thanks @tvrprasad! ❤️

PR #419, opened 2016-08-06
Add `options.trustServerCertificate`.

### LINKED ISSUE #397: Unable to verify certificate/ca from server are valid

arthurschreiber:
This allows accepting or rejecting a connection to a server that presents an untrusted certificate.

This fixes #397 and #399.

---

@SimonHooker Sorry this has taken me so long. 😞 I took your commits, squashed them together, and fixed a few issues (both style wise as well as a "this won't actually do the right thing in case of an certification verification error"). Please take a final look, and if everything looks good, I'll merge this ASAP.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -121,6 +121,10 @@ class Connection extends EventEmitter {
       this.config.options.cryptoCredentialsDetails = {};
     }
 
+    if (this.config.options.trustServerCertificate === undefined) {
+      this.config.options.trustServerCertificate = true;
+    }
+
     if (this.config.options.useUTC == undefined) {
       this.config.options.useUTC = true;
     }
@@ -924,7 +928,7 @@ Connection.prototype.STATE = {
         }
       },
       tls: function() {
-        this.messageIo.startTls(this.config.options.cryptoCredentialsDetails);
+        this.messageIo.startTls(this.config.options.cryptoCredentialsDetails, this.config.options.trustServerCertificate);
         return this.transitionTo(this.STATE.SENT_TLSSSLNEGOTIATION);
       }
     }
--- a/src/message-io.js
+++ b/src/message-io.js
@@ -85,7 +85,7 @@ module.exports = class MessageIO extends EventEmitter {
     return this._packetSize;
   }
 
-  startTls(credentialsDetails) {
+  startTls(credentialsDetails, trustServerCertificate) {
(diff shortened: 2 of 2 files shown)


arthurschreiber:
@SimonHooker Also, the commit is still attributed to you! 👍

SimonHooker:
Hi there, looks good to me.

Life gets in the way sometimes hey, but at least now it's almost done which will let me properly secure my connections :)  Thanks for the help with this.

SimonHooker:
I don't suppose there's any chance of a minor version release today is there?  It would be great to get this implemented, but until it pulls through on npm I can't really get it into my build process without major changes.

arthurschreiber:
You can install the current master via `npm install git+https://github.com/tediousjs/tedious.git#master`

SimonHooker:
Hi there,  I did try that but it results in the following

```
module.js:327
    throw err;
    ^

Error: Cannot find module 'tedious'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (C:\dev\my_project\node_modules\mssql\lib\tedious.js:9:9)
    at Object.<anonymous> (C:\dev\my_project\node_modules\mssql\lib\tedious.js:1323:4)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
```

The version from a plain `npm install tedious` has the /lib and /node_modules folders however `npm install git:https://github.com/tediousjs/tedious.git#master` does not - can you think of anything else that may be needed to enable the `#master` version to properly build - the reason I asked for release is due to the fact that it appears to do a build between the Git repo and what is hosted by npm though I'll admit my understanding of precisely how npm is set up is minimal.

PR #420, opened 2016-08-08
Test only against SQLServer 2016.

arthurschreiber:
This changes the appveyor tests to run only against SQL Server 2016. But it also enables separate testing of all protocol versions.


arthurschreiber:
@tvrprasad These changes here allow running the tedious integration tests using a specific version of the TDS protocol, instead of the latest version supported by the sql server that is used for the tests.

The version can be specified using the `TEDIOUS_TDS_VERSION` environment variable. IIRC, older versions of the TDS protocol (`7_3_A`, `7_2` and `7_1`) are completely unusable with tedious right now. I'd be super grateful if you could take a look and maybe figure out what's going on there? 

We've had a few users complaining that they ran into issues when connecting to older SQL Server versions, getting the failing integration tests working again would be a great first step here. 👍

tvrprasad:
@arthurschreiber I'll take a look at this in the next few days or so.

Sent you a couple of PRs on other issues. If you're able to merge them (or give feedback), that'll be great. Thanks.

tvrprasad:
Just saw the test failures. Seems like ...args syntax is not supported in NodeJS 4.5 and NodeJS 5.12 but supported in older and newer versions. That's odd. I'm looking.

tvrprasad:
The rest parameters (i.e. the ...args) syntax is indeed not support in older versions of NodeJS. See http://node.green/ for the support status for different versions.

The reason tests are passing for 0.12 and 0.10 is because of this magic line in test/setup.js:
```JavaScript
if (require('semver').lt(process.version, '4.0.0')) {
  require('babel-register');
}
```
We have two options:
1) Change 4.0.0 to 6.0.0 in the above snippet.
2) Revert back to function() syntax from => syntax and rely on self=this trick.

I don't quite understand the implications of 1) at the moment. Please me know which option you like better and I'll send a PR. Thanks.

arthurschreiber:
I'm on a plane right now, I'll write up a longer response with all the
details later.

For now, please revert to the 'function' syntax for the cases where you
need the arguments object. Thanks. ❤️
tvrprasad <notifications@github.com> schrieb am Do. 6. Okt. 2016 um 19:26:

tvrprasad:
Responding from a plane, wow! Very much appreciate it :)
Reverted the function() to => change. Tests passing now.

arthurschreiber:
Thanks, this is great! Gonna merge this now. ❤️

arthurschreiber:
@tvrprasad We use a library called `babel` to convert the ES6/ES2016 source code of `tedious` to  ES5 when releasing a new version. This allows us to write "modern" JavaScript while supporting older Node versions like 0.10 and 0.12.

The lines you pointed out in the `test/setup.js` was intended to make sure we don't load babel when running the tests on node versions newer or equal to the current LTS version (that is, at the time of writing, 4.x). The idea here was to make sure we only use modern JS features that are supported in the current Node LTS.

I'm not 100% sure if that's the best approach - maybe we should simply embrace ES2016 as a whole?

tvrprasad:
Thanks for the explanation. That makes sense now.

> I'm not 100% sure if that's the best approach - maybe we should simply embrace ES2016 as a whole?

What in your mind is the value proposition in limiting which modern JS features we use?

The only thing I can think of is that using current Node LTS supported features would make it easier for developers to contribute to the project.

In build.coffee where we run the babel compiler, we're running it for all versions of node. Why is that? I'm guessing it's so we don't introduce modern syntax issues we didn't catch in tests but not sure.

tvrprasad:
One other topic. How do we manage dropping support for older versions of node.js? In fact v0.10 just fell out of maintenance this month per https://nodesource.com/blog/essential-steps-long-term-support-for-nodejs/. Do we stop supporting it for new releases of Tedious? Perhaps this is only of academic interest at this point but could be real if we start run into some issue that's only on v0.10.

arthurschreiber:
> One other topic. How do we manage dropping support for older versions of node.js? In fact v0.10 just fell out of maintenance this month per https://nodesource.com/blog/essential-steps-long-term-support-for-nodejs/. Do we stop supporting it for new releases of Tedious? Perhaps this is only of academic interest at this point but could be real if we start run into some issue that's only on v0.10.

I'd be in favour of dropping support for v0.10 and v0.12 (plus all iojs versions), not with the next release, but the one after. Do you think that makes sense?

arthurschreiber:
> In build.coffee where we run the babel compiler, we're running it for all versions of node. Why is that? I'm guessing it's so we don't introduce modern syntax issues we didn't catch in tests but not sure.

This happens so we can push a single package to npm, that is used across all supported versions of Node.js.

tvrprasad:
> I'd be in favour of dropping support for v0.10 and v0.12 (plus all iojs versions), not with the next release, but the one after. Do you think that makes sense?

That sounds reasonable to me.

> This happens so we can push a single package to npm, that is used across all supported versions of Node.js.

Ah. That makes perfect sense!

Issue #423, opened 2016-08-12
When trustServerCertificate=false, verification failures silently die


SimonHooker:
In message-io.js when @arthurschreiber  rolled up changes he replaced

```
messageIO.socket.destroy();
messageIO.emit('error', verifyError);
```

with 

```
this.socket.destroy(verifyError);
```

However this has the effect that a verification error fails without triggering any promise

SimonHooker:
On further investigation, not sure my initial impression was correct - reverting

PR #425, opened 2016-08-12
Issue #423 Making verifyError failures noisy again 

SimonHooker:


SimonHooker:
Maybe this isn't an issue after all, closing off for now

Issue #426, opened 2016-08-12
verifyError emits error code ESOCKET which node-mssql silently ignores


SimonHooker:
Relating to issue #423 it turns out the actual issue is that if the error emitted is of type code === ESOCKET, node-mssql then mostly sidesteps it instead of reporting the issue.

As such will adapt the error code to be something different for verify errors.

IanChokS:
@SimonHooker Sorry I'm having a hard time understanding what the issue is. If node-mssql is sidestepping an error emitted from Tedious, then perhaps it would be more appropriate to open an issue in node-mssql :)

PR #427, opened 2016-08-12
Issue #426 Making verifyError emit a non-ESOCKET error again

SimonHooker:
This is to allow node-mssql to properly report the error again as at present a verifyError silently
closes the socket without reporting any errors making debugging and verification impossible.

### DIFF
--- a/src/message-io.js
+++ b/src/message-io.js
@@ -99,7 +99,8 @@ module.exports = class MessageIO extends EventEmitter {
 
         if (verifyError) {
           this.securePair.destroy();
-          this.socket.destroy(verifyError);
+          this.socket.destroy();
+          this.emit('error', verifyError);
           return;
         }
       }


arthurschreiber:
I'm not quite sure I understand this change. There is no code in tedious that listens to `error` events that happen on `MessageIO` instances. The way `EventEmitter` instances work is that when no one is listening to `error` events, and an `error` event is dispatched, it will `throw` the error and bring down your whole application because the error can not be caught or handled. Am I missing something?

Also, the current code is doing exactly what `tls.connect` would be doing: https://github.com/nodejs/node/blob/e0b8dd59bcb44a775c7672ac235656c3bbb096f2/lib/_tls_wrap.js#L1074-L1075

SimonHooker:
Am on holiday atm so will confirm when i get home. But think this was due
to the verify certificate not actually exposing a verification failure
externally which resulted in the socket being silently closed instead of
being exposed to the app using tedious which would be desirable.

Intention is that if you don't trust the server certificate then a failure
to verify should emit an error event and the app using tedious should
expect that possibility and have appropriate handling else what is the
point of verifying in the first place?

Again will confirm when home but i think i was just seeing things silently
failing and so causing things to fall over down the chain. Maybe node-mssql
did not expect tedious to close the socket without emitting an error at
that point or something.

As I write this I wish i could remember why i still use node-mssql instead
of using tediousjs directly - possibly something i need to change on my end
:)

SimonHooker:
Hi @arthurschreiber 

This is mostly a node-mssql issue due to the fact that it seems to have handling in place that makes it side-step ESOCKET errors specifically in turn preventing it bubbling up the chain.

My better solution here is more likely to directly integrate tediousjs, ultimately it's upto you whether to integrate this or not of course however at this time without this change I cannot use Tedious - however have made changes on my own fork to allow me to make use of this change without it being in the main tedious branch.  Just thought that others may hit this same roadblock that I hit when trying to verify certificates.

I first raised this under https://github.com/tediousjs/tedious/issues/423 

As stated there the end result is that the failure triggers no promise thus rendering it unusable.

tvrprasad:
Tedious is doing the right thing here. Connection class is handling socket 'error' event and bubbling it up the application correctly, see Connection.socketError().

As @SimonHooker mentioned above this is a node-mssql issue if it's not handling ESOCKET errors correctly. Although mssql code handling ESOCKET the way it does seems quite intentional - https://github.com/patriksimek/node-mssql/blame/master/src/tedious.coffee#L242

Closing this, just part of house cleaning :-) Feel free to re-activate if there is more to discuss.

PR #430, opened 2016-08-17
Use a larger buffer for variable data.

nico-incubiq:
Hi,

Recently I started to get `RangeError: Offset is out of bounds` errors while starting my server, and it started just after I moved the server to a new address.
I debugged the issue, and found that the length of the content in variableData (see code change) is 263, which is too large for the allocated 200.
I hope this fix is fine for you and you can produce a new tagged version very soon.
Thank you very much.

Regards,

Nicolas

### DIFF
--- a/src/login7-payload.js
+++ b/src/login7-payload.js
@@ -161,8 +161,8 @@ module.exports = class Login7Payload {
       this.variableLengthsLength = (9 * 4) + 6 + (2 * 4);
     }
     const variableData = {
-      offsetsAndLengths: new WritableTrackingBuffer(200),
-      data: new WritableTrackingBuffer(200, 'ucs2'),
+      offsetsAndLengths: new WritableTrackingBuffer(400),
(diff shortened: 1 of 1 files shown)


nico-incubiq:
For the record, there seems to be no issue on NodeJS 5.9.1 which I tried today, but newer versions trigger the ArrayOutOfBound exception. Maybe these newer versions check array overflows more strictly...

arthurschreiber:
Is this related to https://github.com/tediousjs/tedious/issues/431?

nico-incubiq:
Yes it definitely is!
Good news if the node guys are going to fix this on their own.

If you think there is no reason for the connection data to be larger than 200 characters in any situation then you can close the pull request.

Thanks.

arthurschreiber:
As this is a bug in Node 6.4.0 and will be fixed in 6.4.1, I'll close this. Many thanks for putting in the work here anyway! ❤️

Issue #431, opened 2016-08-17
NodeJS 6.4.0 Empty String Range Error


spanditcaa:
We are seeing errors (stack below) from tedious in our API under node 6.4 due to this new bug:

https://github.com/nodejs/node/issues/8127

It sounds like this will be fixed in node, a fix isn't necessary in tedious but I wanted to share the finding.

````
RangeError: Offset is out of bounds
    at RangeError (native)
    at Buffer.write (buffer.js:778:21)
    at WritableTrackingBuffer.writeString (\node_modules\tedious\lib\tracking-buffer\writable-tracking-buffer.js:216:38)
    at Login7Payload.addVariableDataString (\node_modules\tedious\lib\login7-payload.js:228:25)
    at Login7Payload.createVariableData (\node_modules\tedious\lib\login7-payload.js:194:12)
    at new Login7Payload (\node_modules\tedious\lib\login7-payload.js:111:25)
    at Connection.sendLogin7Packet (\node_modules\tedious\lib\connection.js:602:21)
    at Connection.message (\node_modules\tedious\lib\connection.js:1054:16)
    at Connection.dispatchEvent (\node_modules\tedious\lib\connection.js:519:45)
    at MessageIO.<anonymous> (\node_modules\tedious\lib\connection.js:439:23)
    at emitNone (events.js:86:13)
    at MessageIO.emit (events.js:185:7)
    at ReadablePacketStream.<anonymous> (\node_modules\tedious\lib\message-io.js:92:15)
    at emitOne (events.js:96:13)
    at ReadablePacketStream.emit (events.js:188:7)
    at readableAddChunk (\node_modules\tedious\node_modules\readable-stream\lib\_stream_readable.js:198:18)
    at ReadablePacketStream.Readable.push (\node_modules\tedious\node_modules\readable-stream\lib\_stream_readable.js:157:10)
    at ReadablePacketStream.Transform.push (\node_modules\tedious\node_modules\readable-stream\lib\_stream_transform.js:123:32)
    at ReadablePacketStream._transform (\node_modules\tedious\lib\message-io.js:57:16)
    at ReadablePacketStream.Transform._read (\node_modules\tedious\node_modules\readable-stream\lib\_stream_transform.js:159:10)
    at ReadablePacketStream.Transform._write (\node_modules\tedious\node_modules\readable-stream\lib\_stream_transform.js:147:83)
[...]

PR #432, opened 2016-08-25
Dummy change to trigger Travis-CI.

tvrprasad:
Just to validate the tests as this is breaking on a trivial Pull Request on an unrelated test.

### DIFF
--- a/src/ntlm-payload.js
+++ b/src/ntlm-payload.js
@@ -1,3 +1,5 @@
+// DUMMY CHANGE TO TRIGGER TRAVIS CI. WILL NOT BE CHECKED IN.
+
 'use strict';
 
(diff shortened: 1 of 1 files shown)


tvrprasad:
@arthurschreiber @pekim Travis-ci seems to fail login7-payload-test on a dummy change. Any ideas? I couldn't repro this on my Windows box for Node 6.4 (version for which the test seems to fail)

arthurschreiber:
This is related to https://github.com/tediousjs/tedious/issues/431, and is a bug in Node.js 6.4.0. I don't want to have a fix for this in tedious, as it looks like a new version of Node.js that fixes this issue will be released soon (probably today).

SaloniSonpal:
@arthurschreiber Good to know!

PR #440, opened 2016-09-04
Improve performance of queries that select GUID columns.

### LINKED ISSUE #429: why does arrayToGuid use toUpperCase?
Code:  https://github.com/tediousjs/tedious/blob/master/src/guid-parser.js#L12-L35

Excerpt from [RFC 4122](https://tools.ietf.org/html/rfc4122#section-3) (emphasis mine):

>Declaration of syntactic structure:
  A UUID is an identifier that is unique across both space and time,
  with respect to the space of all UUIDs.  Since a UUID is a fixed
  size and contains a time field, it is possible for values to
  rollover (around A.D. 3400, depending on the specific algorithm
  used).  A UUID can be used for multiple purposes, from tagging
  objects with an extremely short lifetime, to reliably identifying
  very persistent objects across a network.
>
>  The internal representation of a UUID is a specific sequence of
  bits in memory, as described in Section 4.  To accurately
  represent a UUID as a URN, it is necessary to convert the bit
  sequence to a string representation.
>
>  Each field is treated as an integer and has its value printed as a
  zero-filled hexadecimal digit string with the most significant
  digit first.  **The hexadecimal values "a" through "f" are output as
  lower case characters** and are case insensitive on input.

An implementation without `toUpperCase` is approximately 400 times faster in my testing (chrome 51, windows 10, benchmark.js):  https://gist.run/?id=f3e57d94770c6e7f8d2f783e9ac67c6c .  I think we could fix the perf by improving the `formatHex` algorithm but that's a separate issue.

I'm guessing the `toUpperCase` call exists for a reason... would be great to know why. Changing the behavior would break many apps.  Maybe an option could be added to turn it off?

Looks like node-mssql copied the implementation:  https://github.com/patriksimek/node-mssql/blob/v3.3.0/src/tds.coffee#L103-L131

arthurschreiber:
This is based on an idea suggested by @jdanyow. Instead of converting individual bytes of the GUID to strings using `Number#toString` and uppercasing them later, we use a predefined map that allows us to lookup the uppercased hex representation of the individual bytes extremely fast. This easily improves the performance of the GUID string generation by 2500 times.

This fixes #429.

### DIFF
--- a/src/guid-parser.js
+++ b/src/guid-parser.js
@@ -1,37 +1,48 @@
 'use strict';
 
-function formatHex(number) {
-  let hex = number.toString(16);
-  if (hex.length === 1) {
-    hex = '0' + hex;
-  }
-  return hex;
-}
+const MAP = [
+  '00', '01', '02', '03', '04', '05', '06', '07', '08', '09', '0A', '0B', '0C', '0D', '0E', '0F',
+  '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '1A', '1B', '1C', '1D', '1E', '1F',
+  '20', '21', '22', '23', '24', '25', '26', '27', '28', '29', '2A', '2B', '2C', '2D', '2E', '2F',
+  '30', '31', '32', '33', '34', '35', '36', '37', '38', '39', '3A', '3B', '3C', '3D', '3E', '3F',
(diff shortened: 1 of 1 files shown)


spanditcaa:
@arthurschreiber version and npm publish with this change ?

Issue #363, opened 2016-02-11
Tedious modifies passed in `config` object


arthurschreiber:
Tedious currently modifies the `config` object that is passed into the `Connection` constructor. This is bad practice and causes a lot of problems, as this is quite unexpected.

See #356 for one such problem.

The simplest fix for this would be if `tedious` would create a deep copy of the passed in `config` object.

arthurschreiber:
@tvrprasad The "correct" way to fix this would be to create a copy of the config object (and any nested objects) using `Object.create`. This is also the pattern that Node.js core code is using as a solution for the same problem.

tvrprasad:
@arthurschreiber Submitted pull request - https://github.com/tediousjs/tedious/pull/445.

PR #445, opened 2016-09-14
Connection class constructor keeps a reference of and modifies passed…

tvrprasad:
… in config

object. This is bad practice as this can have unintended side effects down the
line.
- Added a helper function for making deep copies of objects.
- Added unit tests.
- Updated Connection constructor to do a deep copy of config object.
- Added a unit test.

This addresses the issue https://github.com/tediousjs/tedious/issues/363.


tvrprasad:
@arthurschreiber Please take a look.

tvrprasad:
Fixed unit test failures.

tvrprasad:
@arthurschreiber Ping. Please take a look.

arthurschreiber:
I'm really not a fan of using any sort of approach based on deep copying an object here, as that is going to introduce all sorts of potential incompatibilities down the road.

Node.js core generally uses `Object.create` to create new objects with the passed in configuration in this newly created object's prototype chain. Can you try to come up with something based on the same approach?

tvrprasad:
I tried Object.create() but it does not quite do what we want it to do. Object.create() creates a new object with the passed in object as prototype. Any changes you make to the source object will affect destination but changes to destination object won't affect source object. So config passed in to Connection constructor will be protected from changes made to Connection.config but Connection.config will not be protected from any changes to config that may be made outside of the Connection class.

I looked at a few Object.create() examples in Node.js but none of them really suit us from what I can tell. What's being passed in to Object.create() is a prototype object that should never change. I don't think we can or should make that assumption about config object being passed in to Connection constructor.

That deep copy code is the best option I could find searching on the web. If you have other directions or pointers, I'll explore.

Here is some sample JavaScript and corresponding output that illustrates why I think Object.create() does not work.

```JavaScript
source = {
  a: 'InitialValue',
};

destination = Object.create(source);

console.log('After Object.create:');
console.log('source.a: ', source.a);
console.log('destination.a: ', destination.a);
console.log('');

source.a = 'ChangeSource';

console.log('After source change:');
console.log('source.a: ', source.a);
console.log('destination.a: ', destination.a);
console.log('');

destination.a = 'ChangeDestination';

console.log('After destination change:');
console.log('source.a: ', source.a);
console.log('destination.a: ', destination.a);
console.log('');

console.log(source.isPrototypeOf(destination));
```

Output:
***
After Object.create:
source.a:  InitialValue
destination.a:  InitialValue

After source change:
source.a:  ChangeSource
destination.a:  ChangeSource

After destination change:
source.a:  ChangeSource
destination.a:  ChangeDestination

source.isPrototypeOf(destination):  true
***

tvrprasad:
Found two packages that'll work for us lodash and node-deepcopy. Tried both with my sample script from previous post and they work correctly. Should we take a dependency on one of them?

lodash brings bunch more stuff with but has tens of millions of downloads and should be more reliable. node-deepcopy is tiny but only hundredish downloads.

https://www.npmjs.com/package/lodash
Usage:
```JavaScript
....
cloneDeep = require('lodash/cloneDeep');
...
destination = cloneDeep(source);
```

Usage:
https://www.npmjs.com/package/node-deepcopy
```JavaScript
deepCopy = require('node-deepcopy/index').deepCopy;
...
destination = deepCopy(source);
```

tvrprasad:
Noticed that we already bring in lodash npm package. Changed DeepCopy to just be a wrapper over lodash cloneDeep so we can have some unit tests to catch any unexpected change in behavior in lodash.

Please take a look.

tvrprasad:
@arthurschreiber Can you please take a look and see if this is ok?

tvrprasad:
@arthurschreiber I'll merge this if you don't have objections.

arthurschreiber:
@tvrprasad Sorry it took me some while to get back to you on this.

I pushed a few changes that drop the `lodash` dependency again and instead of deep-copying the config, we now create a new config object and copy over the values we need.

This is a bit more code, but is more explicit and helps us to actually document which config options tedious supports, in a central place.

If you agree with this solution, feel free to merge this asap.

test/unit/tedious-test.js
```diff
@@ -23,3 +23,27 @@ exports.connection = function(test) {
...
+  };
+
+  var connection = new Connection(config);
+
+  test.notStrictEqual(connection.config, config);
+  test.notStrictEqual(connection.config.options, config.options);
+
+  // Test that we did not do a deep copy of the cryptoCredentialsDetails,
+  // as we never modify that value inside tedious.
+  test.strictEqual(connection.config.options.cryptoCredentialsDetails, config.options.cryptoCredentialsDetails);
```
tvrprasad: Do we want to pick and choose what we deep copy vs not based on implementation? It would seem like that'll introduce fragility in code.

And it's also not just about protecting the app from Tedious modifying, I think we also need to protect Tedious from the app modifying config.

tvrprasad:
@arthurschreiber I like how this is self-documenting but I also feel this is more fragile, it's easy to make a change that ends up doing a shallow copy of some member of config in future.

Is there some anti-pattern about doing deep-copy of config? One concern I could see is that we end up deep copying something that _should not_ be deep-copied, but that'd seem too broad as we'll never be able to any deep copy ever. I'm curious to understand why your don't like the deep-copy, as I'm sure I'll gain some insight :)

And btw, I'll get back with you on the release and documentation issues soon. Got busy with other stuff, but should be past that in a couple of days.

arthurschreiber:
> Is there some anti-pattern about doing deep-copy of config? One concern I could see is that we end up deep copying something that should not be deep-copied, but that'd seem too broad as we'll never be able to any deep copy ever. I'm curious to understand why your don't like the deep-copy, as I'm sure I'll gain some insight :)

Ah, basically, it's personal preference on one hand, and on the other hand, I'm thinking about switching to either TypeScript or Flow in the future, where we will want to have an explicit `Config` type, instead of just a deep copy of whatever object the user passed in to the constructor.

Also, I found out about a bunch of config values that did not have defaults set through this refactoring.

test/unit/tedious-test.js
tvrprasad: > Do we want to pick and choose what we deep copy vs not based on implementation? It would seem like that'll introduce fragility in code.
And it's also not just about protecting the app from Tedious modifying, I think we also need to protect Tedious from the app modifying config.

@arthurschreiber Any thoughts on this particular code review comment?

arthurschreiber: Ah, yes, sorry, I missed this.

`cryptoCredentialsDetails` is not a tedious specific option, it's just passed forwards to https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options. I don't think we need to care about what happens to this object, and if it get's modified by Node.js itself, that is a bug in Node.js that should be fixed there.

tvrprasad: My comment was not so much about Node.js modifying the object, rather the application modifying it. Something like:

```JavaScript
... setup config ...
config.cryptoCredentialsDetails = { some stuff. }
var conn1 = new Connection(config);
... stuff with conn1...

... change a few fields and reuse config....
config.server = someotherserver;
config.cryptoCredentialsDetails = { some other stuff }
var conn2 = new Connection(config);

// Now you got both conn1 and conn2 operating with cryptoCredentialsDetails from the second config.
```
It would seem like that's an easy mistake to make. Should we protect the user application (not Node.js) from Tedious changing the config, as well as protect Tedious from user application change config?

PR #460, opened 2016-10-02
Tds multi version

tvrprasad:
@arthurschreiber This should move https://github.com/tediousjs/tedious/pull/420 forward. This PR fixes several of the test failures against older versions of TDS. Please merge the changes if they look good.

This does not fix all the failures. There are still 4 tests failing in transactions-test.coffee against 7.1 only. Not sure yet if these tests can be made 7.2 and above only or if these are bugs.


arthurschreiber:
These are looking great! Many thanks! ❤️

PR #463, opened 2016-10-03
Drop the remaining CoffeeScript files

arthurschreiber:
Replaces the remaining CoffeeScript files with JavaScript.


arthurschreiber:
This is not perfect - I think most of the comments got lost in the conversion and the code generated by the coffee-script compiler is quite ugly. It will also probably break many of the existing PRs that might modify one of these files here.

But it will allow us to drop the dependency on coffee-script, which has been on my list for a loooong time.

tvrprasad:
I assume that you're not expecting that I review all these files :)

Some questions, primarily out of curiosity:
- Should we preserve history by using git mv for these files? That'll also allow for a quick comparison before commit.
- What's the motivation for converting the tests from coffee script to js?
- Why were the tests writing in coffee script in the first place?
- Did you do the conversion manually or did you use some tool help?

arthurschreiber:
> I assume that you're not expecting that I review all these files :)

No, definately not - this contains way too many changes for anyone review. 😄 

> Should we preserve history by using git mv for these files? That'll also allow for a quick comparison before commit.

git by itself never tracks renames, even if you use `git mv` or something similar. Rename detection is done "after the fact" and is part of `git diff` and is based on the file contents, solely. As the changes here are complete rewrites (in the eyes of git's rename detection), git will never be able to see these as renames or preserve the history here.

> What's the motivation for converting the tests from coffee script to js?

CoffeeScript has once been very common, but nowadays it's more a burden than help. I don't want people to have to learn CS to be able to contribute to `tedious`. 

> Why were the tests writing in coffee script in the first place?

All of `tedious` was once written in CoffeeScript. I already migrated the files in the `src` directory, but haven't touched the `test` files so far.

> Did you do the conversion manually or did you use some tool help?

I performed the conversion through the `coffee` compiler (part of the `coffee-script` package) . Basically the output is exactly what was executed before (but that was only ever kept in-memory, never written to the file system). Unfortunately, the `coffee` compiler does ignore comments, and some of the comments are quite important. I'm not 100% convinced that this is the best approach.

tvrprasad:
> As the changes here are complete rewrites (in the eyes of git's rename detection), git will never be able to see these as renames or preserve the history here.

To preserve history, how about 'git mv test testhistory' and leave it there just for history?

> Unfortunately, the coffee compiler does ignore comments, and some of the comments are quite important. I'm not 100% convinced that this is the best approach.

I did a grep (or findstr as we call it on Windows :) across all the coffee files under tedious\test and I see there are 327 comments spread across 22 files. That doesn't sound so bad to move the comments to the js files if you think there is value there. I can offer help, we can divvy up the files if you like to port the comments. I searched for '#' in all the test files, here is the list of files with comments:

integration\bulk-load-test.coffee
integration\connection-test.coffee
integration\datatypes-in-results-test.coffee
integration\errors-test.coffee
integration\instance-lookup-test.coffee
integration\parameterised-statements-test.coffee
integration\prepare-execute-statements-test.coffee
integration\rpc-test.coffee
integration\transactions-test.coffee
integration\tvp-test.coffee
performance\big-select.coffee
unit\data-type.coffee
unit\login7-payload-test.coffee
unit\ntlm-payload-test.coffee
unit\token\colmetadata-token-parser-test.coffee
unit\token\env-change-token-parser-test.coffee
unit\token\infoerror-token-parser-test.coffee
unit\token\loginack-token-parser-test.coffee
unit\token\order-token-parser-test.coffee
unit\token\row-token-parser-test.coffee
unit\token\sspi-token-parser-test.coffee
unit\token\token-stream-parser-test.coffee

spanditcaa:
@tvrprasad @arthurschreiber if you need I'd be happy to help.

rossipedia:
I'm _really_ not a fan of CoffeeScript, and I have some spare cycles. I'll work on this over the next couple days (I actually started on this recently and just now found this PR).

It's not all that difficult, really. It's just...

```
( •_•)
( •_•)>⌐■-■
(⌐■_■)
```
_tedious_

arthurschreiber:
Closing this as #578 got rid of the remaining CoffeeScript files. 🎉

PR #465, opened 2016-10-06
Revert "Switch from function() style to => in transaction method call…

tvrprasad:
…backs."

Rest parameters syntax (i.e. the ...args syntax) is not supported in older
versions of NodeJS. Hence the revert.

This reverts commit cd7f4b684ab34e2098a2de4317a305fcc82e2e1e.


tvrprasad:
@arthurschreiber This is the revert we discussed in https://github.com/tediousjs/tedious/pull/420. I'm going ahead with the merge.

PR #476, opened 2016-11-09
Incorrect conditional statement

starspoint:
This will impact on all query if database name is not specified in their
query.


arthurschreiber:
Thanks! ❤️

PR #480, opened 2016-11-13
Verify the server's hostname against the certificate.

arthurschreiber:
Setting `trustServerCertificate`to `false` makes `tedious` check the server's certificate for validity, but it does not actually check that the hostname we tried connecting to is also listed in the certificate. This means the server side could present **any** valid certificate and `tedious` would happily connect to it. That's bad. 😞 

/cc @SimonHooker If you got some time, please take a look!
/cc @tvrprasad I want this to go into 1.15.0 as well.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1058,7 +1058,7 @@ Connection.prototype.STATE = {
         }
       },
       tls: function() {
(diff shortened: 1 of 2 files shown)


SimonHooker:
Looks good to me, feel a little embarrassed that I didn't think to check this, but ah well!

arthurschreiber:
Thanks @SimonHooker for taking a look! ❤️

PR #483, opened 2016-12-06
Do not remove handlers in 'connect' handler.

tvrprasad:
Current code does not remove handlers in 'connect' handler for
the socket over which we're going to communicate in
ParallelConnectionStrategy. But 'connect' handlers in other
places do so. The right behavior is to not remove the handlers,
especially the 'error' handler as errors can happen after a
successful connection is established. With the 'connect' handler
we could go either way.

### DIFF
--- a/src/connector.js
+++ b/src/connector.js
@@ -30,9 +30,6 @@ class Connector {
     };
 
     const onConnect = () => {
-      socket.removeListener('error', onError);
-      socket.removeListener('connect', onConnect);
-
       cb(null, socket);
     };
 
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Hey @tvrprasad,

I'm not sure I understand this correctly! 🤔 

The reason I added these initially is that `Connector` and the different connection strategies are not event based but callback based - either the connection can be successfully established, or not.

Before that callback is fired, we need to detach all `connect` and `error` listeners, otherwise we run risk of executing the callback twice (once when the `connect` event fires, and a second time when `error` fires later on), which breaks the contract of how `Connector` is supposed to behave.

As soon as the connection is established, the code that is using `Connector` needs to attach their own `error` handlers to correctly handle any errors that can happen on a socket after a successful connection was established.

But I might have missed something?

tvrprasad:
That makes sense. I was thinking these listeners are responsible for handling errors on socket for the lifetime of the socket, which doesn't make sense now that I think about it.

I'll close this PR. Thanks for the clarification!

PR #486, opened 2016-12-16
Windows integrated auth draft

tvrprasad:
@arthurschreiber I believe this is close to final form. Can I get some early feedback please? I just want to make sure I'm on track here. CI tests will fail as I don't have sspi-client package published yet. 

I tested this locally, opened 1600 simultaneous connections and sent queries to two different machines. Works just fine. You can see the simple test at - https://github.com/tvrprasad/sspi-client/blob/master/integration/sqlconnect-stress.js

If you got time to make another pass at https://github.com/tvrprasad/sspi-client, that'd be great. Got some more work before I can publish the npm package.

@arobson FYI. I'm able to take advantage of all the work done for password based NTLM support. I welcome any feedback.

I'm very much looking forward to getting this in :-)


tvrprasad:
Opened a PR without the 'draft' at the end, which makes it real :)
Please review and give feedback here:
https://github.com/tediousjs/tedious/pull/497

sanjeev2315:
will it support window based auth without password on Linux

tvrprasad:
@sanjeev2315 This only works on Windows. I plan to add Linux support after this gets merged. I closed this PR in favor of https://github.com/tediousjs/tedious/pull/497, please track that.

PR #493, opened 2017-01-20
Log isolation level string passed by user

harshithkashyap:
This PR adds a small change to log string based isolation level passed by user. 

Eg: 

```
connection.beginTransaction(cb, 'name',  'READ COMMITTED')
```

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -50,7 +50,7 @@ class Transaction {
     return {
       data: buffer.data,
       toString: () => {
-        return 'Begin Transaction: name=' + this.name + ', isolationLevel=' + isolationLevelByValue[this.isolationLevel];
+        return 'Begin Transaction: name=' + this.name + ', isolationLevel=' + (isolationLevelByValue[this.isolationLevel] || this.isolationLevel);
       }
     };
   }


tvrprasad:
Not sure I understand the purpose of this. Also, I believes this fail on transaction.isolationLevelToTSQL() in beginTransaction()

harshithkashyap:
@tvrprasad When you run tedious in debug mode, isolation level would be undefined if we pass `READ_COMMITTED` or any other string instead of `require('tedious').ISOLATION_LEVEL`.

Also, for tds < 7_2, string based isolationLevel already fails. https://github.com/tediousjs/tedious/blob/master/src/connection.js#L815-L824

We can add a default case to return user passed isolation level [here](https://github.com/tediousjs/tedious/blob/master/src/transaction.js#L107-L121)

tvrprasad:
@harshithkashyap Thanks for the clarification. I looked at the API documentation and I don't see any intent to support string based isolation level - http://tediousjs.github.io/tedious/api-connection.html. I'd say this is by design as of now.

One thing we should do is to add validation to the isolationLevel config value to flag an error if an invalid value, which includes strings, is passed. Opened an issue to track that - https://github.com/tediousjs/tedious/issues/510

harshithkashyap:
@tvrprasad I use tedious with SequelizeJS. Disabling string based isolation level would break a lot of existing code. Any reason why hex codes are used instead of strings [here](https://github.com/tediousjs/tedious/blob/master/src/transaction.js#L20-L27)?

tvrprasad:
@harshithkashyap I made another pass at the code and I don't see how string based isolation level could possibly work. All code paths to set transaction level hit either Transaction.isolationLevelToTSQL() or Connection.getIsolationLevelText() and neither does the right thing with a string. I took a look at the Sequelize code and I see that it's passing a string to beginTransaction. This looks to me like a bug in Sequelize.

Of course, I could be missing something. If you can put together a self-contained script that sets transaction level using a string, with or without Sequelize, I can take another look.

> Any reason why hex codes are used instead of strings here?

These hex codes are used directly at the protocol level. See:
https://github.com/tediousjs/tedious/blob/master/src/transaction.js#L46

tvrprasad:
@harshithkashyap I looked at this more and confirmed that Sequelize is passing a string to beginTransaction() which is a bug. I opened an issue in Sequelize for this https://github.com/sequelize/sequelize/issues/7296. I'll close this PR. Feel free to re-activate if you think I'm missing something.

PR #496, opened 2017-02-01
Drop support for node versions that fell out of LTS maintenance window.

tvrprasad:
Status of various versions of node may be seen at:
https://github.com/nodejs/LTS

Per that, versions v0.10, v0.12 and v5 fell out of maintenance mode. It
might make sense drop support for these versions for future releases of
Tedious.

Specific motivation for this is the Windows Integrated Authentication
work which has test failures due to bugs in those versions, and forcing a
hacky work-around that requires installing latest version of npm before
running the tests for each version of nodejs. See:
https://github.com/tediousjs/tedious/pull/486/commits/15a52b798f14ce90f2a60c2e0000b14b99d9fd97

This part of this PR: https://github.com/tediousjs/tedious/pull/486

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,7 @@ sudo: false
 before_install:
   - npm install -g npm
 node_js:
-  - "0.10"
-  - "0.12"
   - "4"
-  - "5"
   - "6"
   - "7"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,10 +2,7 @@ version: "{build}"
 
 environment:
   matrix:
-    - nodejs_version: "0.10"
-    - nodejs_version: "0.12"
     - nodejs_version: "4"
-    - nodejs_version: "5"
     - nodejs_version: "6"
     - nodejs_version: "7"
 


tvrprasad:
@arthurschreiber Thoughts please.

arthurschreiber:
This is something I'm hugely in favor of, but _after_ the horribly delayed release of 1.15.0.

As dictated by semver, the release that drops support for these versions will have to be labeled 2.0.0.

tvrprasad:
Sounds good. We'll can get Windows Integrated Authentication in after release of 1.15.0.
Please let me know if there anything I can do to help with release of 1.15.0.

arthurschreiber:
@tvrprasad Sounds great! I'd really love if we could increase the pace of releases after 1.15.0 is out of the door. I'm really feeling bad that this release is taking so long.

The last blocker for the 1.15.0 release is the missing UDPv6 support in https://github.com/tediousjs/tedious/pull/362. Adding the support should be pretty straightforward:
* We need a create second socket for UDPv6 around https://github.com/tediousjs/tedious/pull/362/files#diff-496305d6fe353ac0f107f60b35b814bcR73.
* When looking up the dns names at https://github.com/tediousjs/tedious/pull/362/files#diff-496305d6fe353ac0f107f60b35b814bcR73, we need to remove the parameter that limits the returned addresses to be IPv4 only.
* When iterating over the addresses in https://github.com/tediousjs/tedious/pull/362/files#diff-496305d6fe353ac0f107f60b35b814bcR85, we need to send a request on the correct udp socket based on the family of the current address.

If you can spend some time on this, that'd be great. Otherwise, I'll try to finish this up as soon as I can.

tvrprasad:
@arthurschreiber I'll spend some time on this later this week or early next week. Wrapping up Windows Integrated Authentication.

We'll get this past the finish line soon :-)

chdh:
> As dictated by semver, the release that drops support for these versions will have to be labeled 2.0.0.

@arthurschreiber Could you tell us which of the [semver](http://semver.org/) rules you refer to? The public Tedious API will not change when old Node runtime versions are no longer supported.

Rule 8:

> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the **public API**.

FAQ:

> What should I do if I update my own dependencies **without changing the public API**?
That would be considered compatible since it does not affect the public API.

arthurschreiber:
> Droping support of old Node versions does not justify incrementing the major version number.

I'm pretty sure it does. Dropping support of an older node.js version is a pretty backwards incompatible change.

I checked some other big node.js libraries and how they handled this upgrade:

* ESLint: Dropped support for 0.10 and 0.12 in the 3.0 release.
* Babel: Will drop support for 0.10 and 0.12 in the 7.0 release.
* Libraries part of LoopBack: Dropped support in a major release (see https://strongloop.com/strongblog/loopback-drops-support-for-node-0-10-and-0-12/).

I know there is also other libraries (like Ember.js) that dropped support in a minor or even patch release.

The issue with not bumping the major release is that everyone who has their tedious dependency specified in their `package.json` using a specifier like `^1.14.0` (which has been the default since nm 2 or so) will always be automatically upgraded to the latest `1.x` version of tedious when running `npm install` (or `npm update`). This will break their apps and require manual fixing if they still run on 0.12 and 0.10, which is not acceptable to me.

In semver release numbers are "cheap" anyway, so there is little reason for not staying on the side of "better safe than sorry".

Also, this does not mean we need to hold of on the 2.0.0 release for an arbitrary set of features, I'm totally fine with releasing 2.0.0 right after 1.15.0 with nothing else changed other than officially dropping node 0.10 and 0.12 support.

tvrprasad:
> I'm totally fine with releasing 2.0.0 right after 1.15.0 with nothing else changed other than officially dropping node 0.10 and 0.12 support.

Let's do that actually!

tvrprasad:
This got merged via https://github.com/tediousjs/tedious/pull/540

PR #500, opened 2017-02-10
Add option to set DateFirst

edwardsmit:
`DATEFIRST` is set to the default value of `7`, but there is no way to override this setting.
As I'm having the need to set this to an other value I created this PR.
I did not find any rules for PR's, so let me know when you need additional code in order to merge this PR.


edwardsmit:
Closed in favor of #503

PR #501, opened 2017-02-10
Make SET ARITHABORT configurable

carlosman:
I need the behavior afforded by the SET ARITHABORT ON setting, in order to use this driver, as part of patriksimek's node-mssql driver. I have been modifying this setting manually in the lib/connection.js file for a few versions now, every time I upgrade the node_modules for my product.

My changes leave the ARITHABORT default "off", as you currently have it, but I should also point out the recommendation from the RDBMS developer's [SET ARITHABORT (Transact-SQL) manual page](https://msdn.microsoft.com/en-us/library/ms190306.aspx), on the "Remarks" paragraph, states: "You should always set ARITHABORT to ON in your logon sessions. Setting ARITHABORT to OFF can negatively impact query optimization leading to performance issues."

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -135,7 +135,7 @@ class Connection extends EventEmitter {
         if (typeof config.options.enableArithAbort !== 'boolean') {
           throw new TypeError('options.enableArithAbort must be a boolean (true or false).');
         }
-        
+
         this.config.options.enableArithAbort = config.options.enableArithAbort;
       }
 
--- a/src/connection.js
+++ b/src/connection.js
@@ -61,6 +61,7 @@ class Connection extends EventEmitter {
         connectionIsolationLevel: ISOLATION_LEVEL.READ_COMMITTED,
         cryptoCredentialsDetails: {},
         database: undefined,
+        enableArithAbort: false,
         enableAnsiNullDefault: true,
         encrypt: false,
         fallbackToDefaultDb: false,
@@ -128,6 +129,10 @@ class Connection extends EventEmitter {
 
       if (config.options.enableAnsiNullDefault != undefined) {
         this.config.options.enableAnsiNullDefault = config.options.enableAnsiNullDefault;
+      }
+
+      if (config.options.enableArithAbort != undefined) {
+        this.config.options.enableArithAbort = config.options.enableArithAbort;
       }
 
       if (config.options.encrypt != undefined) {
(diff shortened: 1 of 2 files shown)


tvrprasad:
@arthurschreiber I can merge this if you don't have objections. This is a useful config option. The change looks safe and does not affect default behavior.

Based on the recommendation on MSDN page, perhaps we should change the default value for 'arithabort' to 'on' for 2.0 release. It'll be a breaking change. From the MSDN link https://msdn.microsoft.com/en-us/library/ms190306.aspx:

"
The default ARITHABORT setting for SQL Server Management Studio is ON. Client applications setting ARITHABORT to OFF can receive different query plans making it difficult to troubleshoot poorly performing queries. That is, the same query can execute fast in management studio but slow in the application. When troubleshooting queries with Management Studio always match the client ARITHABORT setting.
"

tvrprasad:
@carlosman Can you please add integration test coverage for this option to test\integration\connection-test.coffee?

tvrprasad:
@carlosman Thinking about this further, not sure if we need this as a global option. You can just set this in your SQL, right?
SET ARITHABORT ON;

I still think we should change the default to ON in version 2.0.

carlosman:
@tvrprasad It cannot be set at the SQL level, because once the connection is attempted with the default options set in _connection.js_, the error condition I am facing with my server immediately triggers. The **only** solution I have found for this problem is patching _connection.js_ after updating the _node-mssql_ module in each of my _Node.js_ web services nodes.

I will look into the integration test coverage portion of it. I am not that experienced with that, so I will appreciate any comments on that.

tvrprasad:
I can review your integration test changes.

carlosman:
@tvrprasad I added the integration tests on pull request #511. Sorry for the delay, but my company landed a new contract and I have been on an extreme rush. I hope you have time to review these tests. Thanks.

carlosman:
@tvrprasad As you requested, I added the changes to test/integration/connection-test.coffee to this pull request, and removed pull request #511.

**Thanks for helping me** understand how to do this and I hope these changes are appropriate to get this config option added.

test/integration/connection-test.coffee
```diff
@@ -1020,3 +1020,68 @@ exports.disableAnsiNullDefault = (test) ->
...
+	connection.on('connect', (err) ->
+		connection.execSql(request)
+	)
+
+	connection.on('end', (info) ->
+		test.done()
+	)
+
+# Test that ARITHABORT can be set to on
+exports.testArithAbortCanBeSetToOn = (test) ->
```
tvrprasad: These tests are almost identical, except one checks for '0' and the other for '1' for value of arithabort. Please refactor the common implementation. Something like:

function testArithAbortImpl(test, arithAbortSetting)
..... Implementation here. ....

exports.testArithAbortDefault = (test) -> 
    testArithAbortImpl(test, undefined)

exports.testArithAbortOff = (test) -> 
    testArithAbortImpl(test, false)

exports.testArithAbortOn = (test) -> 
    testArithAbortImpl(test, true)

carlosman:
@tvrprasad Done. I had to learn some CoffeeScript to make this tests, which may not be a bad thing ;-)

carlosman:
@tvrprasad The Travis CI build has passed, but it is still shown as "in progress" here.  Is there a way for me to refresh that?

tvrprasad:
@carlosman That's odd. I don't know how to re-run tests other than to publish a tweak. Seems like you have tabs instead of spaces in your coffee test changes. Change that to spaces and push the change. It'll fix the issue as well as re-trigger the tests.

src/connection.js
```diff
@@ -130,6 +131,10 @@ class Connection extends EventEmitter {
         this.config.options.enableAnsiNullDefault = config.options.enableAnsiNullDefault;
       }
 
+      if (config.options.enableArithAbort != undefined) {
+        this.config.options.enableArithAbort = config.options.enableArithAbort;
```
tvrprasad: Please add validation that enableArithAbort is of type Boolean. We don't want someone doing:
config.options.enableArithAbort = 'off' and have that treated as 'on'.
You can follow the pattern for validation for 'config.server' near the top.
Please add a test. You can follow the pattern for the 'badPort' test.

Issue #477, opened 2016-11-12
Release 1.15.0


arthurschreiber:
I plan to release the next version of `tedious`, `1.15.0` "Alive and Kicking" soon.

Here's the changelog so far:

---

### What's New

* #362 `tedious` now has better support for SQLServer instances configured for High Availability! If the DNS name for the SQLServer instance you're connecting to resolves to multiple IP addresses, `tedious` will now sequentially try to connect to each of them, and will stop after the first successful connection. There's now also support for the `multiSubnetFailover` connection option, which allows to connect to all resolved IP addresses in parallel and will use whichever connection was established first. Thanks @chrislukkk!

* #419 When establishing an encrypted connection to a server, `tedious` never actually verified the certificate presented by the server. As this is probably not a good idea from a security standpoint, there is a new `trustServerCertificate` connection option that you can set to `false` to enforce certificate verification. The default behaviour did not change, but will in a future release. Thanks @SimonHooker!

### Fixed

* #461 We accidentally broke support for TDS 7.1 (used by SQLServer 2000) a few versions ago. If you're still running software from the year 2000, rejoice - TDS 7.1 support is back! Thanks @tvrprasad!

* #367 Encrypted connections and NTLM authentication didn't work together, at all - but now they do! Thanks @leerlemoniii and @mcrummey!

* #417 When authenticating via NTLM, the domain name was sent to the server as provided in the connection options. As it turns out, NTLM prefers domain names to be specified in upper case, so that's what we're doing now. Thanks @tvrprasad!

* #440 #443 Parsing and writing the contents of GUID columns was pretty slow. Now it's many times faster! Thanks @jdanyow!

* #445 The configuration object that was passed to the `Connection` constructor was modified in multiple places. That was pretty careless and unexpected, so `tedious` now will keep an internal copy of the configuration. Thanks @tvrprasad!

---

Did I miss anything important?

/cc @tvrprasad @SaloniSonpal @patriksimek

patriksimek:
Great contributions! Thanks for all your efforts.

tvrprasad:
This is great @arthurschreiber!
> Did I miss anything important?

Can we merge the unit tests for Multi Subnet Failover Feature (#362)
https://github.com/tediousjs/tedious/pull/444

jjg1914:
@arthurschreiber Is there any ETA on this being released? Would love to have the multiSubnetFailover feature!

arthurschreiber:
[...]

PR #503, opened 2017-02-13
Add option to set DateFirst

edwardsmit:
`DATEFIRST` is set to the default value of `7`, but there is no way to override this setting.
As I'm having the need to set this to an other value I created this PR.
I did not find any rules for PR's, so let me know when you need additional code in order to merge this PR.


tvrprasad:
@edwardsmit Not sure if this belongs as a global option. You can set this in your SQL, right?
SET DATEFIRST 7;
https://msdn.microsoft.com/en-us/library/ms181598.aspx

edwardsmit:
@tvrprasad Yes you can, but I find that rather ~~tedious~~ tiresome.
It's up to the client to set this, SSMS does this for you based on your locale, so you won't have to set it explicitly in each of your scripts. I think it's preferable to make sure that your client, once set, uses the exact desired DateFirst throughout wether you're using a single connection or create a connection for every call. Now you have to add boilerplate code where a simple option-config would suffice.

tvrprasad:
Can you please add an integration test to
test\integration\connection-test.coffee?

edwardsmit:
@tvrprasad Added the tests

arthurschreiber:
https://github.com/tediousjs/tedious/pull/508 also contains this option, but does not contain the tests from this.

edwardsmit:
@arthurschreiber Maybe @ariskemper can pull the test into his PR?

test/integration/connection-test.coffee
```diff
@@ -1020,3 +1020,55 @@ exports.disableAnsiNullDefault = (test) ->
...
+
+  connection.on 'connect', (err) ->
+    test.ifError(err)
+    connection.execSql(request)
+
+  connection.on 'end', (info) ->
+    test.done()
+
+# Test that the DATEFIRST setting can be changed via an optional configuration
+exports.testDatefirstDefault = (test) ->
```
tvrprasad: Test name duplicated. Needs to be renamed, perhaps testDatafirstNotDefault.
And yes, makes sense for @ariskemper to grab this test and merge into his PR.

Thanks for putting this together!

tvrprasad: Since these tests are identical, these can be refactored as below:

```JavaScript
function testDateFirstImpl(test, dateFirstSetting)
..... Implementation here. ....

exports.testDateFirstDefault = (test) ->
    testDateFirstImpl(test, undefined)

exports.testDateFirstNonDefault = (test) ->
    testDateFirstImpl(test, 3)
```

edwardsmit:
@tvrprasad Refactored tests

src/connection.js
```diff
@@ -126,6 +128,10 @@ class Connection extends EventEmitter {
         this.config.options.database = config.options.database;
       }
 
+      if (config.options.datefirst != undefined) {
+        this.config.options.datefirst = config.options.datefirst;
```
tvrprasad: Please add a range check here. You can follow the pattern for 'port' config option.
Please add a test. You can follow the pattern for 'badPort' test.
Sorry for not catching this sooner. Thanks.

tvrprasad:
@edwardsmit  I just merged changes from #501 by @carlosman for enableArithAbort that includes tests. Please rebase.

edwardsmit:
@tvrprasad 🔔

test/integration/connection-test.coffee
```diff
@@ -1106,3 +1105,48 @@ exports.badArithAbort = (test) ->
...
+  connection = new Connection(config)
+
+  request = new Request('select @@datefirst', (err) ->
+    test.ifError(err)
+    connection.close()
+  )
+
+  request.on('row', (columns) ->
+    dateFirst = columns[0].value
+    test.strictEqual(dateFirst, dateFirst || 7)
```
tvrprasad: Change variable to name something different than the parameter passed to the function, say datefirstReturned. As it stands, you're comparing the variable to itself.

Nit: I suggest changing parameter name from 'dateFirst' to 'datefirst' for consistency.

edwardsmit:
@tvrprasad 🛎

edwardsmit:
@tvrprasad Thanks for merging. Do you have a release planned in the near future so that I can use this in my projects?

tvrprasad:
Stay tuned to this - https://github.com/tediousjs/tedious/issues/477
Hopefully soon.

Issue #482, opened 2016-11-19
Closing connection while still in use causes undefined behavior


tvrprasad:
@arthurschreiber You mentioned this in https://github.com/tediousjs/tedious/issues/470#issuecomment-255947375. Do you have a fix in mind? I can take a crack at it some time.

arthurschreiber:
I think closing a connection with open requests should pass down some sort of error down the request's callback, so that user-defined code can take the appropriate steps of handling it.

tvrprasad:
@arthurschreiber Sent PR to address this - https://github.com/tediousjs/tedious/pull/505

PR #505, opened 2017-02-15
Invoke Request callback with error if connection closed prematurely.

tvrprasad:
This addresses https://github.com/tediousjs/tedious/issues/482

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -245,6 +245,11 @@ class Connection extends EventEmitter {
       } else {
         this.emit('rerouting');
       }
+      if (this.request) {
+        const err = RequestError('Connection closed before request completed.', 'ECLOSE');
+        this.request.callback(err);
+        this.request = void 0;
+      }
       this.closed = true;
       this.loggedIn = false;
       return this.loginError = null;
--- a/test/integration/connection-test.coffee
+++ b/test/integration/connection-test.coffee
@@ -601,6 +601,39 @@ exports.execBadSql = (test) ->
     #console.log(text)
   )
 
+exports.closeConnectionRequestPending = (test) ->
+  test.expect(1)
+
+  config = getConfig()
+
+  request = new Request('select 8 as C1', (err, rowCount) ->
+    test.ok(err)
+    test.strictEqual(err.code, 'ECLOSE')
+  )
+
+  connection = new Connection(config)
+
+  connection.on('connect', (err) ->
+    test.ifError(err);
+    connection.execSql(request)
+
+    # This should trigger request callback with error as there is
+    # request pending now.
+    connection.close()
+  )
+
+  connection.on('end', (info) ->
+      test.done()
+  )
+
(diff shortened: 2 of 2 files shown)


tvrprasad:
If someone wrote code like this, they will now get request callback with error where they didn't before:
```JavaScript
  // Any one of 'done', 'doneInProc' and 'doneProc'
  request.on('done', function (rowCount, more, rows) {
    connection.close();
  });
```
This is invalid code as TDS allows for multiple of these tokens and request is not complete till the callback passed in constructor is invoked.

I think we should not consider this a API breaking change, but calling it out nevertheless in case there is a difference of opinion on this.

arthurschreiber:
> This is invalid code as TDS allows for multiple of these tokens and request is not complete till the callback passed in constructor is invoked.

I believe that should be okay.

src/connection.js
```diff
@@ -245,6 +245,11 @@ class Connection extends EventEmitter {
       } else {
         this.emit('rerouting');
       }
+      if (this.request) {
+        const err = RequestError('Connection closed before request completed.', 'ECLOSE');
```
arthurschreiber: Could you change this to use `new` here?

```js
const err = new RequestError('Connection closed before request completed.', 'ECLOSE');
```

tvrprasad: Thanks for merging the change. Looks like we're using the non-new form of RequestError everywhere, so leaving as is.

neoadventist:
@tvrprasad @arthurschreiber hey guys, yeah this is breaking my application. any guidance on how to fix? 

```
   30 module.exports = {
   29         query:function(str){
   28                 return new Promise(function (resolve, reject) {
   27                         var connection  = new Connection(config);
   26                         connection.on('connect', function(err) {
   25                                 if(err){
   24                                         return reject(err);
   23                                 }
   22                                 var request = new Request(str, function(err, rowCount) {
   21                                         if (err) {
   20                                                 console.log(err);
   19                                                 return reject(err);
   18                                         } else {
   17                                                 console.log(rowCount + ' rows');
   16                                         }
   15                                 });
   14                                 request.on('done',function (rowCount, more, rows) {
   13                                         rows = rows.map(function(row){
   12                                                 return Object.keys(row).reduce(function(obj,key){
   11                                                         obj[key] = row[key].value;
   10                                                         return obj;
    9                                                 },{});
    8                                         });
    7                                         connection.close();
    6                                         return resolve(rows);
    5                                 });
    4                                 connection.execSqlBatch(request);
    3                         });
    2                 });
    1         }
  38  };
```

removing line seven seems to work, I just would like to know if not explicitly closing the connection will break anything. 

Thanks!

tvrprasad:
@neoadventist My comment up above speaks to this specific thing :-) You should not close the connection in 'done' callback, instead you should do it in Request callback. You should move line 7 to line 15.

neoadventist:
@tvrprasad thanks!

PR #514, opened 2017-02-23
Connection setup failing silently on TLS error.

tvrprasad:
If SQL Server is setup to require encrypted connections, client must
specifiy encrypt: true. When this is not so, there is no error emitted.
Fixed this to emit 'connect' event with error.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -652,6 +652,11 @@ class Connection extends EventEmitter {
     });
 
     if (preloginPayload.encryptionString === 'ON' || preloginPayload.encryptionString === 'REQ') {
+      if (!this.config.options.encrypt) {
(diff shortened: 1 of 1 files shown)


tvrprasad:
@arthurschreiber I'll merge if no objections. Change is simple. I ran into this issue myself and it's annoying when it happens.

Unfortunately no easy way to add an integration test as that'll require a separate instance of SQL Server configured to force encryption. I tested manually.

clark800:
Any chance you could cut an npm release with this fix? We're anxiously awaiting.

tvrprasad:
@clark800 This will be in the 1.15 release, hopefully soon. Stay tuned to https://github.com/tediousjs/tedious/issues/477

Issue #181, opened 2014-08-30
Pause Query Results


chacent:
when lot of rows will be received,i want to pause when sometime ,and then resume ,so that memory can in ctrl

tracker1:
+1

francesconero:
+1

sarus:
Is there anyway to do this in the latest Tedious which appears to use streams?  Can it be implemented using cursors with the existing release?

otopsy:
:+1:

JulianLaNeve:
@bretcope any updates on this?

bretcope:
@JulianLaNeve, @arthurschreiber would be the one to talk to about it at this point. I'm not very involved in Tedious these days.

chdh:
In the past 3 years, there have been many requests for the missing pause/resume/back-pressure functionality in Tedious:
#149, #186, #257, #271 

And there have been similar requests for the upper level driver node-mssql:
patriksimek/node-mssql#67, patriksimek/node-mssql#225, patriksimek/node-mssql#363

Back in 2015, the expectation was that #225, which uses Dissolve to parse the TDS stream, would solve the problem. But then #225 was replaced by #285 (Generators) and that was replaced by #314 (Callbacks).

What is the current state of this problem? I am considering to move from Java and C# to Node.js/TypeScript, not only for web applications, but also for utility programs that import/export/transfer large database tables. Without control over the flow of incoming data from the database, I cannot make this technological transition. There may be work-arounds like using OFFSET/FETCH in SQL, but they are too inconvenient. We need a way to apply back-pressure and control the flow of data for a query result.

In the old days, this feature was taken for granted. In Cobol, PL/1, Visual Basic, PHP, Java, C#, ... this was never a problem. ODBC, DAO, ADO, etc. all have this of course. In MS-Access with VBA e.g., you can open a forward-only recordset for a large table, read it record by record, and send the data slowly to e.g. an Excel sheet or another database, without having to worry that the memory would fill up in the background. Why isn't this nowadays possible with Node.js and SQL Server? Is it too early to adopt this technology?

chdh:
I have analyzed the data flow of the query result data:
```
Bytes:
  Socket.pipe(MessageIO.ReadablePacketStream)
  MessageIO.ReadablePacketStream/Transform.write()
  MessageIO.ReadablePacketStream._transform()
Packets:
  MessageIO.ReadablePacketStream/Transform.push()
  MessageIO.packetStream.on('data')
Messages:
[...]

Issue #512, opened 2017-02-23
Future Roadmap / Ideas


arthurschreiber:
Since the development on tedious has slowed down quite a bit in recent times, I want us (the tedious community) to put down a more or less formal roadmap of areas in tedious that need attention and should be worked on. This will help us maintainers to focus on these topics, while giving new contributors an idea of what they could be working on.

These topics don't have to necessarily be focusing on a specific bug or issue, but can be more general as well.

I will put down things that come to my mind in this issue, please feel free to suggest more topics. Once we have a good sized list, we can prioritize these and map out a rough timeline.

---

## General / Documentation

* **Better contributor guidelines**: We should make it clear what we expect from contributions (tests, coding style, etc). We should also document why some test cases are still written in CoffeeScript.

* **Better README**: The readme should better reflect the goals and history of tedious.

* **More Contributors**: Everyone contributes to ´tedious´ should get commit access.

## Documentation

* **Better documentation**: The documentation is currently stored in the `gh-pages` branch and maintained separately from everything else. We should move all documentation to be generated from the source code using `documentation.js` or a similar tool.

## Performance

* The parser is quite slow. This is mostly due to the fact that we use a lot of nested callbacks. I've done some experiments using a function trampoline instead, where each parsing function returns the next parsing function (or `undefined` if not enough data was received), with very promising performance benefits. Unfortunately, rewriting all of the token parsing functions is a lot of work.

* Many other areas of the code follow coding style practices that keep them from being JIT compiled by V8.

## Tests

* Tests are still written in CoffeeScript, and should be converted everything to ES2017 compatible code instead.

* `nodeunit` test cases are often really hard to debug and sometimes it's not quite clear where bugs happen or assertions fail. We probably should move over to ´mocha´ or ´jest´.

## Features

* ??

lloydcotten:
For test framework / runner... might I suggest a couple of more possibilities:

* [node-tap](http://www.node-tap.org/)
* [tape](https://github.com/substack/tape)
* [ava](https://github.com/avajs/ava)

This is just coming from my own experience of using mocha and finding that my tests tend to run faster and are cleaner with a TAP-style test framework (and supports concurrent test running).  I've personally used `node-tap` and `ava` and I like them both.  `node-tap` is more of framework that includes test coverage using nyc, etc.

Any of them (including `mocha` or `jest`) is better than `nodeunit` IMO.  I would be willing to contribute to this effort once clear guidelines and direction is chosen.

[...]

PR #518, opened 2017-03-01
Implements Request.pause() / resume()

chdh:
This pull request adds the two methods Request.pause() and Request.resume() to the public Tedious API. They can be used to pause and resume the flow of data rows from a query result. A test case is included.

See #181 for the reasons why this extension is important.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -577,7 +577,9 @@ class Connection extends EventEmitter {
         if (this.config.options.rowCollectionOnDone) {
           this.request.rst.push(token.columns);
         }
-        return this.request.emit('row', token.columns);
+        if (!(this.state === this.STATE.SENT_ATTENTION && this.request.paused)) {
+          this.request.emit('row', token.columns);
+        }
       } else {
         this.emit('error', new Error("Received 'row' when no sqlRequest is in progress"));
         return this.close();
@@ -641,6 +643,12 @@ class Connection extends EventEmitter {
       }
     });
 
+    this.tokenStreamParser.on('endOfMessage', () => {      // EOM pseudo token received
+      if (this.state === this.STATE.SENT_CLIENT_REQUEST) {
+        this.dispatchEvent('endOfMessageMarkerReceived');
+      }
+    });
+
     this.tokenStreamParser.on('resetConnection', () => {
       return this.emit('resetConnection');
     });
@@ -648,6 +656,12 @@ class Connection extends EventEmitter {
     this.tokenStreamParser.on('tokenStreamError', (error) => {
       this.emit('error', error);
       return this.close();
+    });
+
+    this.tokenStreamParser.on('drain', () => {
+      // Bridge the release of backpressure from the token stream parser
+      // transform to the packet stream transform.
+      this.messageIo.resume();
     });
   }
 
@@ -973,8 +987,31 @@ class Connection extends EventEmitter {
     }
   }
 
+  // Returns false to apply backpressure.
   sendDataToTokenStreamParser(data) {
     return this.tokenStreamParser.addBuffer(data);
+  }
+
+  // This is an internal method that is called from Request.pause().
+  // It has to check whether the passed Request object represents the currently
+  // active request, because the application might have called Request.pause()
+  // on an old inactive Request object.
+  pauseRequest(request) {
+    if (this.isRequestActive(request)) {
+      this.tokenStreamParser.pause();
+    }
+  }
+
+  // This is an internal method that is called from Request.resume().
+  resumeRequest(request) {
+    if (this.isRequestActive(request)) {
+      this.tokenStreamParser.resume();
+    }
+  }
+
+  // Returns true if the passed request is the currently active request of the connection.
+  isRequestActive(request) {
+    return request === this.request && this.state === this.STATE.SENT_CLIENT_REQUEST;
   }
 
   sendInitialSql() {
@@ -1266,6 +1303,7 @@ class Connection extends EventEmitter {
       }
 
       this.request = request;
+      this.request.connection = this;
       this.request.rowCount = 0;
       this.request.rows = [];
       this.request.rst = [];
@@ -1275,7 +1313,10 @@ class Connection extends EventEmitter {
       this.debug.payload(function() {
         return payload.toString('  ');
       });
-      return this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
+      this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
+      if (request.paused) {                                // Request.pause() has been called before the request was started
+        this.pauseRequest(request);
+      }
     }
   }
 
@@ -1563,6 +1604,7 @@ Connection.prototype.STATE = {
     name: 'SentClientRequest',
     exit: function() {
       this.clearRequestTimer();
+      this.tokenStreamParser.resume();
     },
     events: {
       socketError: function(err) {
@@ -1573,9 +1615,20 @@ Connection.prototype.STATE = {
       },
       data: function(data) {
         this.clearRequestTimer();                          // request timer is stopped on first data package
-        return this.sendDataToTokenStreamParser(data);
+        const ret = this.sendDataToTokenStreamParser(data);
+        if (ret === false) {
+          // Bridge backpressure from the token stream parser transform to the
+          // packet stream transform.
+          this.messageIo.pause();
+        }
       },
       message: function() {
+        // We have to channel the 'message' (EOM) event through the token stream
+        // parser transform, to keep it in line with the flow of the tokens, when
+        // the incoming data flow is paused and resumed.
+        return this.tokenStreamParser.addEndOfMessageMarker();
+      },
+      endOfMessageMarkerReceived: function() {
         this.transitionTo(this.STATE.LOGGED_IN);
         const sqlRequest = this.request;
         this.request = undefined;
--- a/src/connection.js
+++ b/src/connection.js
@@ -780,7 +780,7 @@ class Connection extends EventEmitter {
     }
 
     if (this.state && this.state.exit) {
-      this.state.exit.apply(this);
+      this.state.exit.call(this, newState);
     }
 
     this.debug.log('State change: ' + (this.state ? this.state.name : undefined) + ' -> ' + newState.name);
@@ -1602,9 +1602,12 @@ Connection.prototype.STATE = {
   },
   SENT_CLIENT_REQUEST: {
     name: 'SentClientRequest',
-    exit: function() {
+    exit: function(nextState) {
       this.clearRequestTimer();
-      this.tokenStreamParser.resume();
+
+      if (nextState !== this.STATE.FINAL) {
+        this.tokenStreamParser.resume();
+      }
     },
     events: {
       socketError: function(err) {
--- a/src/request.js
+++ b/src/request.js
@@ -7,10 +7,11 @@ module.exports = class Request extends EventEmitter {
     super();
 
     this.sqlTextOrProcedure = sqlTextOrProcedure;
-    this.callback = callback;
     this.parameters = [];
     this.parametersByName = {};
-    this.userCallback = this.callback;
+    this.canceled = false;
+    this.paused = false;
+    this.userCallback = callback;
     this.callback = function() {
       if (this.preparing) {
         this.emit('prepared');
@@ -133,5 +134,28 @@ module.exports = class Request extends EventEmitter {
       parameter.value = value;
     }
     return null;
+  }
+
+  // Temporarily suspends the flow of data from the database.
+  // No more 'row' events will be emitted until resume() is called.
+  pause() {
+    if (this.paused) {
+      return;
+    }
+    this.paused = true;
+    if (this.connection) {
+      this.connection.pauseRequest(this);
+    }
+  }
+
+  // Resumes the flow of data from the database.
+  resume() {
+    if (!this.paused) {
+      return;
+    }
+    this.paused = false;
+    if (this.connection) {
+      this.connection.resumeRequest(this);
+    }
   }
 };
--- a/src/token/token-stream-parser.js
+++ b/src/token/token-stream-parser.js
@@ -25,14 +25,36 @@ class Parser extends EventEmitter {
         this.emit(token.event, token);
       }
     });
+    this.parser.on('drain', () => {
+      this.emit('drain');
+    });
   }
 
+  // Returns false to apply backpressure.
   addBuffer(buffer) {
     return this.parser.write(buffer);
   }
 
+  // Writes an end-of-message (EOM) marker into the parser transform input
+  // queue. StreamParser will emit a 'data' event with an 'endOfMessage'
+  // pseudo token when the EOM marker has passed through the transform stream.
+  // Returns false to apply backpressure.
+  addEndOfMessageMarker() {
+    return this.parser.write(this.parser.endOfMessageMarker);
+  }
+
   isEnd() {
     return this.parser.buffer.length === this.parser.position;
+  }
+
+  // Temporarily suspends the token stream parser transform from emitting events.
+  pause() {
+    this.parser.pause();
+  }
+
+  // Resumes the token stream parser transform.
+  resume() {
+    this.parser.resume();
   }
 }
 module.exports.Parser = Parser;
--- a/test/integration/pause-resume-test.js
+++ b/test/integration/pause-resume-test.js
@@ -1,9 +1,7 @@
-// This module contains tests cases for the Request.pause()/resume() methods.
+const fs = require('fs');
 
 const Connection = require('../../src/connection');
 const Request = require('../../src/request');
-const fs = require('fs');
-const semver = require('semver');
 
 function getConfig() {
   const config = JSON.parse(fs.readFileSync(process.env.HOME + '/.tedious/test-connection.json', 'utf8')).config;
@@ -117,10 +115,8 @@ exports.testLargeQuery = function(test) {
   function verifyStreamStatesAfterPause() {
     const packetSize = connection.messageIo.packetSize();
     const socketRs = connection.socket._readableState;
-    if (semver.gte(process.version, '0.12.18')) {
-      test.ok(!socketRs.flowing,
-        'Socket is not paused.');
-    }
+    test.ok(!socketRs.flowing, 'Socket is not paused.');
+
     const minimalSocketFillTestLevel = 0x2000;             // (heuristic value)
     const highWaterReserve = 512;                          // (heuristic value)
     test.ok(socketRs.length >= Math.min(socketRs.highWaterMark - highWaterReserve, minimalSocketFillTestLevel),
--- a/test/integration/pause-resume-test.js
+++ b/test/integration/pause-resume-test.js
@@ -33,28 +33,27 @@ exports.tearDown = function(done) {
 //    receive all the rows.
 exports.testLargeQuery = function(test) {
   const debugMode = false;
-  const totalRows = 200000;                                // total number of rows to read
-  const delayTime = 1000;                                  // pause delay time in ms
+  // total number of rows to read
+  const totalRows = 200000;
+  // pause delay time in ms
+  const delayTime = 1000;
   const connection = this.connection;
-  let request;
   let rowsReceived = 0;
-  let failed = false;                                      // used to suppress further error messages
   let paused = false;
 
   connection.on('error', function(err) {
     test.ifError(err);
   });
-  openRequest();
 
-  function openRequest() {
-    const sql =                                            // recursive CTE to generate rows
-      'with cte1 as ' +
-        '(select 1 as i union all select i + 1 from cte1 where i < ' + totalRows + ') ' +
-      'select i from cte1 option (maxrecursion 0)';
-    request = new Request(sql, onRequestCompletion);
-    request.on('row', processRow);
-    connection.execSql(request);
-  }
+  // recursive CTE to generate rows
+  const sql = `
+    with cte1 as
+      (select 1 as i union all select i + 1 from cte1 where i < ${totalRows})
+    select i from cte1 option (maxrecursion 0)
+  `;
+  const request = new Request(sql, onRequestCompletion);
+  request.on('row', processRow);
+  connection.execSql(request);
 
   function onRequestCompletion(err) {
     test.ifError(err);
@@ -64,13 +63,14 @@ exports.testLargeQuery = function(test) {
 
   function processRow(columns) {
     if (paused) {
-      fail('Row received in paused state.');
+      test.ok(false, 'Row received in paused state.');
     }
     rowsReceived++;
+
     if (columns[0].value !== rowsReceived) {
-      fail('Invalid row counter value, value=' + columns[0].value + ', expected=' + rowsReceived + '.');
-      return;
+      test.ok(false, `Invalid row counter value, value=${columns[0].value}, expected=${rowsReceived}.`);
     }
+
     if (rowsReceived === Math.round(totalRows / 4)) {
       pause();
     }
@@ -121,15 +121,6 @@ exports.testLargeQuery = function(test) {
     test.ok(tokenTransformRs.length < packetSize / 3,
       'Token transform output buffer has large amount of data buffered.');
   }
-
-  function fail(msg) {
-    if (failed) {
-      return;
-    }
-    failed = true;
-    test.ok(false, msg);
-    connection.close();
-  }
 };
 
 // This test reads only a few rows and makes a short pause after each row.
@@ -146,10 +137,14 @@ exports.testLargeQuery = function(test) {
 exports.testTransitions = function(test) {
   const totalRequests = 4;
   const rowsPerRequest = 4;
-  const delayTime = 100;                                   // pause delay time in ms
-  const requestToCancel = 2;                               // 1-based position of request to be canceled
-  const rowToCancel = 2;                                   // 1-based position of row at which connection.cancel() will be called
-  const requestToStartPaused = 4;                          // 1-based position of request to start paused
+  // pause delay time in ms
+  const delayTime = 100;
+  // 1-based position of request to be canceled
+  const requestToCancel = 2;
+  // 1-based position of row at which connection.cancel() will be called
+  const rowToCancel = 2;
+  // 1-based position of request to start paused
+  const requestToStartPaused = 4;
   const connection = this.connection;
   let request;
   let requestCount = 0;
@@ -165,7 +160,7 @@ exports.testTransitions = function(test) {
   function openRequest() {
     let sql = 'select 1';
     for (let i = 2; i <= rowsPerRequest; i++) {
-      sql = sql + ' union all select ' + i;
+      sql = `${sql} union all select ${i}`;
     }
     request = new Request(sql, onRequestCompletion);
     request.on('row', processRow);
@@ -195,8 +190,8 @@ exports.testTransitions = function(test) {
   }
 
   function processRow(columns) {
-    test.ok(!canceled, 'Row received in canceled state, requestCount=' + requestCount + ' rowCount=' + rowCount);
-    test.ok(!paused, 'Row received in paused state, requestCount=' + requestCount + ' rowCount=' + rowCount);
+    test.ok(!canceled, `Row received in canceled state, requestCount=${requestCount} rowCount=${rowCount}`);
+    test.ok(!paused, `Row received in paused state, requestCount=${requestCount} rowCount=${rowCount}`);
     rowCount++;
     test.equal(columns[0].value, rowCount);
     pause();
--- a/test/integration/pause-resume-test.js
+++ b/test/integration/pause-resume-test.js
@@ -123,95 +123,79 @@ exports.testLargeQuery = function(test) {
   }
 };
 
-// This test reads only a few rows and makes a short pause after each row.
-// This test verifies that:
-//  - Pause/resume works correctly when applied after the last packet of a TDS
-//    message has already been dispatched by MessageIO.ReadablePacketStream.
-//    This is the case when EOM / packet.isLast() has already been detected
-//    at the time when Request.pause() is called.
-//    The 'message' event emitted by MessageIO has to be channeled through
-//    the token parser transform.
-//  - No more 'row' events are emitted after a paused request has been canceled.
-//  - The internal data flow is resumed after a paused request has been canceled.
-//  - A request can be paused before Connection.execSql() is called.
-exports.testTransitions = function(test) {
-  const totalRequests = 4;
-  const rowsPerRequest = 4;
-  // pause delay time in ms
-  const delayTime = 100;
-  // 1-based position of request to be canceled
-  const requestToCancel = 2;
-  // 1-based position of row at which connection.cancel() will be called
-  const rowToCancel = 2;
-  // 1-based position of request to start paused
-  const requestToStartPaused = 4;
-  const connection = this.connection;
-  let request;
-  let requestCount = 0;
-  let rowCount;
-  let paused = false;
-  let canceled = false;
-
-  connection.on('error', function(err) {
+exports.testPausedRequestCanBeCancelled = function(test) {
+  this.connection.on('error', (err) => {
     test.ifError(err);
   });
-  openRequest();
 
-  function openRequest() {
-    let sql = 'select 1';
-    for (let i = 2; i <= rowsPerRequest; i++) {
-      sql = `${sql} union all select ${i}`;
-    }
-    request = new Request(sql, onRequestCompletion);
-    request.on('row', processRow);
-    rowCount = 0;
-    paused = false;
-    canceled = false;
-    if (requestCount === requestToStartPaused - 1) {
-      pause();
-    }
-    connection.execSql(request);
-  }
+  const pauseAndCancelRequest = (next) => {
+    const sql = `
+      with cte1 as
+        (select 1 as i union all select i + 1 from cte1 where i < 20000)
+      select i from cte1 option (maxrecursion 0)
+    `;
 
-  function onRequestCompletion(err) {
-    requestCount++;
-    if (requestCount === requestToCancel) {
-      test.ok(err && err.code === 'ECANCEL');
-      test.equal(rowCount, rowToCancel);
-    } else {
-      test.ifError(err);
-      test.equal(rowCount, rowsPerRequest);
-    }
-    if (requestCount < totalRequests) {
-      openRequest();
-    } else {
+    const request = new Request(sql, (error) => {
+      test.ok(error);
+      test.strictEqual(error.code, 'ECANCEL');
+
+      next();
+    });
+
+    request.on('row', (columns) => {
+      if (columns[0].value == 1000) {
+        request.pause();
+
+        setTimeout(() => {
+          this.connection.cancel();
+        }, 100);
+      } else if (columns[0].value > 1000) {
+        test.ok(false, 'Received rows after pause');
+      }
+    });
+
+    this.connection.execSql(request);
+  };
+
+  pauseAndCancelRequest(() => {
+    const request = new Request('SELECT 1', (error) => {
+      test.ifError(error);
       test.done();
-    }
-  }
+    });
 
-  function processRow(columns) {
-    test.ok(!canceled, `Row received in canceled state, requestCount=${requestCount} rowCount=${rowCount}`);
-    test.ok(!paused, `Row received in paused state, requestCount=${requestCount} rowCount=${rowCount}`);
-    rowCount++;
-    test.equal(columns[0].value, rowCount);
-    pause();
-  }
+    request.on('row', (columns) => {
+      test.strictEqual(columns[0].value, 1);
+    });
 
-  function pause() {
-    paused = true;
-    request.pause();
-    setTimeout(afterDelay, delayTime);
-  }
+    this.connection.execSql(request);
+  });
+};
 
-  function afterDelay() {
-    if (requestCount === requestToCancel - 1 && rowCount === rowToCancel) {
-      canceled = true;
-      connection.cancel();
-    } else {
-      paused = false;
-      request.resume();
-    }
-  }
+exports.testImmediatelyPausedRequestDoesNotEmitRowsUntilResumed = function(test) {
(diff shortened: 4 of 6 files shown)


Congelli501:
I tried you branch, and I have the following error using this code:

```javascript
const Connection = require('tedious').Connection;
const Request = require('tedious').Request;

        server: '???',
        userName: '???',
        password: '???',
        options: {
                database: '???'
        }
});

connection.on('connect', (err) => {
        if (err) {
                throw err;
        }

        const request = new Request("select * from RANDOM_TABLE", (err2, rowCount) => {
                if (err2) {
                        throw err2;
                }

                console.log(rowCount + ' rows');
                connection.close();
        });

        request.on('row', (columns) => {
                console.log(columns.length);
                request.pause();
                setTimeout(() => {
                        request.resume();
                }, 500);
        });

        connection.execSql(request);
});
```

Error:
```
events.js:161
      throw er; // Unhandled 'error' event
      ^

Error: Received 'row' when no sqlRequest is in progress
    at Parser.<anonymous> (/root/node_tedious/node_modules/tedious/lib/connection.js:459:32)
    at emitOne (events.js:96:13)
    at Parser.emit (events.js:189:7)
    at Parser.<anonymous> (/root/node_tedious/node_modules/tedious/lib/token/token-stream-parser.js:54:15)
    at emitOne (events.js:96:13)
    at Parser.emit (events.js:189:7)
    at Parser.Readable.read (/root/node_tedious/node_modules/tedious/node_modules/readable-stream/lib/_stream_readable.js:391:26)
    at flow (/root/node_tedious/node_modules/tedious/node_modules/readable-stream/lib/_stream_readable.js:739:34)
    at resume_ (/root/node_tedious/node_modules/tedious/node_modules/readable-stream/lib/_stream_readable.js:722:3)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
```

The table contains 20 rows. I tried to pause the stream every 5 rows, but I had the same error.

I'm using node 7.6 on Ubuntu x64.
The SQL server has the following version:
```
Microsoft SQL Server 2005 - 9.00.1399.06 (Intel X86) 
        Oct 14 2005 00:33:37 
        Copyright (c) 1988-2005 Microsoft Corporation
        Developer Edition on Windows NT 6.1 (Build 7601: Service Pack 1)
```

Am I misusing the feature ?
Do you want me to open a bug request for that ?

Don't hesitate if you wan any more feedback / test.

chdh:
@Congelli501 Thanks for your bug report. I will analyze the problem and provide a fix.

chdh:
@Congelli501 I have fixed the problem. It happened when there were only a few records left when pause() was called. I have added another test case to verify that this problem is solved. Could you please test again?

arthurschreiber:
Hey @chdh,

Thank you very much for this PR - it definitely shows that pausing and resuming requests is possible  in `tedious` with a bit of tweaking here and there. 👍 

But as you mentioned over in #512:

> I think that the internal architecture of the TDS protocol driver should be reworked.

I think so as well. Tedious internals should be a bunch of streams that are piped into each other, so that `Request#pause()` would be implemented in terms of pausing the "last" stream in this pipe. If I understand the way Node.JS streams work correctly, this should in turn cause all other streams to be paused due to reaching the `highWatermark` and automatic backpressure handling kicking in.

Would you be interested in helping to transform the tedious internals for this? I have played with a few ideas for this, and we could open an issue to discuss this further. If not, I believe these changes could be merged as they are, I'm just concerned that doing so might make tedious internals even more complex (and fragile) than they already are. 🤔

I'm scheduling this for the release after the planned 1.15.0 release, as I don't want to delay the 1.15.0 release even further.

/cc @tvrprasad What do you think.

chdh:
@arthurschreiber, thanks for you comment. I will try to simplify the PR.

chdh:
I have simplified the PR to apply backpressure (pause/resume) only to the last stream (the token stream parser transform) and bridge the backpressure gap between the token stream parser transform and the packet stream transform. That way we have a "minimal invasive" patch with the data flow control concept we want in the long term.

ElfenLiedGH:
@chdh pause's great, but what do you think about timeout? should it pause too?
I take several record, then paused and then timeout, but query is Ok. 
I'm not sure that's right

arthurschreiber:
From the "Client Request Timer" section of the TDS specification:

> Controls the maximum time spent waiting for a query response from the server for a client request sent after the connection has been established.

This means that as soon as we received the first byte from the response message for a query, we need to stop the request timer. If the user wants some sort of time out on the processing of the data, it's up to them to define a more suitable timer.

chdh:
> as soon as we received the first packet from the response message for a query, we need to stop the request timer

You are right, but according to the documentation of [SqlCommand.CommandTimeout](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommand.commandtimeout.aspx) in the .NET API, the timeout can also occur between two consecutive rows.

> This property is the cumulative time-out (for all network packets that are read during the invocation of a method) for all network reads during command execution or processing of the results. A time-out can still occur after the first row is returned, and does not include user processing time, only network read time.
For example, with a 30 second time out, if Read requires two network packets, then it has 30 seconds to read both network packets. If you call Read again, it will have another 30 seconds to read any data that it requires.

I don't know how we could implement this correctly.

arthurschreiber:
> You are right, but according to the documentation of SqlCommand.CommandTimeout in the .NET API, the timeout can also occur between two consecutive rows.

Weird. This is not how `mssql-jdbc` is implemented (it also follows the spec). I'd suggest we stick to the side of the specification.

chdh:
@ElfenLiedGH please have a look at the new version. Does it solve your problem?

tvrprasad:
@chdh @arthurschreiber None of the Microsoft implemented drivers implement the notion of pause/resume. The reasoning as I understand is that you can't really slow down how fast the server sends data as there is no support this throttling in TDS specification. So the memory pressure is going to build one place or the other in the system. Is there something different about NodeJS that it makes sense here? Curious to know.

@arthurschreiber Agree with not delaying release of 1.15. I'll need to read up on streaming in NodeJS before I can share thoughts on it :-)

src/connection.js
```diff
@@ -522,7 +534,15 @@ class Connection extends EventEmitter {
     this.socket.on('close', this.socketClose);
     this.socket.on('end', this.socketEnd);
     this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-    this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
+    this.messageIo.on('data', (data) => {
+      this.clearRequestTimer();
```
tvrprasad: Can you do a separate PR for this bug fix? Though this got caught as part of this work, I think we can call it unrelated.

src/request.js
```diff
@@ -136,4 +136,26 @@ module.exports = class Request extends EventEmitter {
...
+      return;
+    }
+    this.connection.resumeTokenStreamParser();
+  }
+
+  // Returns true if this request is the currently active request of the connection.
+  isActive() {
+    return this.connection && this.connection.request === this && this.connection.state === this.connection.STATE.SENT_CLIENT_REQUEST;
+  }
 };
```
tvrprasad: This introduces pretty deep coupling between the Connection class and Request class. Perhaps this suggests pause/resume methods should be on the Connection class?

chdh:
@tvrprasad The need for pause/resume only arises in an asynchronous environment like Node.js. In the traditional programming environments with synchronous read-next calls, the data flow control is implicit. All Microsoft database drivers implement this implicit data flow control.

Example with MS-Access / VBA:

```
Dim rs As Recordset
Set rs = db.OpenRecordset("SELECT * FROM HugeTable", dbOpenForwardOnly)
Do While Not rs.EOF
  Debug.Print rs.Fields(0)
  ... wait a second ...
  rs.MoveNext
  Loop
```

In this example with MS-Access, the memory is not filled up, even if HugeTable has tens of millions of records. If you do the same thing with the current Tedious release, the memory fills up and after a few seconds the Node.js runtime will crash. Without pause/resume there is no way to control the data flow. The only solution is using SQL queries that don't generate a lot of data.

src/connection.js
chdh: Yes, I could put this into a separate PR, but it's not really a problem as long as you don't have a large result set or pause/resume. There was a real bug with clearRequestTimer() which I already fixed in PR #527.
@arthurschreiber what do you think?

src/request.js
chdh: If we have multiple active result sets (MARS) in the future, we need pause/resume per request. But you are right, the deep coupling is not clean. I will move this logic into the Connection class, but keep the pause/resume methods in the Request class.

It's a general problem that we cannot indicate public/protected/private with class methods and properties. Therefore is suggest to use TypeScript (see #512).

src/connection.js
chdh: OK, it's in #530 now.

tvrprasad:
> In this example with MS-Access, the memory is not filled up, even if HugeTable has tens of millions of records.

Isn't the memory being filled up at the network layer (perhaps the TCP buffers), in that case? I assume the server is still sending data as fast as it can.

arthurschreiber:
TCP has built in flow control and the sender will slow down sending data to
the receiver automatically.

tvrprasad:
> TCP has built in flow control and the sender will slow down sending data to
the receiver automatically.

Right, thought about that but it was not clear to me how that'd kick in here. The TCP layer on the client would have to stop sending ACKs while it's in the pause mode to throttle the server sending data. But then I thought that would cause TCP timeouts, retransmissions and ultimately errors on the server side. Now I see that the pause has to be really long for that to happen. Presumably server has mechanisms to handle slow clients. Makes sense now. Thanks :-)

Microsoft drivers like ADO.net support Async mode API but don't have pause/resume. So they'd be running into this issue, right? I wonder why there is not a demand for that support in those drivers.

chdh:
> Microsoft drivers like ADO.net support Async mode API but don't have pause/resume. So they'd be running into this issue, right? I wonder why there is not a demand for that support in those drivers.

It's another concept. In the .NET API, you have to call [DbDataReader.ReadAsync()](https://msdn.microsoft.com/en-us/library/hh223681.aspx) for each row. When you stop calling ReadAsync(), the row stream is paused implicitly. With Tedious and other Node libraries, the rows are continuously delivered via events and you have to call pause() explicitly.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,239 @@
...
+  const totalRows = 200000;                                // total number of rows to read
+  const delayTime = 500;                                   // pause delay time in ms
+  let connection;
+  let request;
+  let rowsReceived = 0;
+  let failed = false;                                      // used to suppress further error messages
+  let paused = false;
+  openConnection();
+
+  function openConnection() {
```
chdh: OK, I have refactored the pause/resume test cases with setUp/tearDown.

tvrprasad:
@chdh Thanks for the clarification! Certainly learnt some stuff from this thread :-) I'll send you a few more comments on this PR.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  function verifyStreamStatesAfterPause() {
+    const packetSize = connection.messageIo.packetSize();
+    const socketRs = connection.socket._readableState;
+    if (semver.gte(process.version, '0.12.18')) {
+      test.ok(!socketRs.flowing,
+        'Socket is not paused.');
+    }
+    test.ok(socketRs.length >= Math.min(socketRs.highWaterMark - 512, 0x4000),
```
tvrprasad: Please define consts for these numbers to describe what they represent.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+      'Socket does not feel backpressure.');
+    const packetTransformWs = connection.messageIo.packetStream._writableState;
+    const packetTransformRs = connection.messageIo.packetStream._readableState;
+    test.ok(!packetTransformRs.flowing,
+      'Packet transform is not paused.');
+    test.ok(packetTransformWs.length <= packetTransformWs.highWaterMark && packetTransformRs.length <= packetTransformRs.highWaterMark,
```
tvrprasad: Line too long. Perhaps you can break them into two assertions.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+      'Token transform is not paused.');
+    test.ok(tokenTransformWs.length <= tokenTransformWs.highWaterMark,
+      'Token transform input buffer overflow.');
+    test.ok(tokenTransformRs.length < packetSize / 3,
+      'Token transform output buffer has large amount of data buffered.');
+  }
+
+  function fail(msg) {
+    if (failed) {
+      return; }
```
tvrprasad: Formatting - move } to new line.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+    }
+    if (requestCount < totalRequests) {
+      openRequest();
+    } else {
+      test.done();
+    }
+  }
+
+  function processRow(columns) {
+    if (canceled) {
```
tvrprasad: line:193 validates rowCount for cancelled request but rowCount is not updated here in canceled state. Should it be? What's the expectation in terms of getting 'row' events after canceling a paused request?

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+    failed = true;
+    test.ok(false, msg);
+    connection.close();
+  }
+};
+
+// This test reads only a few rows and makes a short pause after each row.
+// The test verifies that:
+//  - Pause/resume works correctly when applied after the last packet of a TDS
+//    message has already been dispatched by MessageIO.ReadablePacketStream.
```
tvrprasad: I see a pause/resume after each row. Which part of the code validates this?

src/request.js
```diff
@@ -136,4 +136,19 @@ module.exports = class Request extends EventEmitter {
     }
     return null;
   }
+
+  // Temporarily suspends the flow of data from the database.
+  // No more 'row' events will be emitted until resume() is called.
+  pause() {
```
tvrprasad: Should we remove pause/resume on Request class and leave them on Connection class only? These seem roughly equivalent of cancel() which is on Connection class only.

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+    rowCount = 0;
+    paused = false;
+    canceled = false;
+    connection.execSql(request);
+  }
+
+  function onRequestCompletion(err) {
+    requestCount++;
+    if (requestCount == requestToCancel) {
+      test.ok(err && err.code == 'ECANCEL');
```
tvrprasad: Change '==' to '===' and '!=' to '!==' in the file?

test/integration/pause-resume-test.js
```diff
@@ -0,0 +1,246 @@
...
+    test.ok(false, msg);
+    connection.close();
+  }
+};
+
+// This test reads only a few rows and makes a short pause after each row.
+// The test verifies that:
+//  - Pause/resume works correctly when applied after the last packet of a TDS
+//    message has already been dispatched by MessageIO.ReadablePacketStream.
+//    (EOM / packet.isLast() has already been detected when pause() is called.
```
tvrprasad: nit: (EOM / packet.isLast() => (EOM / packet.isLast())

test/integration/pause-resume-test.js
tvrprasad: I suppose this is implicitly tested since there is a pause after the last row...

chdh:
@tvrprasad 

Only the two new methods Request.pause() and Request.resume() are intended to be part of the public Tedious API. All other changes are internal. The reasons that these two methods are members of the Request class and not of the Connection class are:

1. The 'row' events are emitted from the Request class and not from the Connection class. The pause/resume methods control these 'row' events. From the application point of view, the Request class represents the data source.

2. It's true that the cancel() method is currently a member of the Connection class. But that is because it has been designed from the implementation point of view and not from the application point of view. In the upper level driver node-mssql, the [cancel()](https://github.com/patriksimek/node-mssql#cancel) method is a member of the Request class. In the .NET API, the [Cancel()](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommand.cancel.aspx) method ist a member of the SqlCommand class. It's not the connection that is beeing canceled, it's the request / command. Therefore the cancel() method should not be a member of the Connection class.

3. If we later have multiple active requests at the Tedious API level (MARS or connection pooling), it must be clear which request is affected by the pause/resume call. The application logic uses the Request object as a handle for the data source and should not be forced to know the associated Connection object for using data flow control.

It's true that with the current implementation only a single Request object can be active at a time. But thats the internal view and we always have to think both sides, the internal implementation side and the external API user / application side. The application might still have old instances of Request objects that are no longer active. When pause/resume is called accidentally on an inactive Request object, the implementation must detect this and ignore the call (or throw an exception). This is the reason why the pause/resume methods check whether the associated Request object represents the currently active request.

I hope you understand the reasons for these design principles. I would like to also write a first version of a streaming BulkLoad implementation (#523), because I need that for my customer projects and I need it soon. But I don't have time to explain and justify everything in this detail.

src/connection.js
```diff
@@ -522,7 +534,15 @@ class Connection extends EventEmitter {
...
+      const ret = this.dispatchEvent('data', data);
+      if (ret === false && this.state === this.STATE.SENT_CLIENT_REQUEST) {
```
chdh: OK, I will move it to SENT_CLIENT_REQUEST.data.

src/connection.js
```diff
@@ -708,10 +729,28 @@ class Connection extends EventEmitter {
     });
   }
 
+  // Returns false to apply backpressure.
   sendDataToTokenStreamParser(data) {
     return this.tokenStreamParser.addBuffer(data);
   }
 
+  pauseRequest(request) {
```
chdh: This is only an internal method that is called from Request.pause(). It has to check whether this Request object represents the currently active request. I will add a comment.

src/request.js
chdh: No. From the API point of view, cancel() should be a member of the Request class. It's the request that is beeing canceled, not the connection.

src/token/token-stream-parser.js
```diff
@@ -27,14 +27,36 @@ class Parser extends EventEmitter {
...
   }
 
+  // Returns false to apply backpressure.
   addBuffer(buffer) {
     return this.parser.write(buffer);
   }
 
+  // Writes an end-of-message (EOM) marker into the parser transform input
+  // queue. StreamParser will emit a 'data' event with an 'endOfMessage'
+  // pseudo token when the EON marker has passed through the transform stream.
```
chdh: Thanks.

test/integration/pause-resume-test.js
chdh: OK

test/integration/pause-resume-test.js
chdh: OK, but it will not make it much clearer, because these are more heuristic values, not exact science. :)

test/integration/pause-resume-test.js
chdh: OK

test/integration/pause-resume-test.js
chdh: No, this test is a reaction to the problem reported by Congelli501 (second message in #518),
The problem was, that the last packet with the EOM mark has already been received and processed when there are still rows in the queue of the token parser transform. With the first test case which uses 200,000 rows and stops only once after 50'000 rows, the error could not be detected, because the end of the TDS message was still far away when stop() was called. In this second test case, all the rows are probably sent within the same and only packet. After the first 100ms delay, the 'message' event has surely been emitted by the MessageIO class, because it has already detected the reception of the last packet of the message.

test/integration/pause-resume-test.js
chdh: No, the end-bracket it at the end of the paragraph. But I will remove the brackets to make it clearer.

test/integration/pause-resume-test.js
chdh: OK

test/integration/pause-resume-test.js
chdh: The API documentation does not specify whether rows are emitted after cancel(). The current implementation does this, but it's not specified as part of the API. It also depends on what the server does when it receives the ATTENTION message. So the test case assumes that the reception of rows after cancel() is undefined and ignores them.

src/connection.js
tvrprasad: Sounds good.

src/request.js
tvrprasad: Sounds reasonable. Opened https://github.com/tediousjs/tedious/issues/534 to add 'cancel' on Request and deprecate it on Connection.

src/token/token-stream-parser.js
tvrprasad: np :-)

test/integration/pause-resume-test.js
tvrprasad: A brief comment would be helpful.

test/integration/pause-resume-test.js
tvrprasad: Got it. Thanks for the explanation.

test/integration/pause-resume-test.js
tvrprasad: ok. Thanks.

test/integration/pause-resume-test.js
tvrprasad: From TDS spec:
"The client can interrupt and cancel the current request by sending an Attention message. This is also known as out-of-band data, but any TDS packet that is currently being sent MUST be finished before sending the Attention message. After the client sends an Attention message, the client MUST read until it receives an Attention acknowledgment. "

Either way, blocking or emitting remaining rows, seems consistent with the spec as long as we emit all the rows before sending the ECANCEL and the code seems to do that correctly. As far as I'm concerned we can keep the current behavior.

I would modify the test though to validate the current implementation. I think it's best to be intentional if we decide to change the implementation in future, and make the correspond test modification.

test/integration/pause-resume-test.js
tvrprasad: Thanks.

test/integration/pause-resume-test.js
tvrprasad: No idea. @arthurschreiber ?

test/integration/pause-resume-test.js
tvrprasad: Thanks.

src/connection.js
tvrprasad: Thanks.

tvrprasad:
@chdh Thanks for the detailed design rationale, makes sense and sounds good.

[rant:begin] It's too bad that we can't really stop anyone from using the internal pauseRequest/resumeRequest on the connection object. I guess that's life with JavaScript. [rant:end]

Congelli501:
You can prepend protected methods by an underscore.
This convention is widely used and static code checkers (at least eslint) can check that you only use _ methods with `this`.

tvrprasad:
@Congelli501 Glad I ranted :-) Opened issue https://github.com/tediousjs/tedious/issues/536 to track this. Thanks!

test/integration/pause-resume-test.js
chdh: After some thoughts I think it's better to suppress further 'row' events after a stopped request has been cancelled or timed out. This is what the application expects. If the application explicitly wants to receive the remaining rows on a paused request, it can call resume() before calling cancel().

tvrprasad: Agreed, this is a better approach. This prevents the possibility of an application _bug_ where it chokes on 'row' events coming in after 'cancel'. Making it difficult to write bugs is good :-)

chdh:
(With the newest Node v7.7.2, "npm test" fails because of nodejs/node#11761. It has nothing to do with this PR.)

ElfenLiedGH:
@chdh 

> @ElfenLiedGH please have a look at the new version. Does it solve your problem?

i have taken lasted version and "Timeout: Request failed to complete in 15000ms" on paused request

tvrprasad:
> Note that with the newest Node v7.7.2, "npm test" fails because of nodejs/node#11761. 

I pushed a work-around for this, see https://github.com/tediousjs/tedious/pull/538.

> I have uploaded a rebased new version of this PR with all the discussed corrections and enhancements.

It'd be helpful for incremental review if you can make rebasing it's own commit and push corrections and enhancements on top as a separate commit.

chdh:
@tvrprasad 
> It'd be helpful for incremental review if you can make rebasing it's own commit and push corrections and enhancements on top as a separate commit.

OK, i will do that in the future.

src/connection.js
```diff
@@ -1245,6 +1286,7 @@ Connection.prototype.STATE = {
     name: 'SentClientRequest',
     exit: function() {
       this.clearRequestTimer();
+      this.tokenStreamParser.resume();
```
tvrprasad: Request class has 'paused' state now and that'll be inconsistent with the actual state of the request at this point.
Move the 'paused' state to Connection class and have Request pause/resume methods just forward the calls to Connection. That'll remove this coupling.

src/request.js
```diff
@@ -136,4 +137,27 @@ module.exports = class Request extends EventEmitter {
...
+    if (this.paused) {
+      return;
+    }
+    this.paused = true;
```
tvrprasad: Move 'paused' to Connection class, keep all the logic and state in one place.

src/connection.js
```diff
@@ -416,7 +416,9 @@ class Connection extends EventEmitter {
         if (this.config.options.rowCollectionOnDone) {
           this.request.rst.push(token.columns);
         }
-        return this.request.emit('row', token.columns);
+        if (!(this.state === this.STATE.SENT_ATTENTION && this.request.paused)) {
```
chdh: Because it would change the current behaviour and break the test case connection-test / cancelRequest.

It's not only cancel(), it's also the timeout.

The current behaviour is that all the remaining rows are emitted until the server receives the ATTENTION message and stops sending rows.

And there is the effect that the server sends the rows it has buffered to far when it receives the ATTENTION message. In an interactive tool (like SQL Server Management Studio) the user would want to see the rows that have been produced so far up to the time when the user clicks the cancel button.

src/connection.js
```diff
@@ -717,10 +731,33 @@ class Connection extends EventEmitter {
...
     return this.tokenStreamParser.addBuffer(data);
   }
 
+  // This is an internal method that is called from Request.pause().
+  // It has to check whether the passed Request object represents the currently
+  // active request, because the application might have called Request.pause()
+  // on an old inactive Request object.
+  pauseRequest(request) {
+    if (this.isRequestActive(request)) {
+      this.tokenStreamParser.pause();
```
chdh: It's not only a test whether the request belongs to this connection. It could also be a request that is no longer active. This can happen because of a timeout. 

When the request has already timed out, has completed normally or has been previously canceled, it's better to ignore an additional cancel() call.

src/connection.js
```diff
@@ -986,6 +1023,7 @@ class Connection extends EventEmitter {
       }
 
       this.request = request;
+      this.request.connection = this;
```
chdh: see my comment in #536.

And it wouldn't be enough to set only a reference to the function, because you also need the context, e.g.:

`this.request.pauseCallback = () => {this.pauseRequest(); };`

And then the question arises, whether Request.pauseCallback is a public member of the Request class. You have more or less the same ugliness and more confusion.

src/connection.js
```diff
@@ -995,7 +1033,10 @@ class Connection extends EventEmitter {
       this.debug.payload(function() {
         return payload.toString('  ');
       });
-      return this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
+      this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
+      if (request.paused) {                                // Request.pause() has been called before the request was started
```
chdh: OK

src/connection.js
chdh: It's more complicated than that.

1. As long as Connection.execSql() has not yet been called, the request does not know the connection.
2. When the request is cancelled or time out, we have to resume the token parser transform, but the request can still remain in the paused state.

So the paused state of the Request objects are independent of the paused state of the token parser transform.

src/request.js
chdh: The paused state of the request objects is not the same as the paused state of the token stream transform.
See above comment for the SENT_CLIENT_REQUEST exit event.

arthurschreiber:
So just as a heads up - I'll be on vacation for a week. I'd like to give this a thorough review before this gets merged, but won't get to it until after my vacation.

chdh:
> So just as a heads up - I'll be on vacation for a week. I'd like to give this a thorough review before this gets merged, but won't get to it until after my vacation.

OK.  What about #523, if I implement it in the same manner? Will you accept a PR as a preliminary solution?

src/connection.js
tvrprasad: > As long as Connection.execSql() has not yet been called, the request does not know the connection.

That's true. You could call this.request.resume() to keep the paused state in a consistent manner between Request class and Connection class. If that were to be the only concern...

> So the paused state of the Request objects are independent of the paused state of the token parser transform

This boils down to Request classes decides if the Request is paused state. And Connection class decides what that actually means. Sounds like a reasonable separation to me.

src/request.js
tvrprasad: Alright.

src/connection.js
chdh: OK, I see your point. It would be the same kind of check as in Connection.cancel(). But it's problematic.

Consider the following sequence:

- The application calls connection.execSql() to start a query.
-  The query takes longer than expected. A request timeout occurs. The driver sends ATTENTION to the server and enters the SENT_ATTENTION state. The application has no way of knowing this.
- While the driver is in the SENT_ATTENTION state and waiting for the attention response from the server, the application calls Connection.cancel(). The current implementation of the cancel() method considers this an error, writes the debug message "_Requests can only be canceled in the SentClientRequest state, not the SentAttention state_" and returns false.

From an API point of view, I would rather be forgiving than fussy. When the application wants to pause or resume a request that has already completed, then it's OK, because there are no more rows anyway that would be affected. Also when the application wants to cancel a request that has already been canceled, or close a file that is already closed, it's common to just ignore it. The driver has to protect it's inner state and the consistency of the API and therefore ignore pause/resume calls when they are not called with the currently active Request object. But it's not the purpose of the driver to educate the application.

src/connection.js
chdh: I don't understand. Request.pauseCallback/resumeCallback would be private but still can be accessed from within the Connection class. Why can't we make Request.connection private and access it from within the Connection class and make Connection.pauseRequest()/resumeRequest() private and access it from within the Request class?

src/connection.js
chdh: > This boils down to Request classes decides if the Request is paused state. And Connection class decides what that actually means. Sounds like a reasonable separation to me.

Yes exactly. Whether the request is in the paused state or not is determined by the application. This is consistent from the API point of view. Whether the token parser transform and all the other internal transforms are in the paused state, is internal to the driver and not exposed to the API.

chdh:
@ElfenLiedGH 

> i have taken lasted version and "Timeout: Request failed to complete in 15000ms" on paused request

I have extended the test cases and cannot reproduce this problem. Did you include #530? Could it be that your query takes longer than 15 seconds before the first data package is sent from the server to the client? Could you provide a test case to reproduce the problem?

src/connection.js
tvrprasad: > While the driver is in the SENT_ATTENTION state and waiting for the attention response from the server, the application calls Connection.cancel(). The current implementation of the cancel() method considers this an error, writes the debug message "Requests can only be canceled in the SentClientRequest state, not the SentAttention state" and returns false.

I think that's a bug and we should fix that. Opened https://github.com/tediousjs/tedious/issues/545. 

My comment is about after the Request is done and the callback is invoked. At this point the application knows the request is done. I'm not able to come up with an application pattern where there could reasonably be more calls on the Request object other than a bug in the code. Being forgiving here I think hurts than helps the developer by hiding bugs. And there is no need for a isCompleted either, as the callback already gives that information.

src/connection.js
tvrprasad: The difference, at least as I see it, is the coupling.

In one case Request has knowledge of Connection class and internal method names inside the Connection class. For example if I rename the methods in the Connection class or move them to a sub-class I'd have to change the Request class.

In the other case, the Connection class is just passing the callbacks to Request and Request class honors the contract by invoking them, just as it invokes user callback on completion.

src/connection.js
tvrprasad: Agreed.

src/connection.js
chdh: > The difference, at least as I see it, is the coupling.

Yes, that's true. But I would like to finish my work on this feature now and work on other things. Can't we agree that the current implementation is good enough? There is always something that can be improved and de-coupling like this has not been a design principle in Tedious until now.

tvrprasad: @chdh The implementation is certainly good enough and I'll approve. The couple of tweaks left I can make them after merge. I'll let @arthurschreiber merge the change as he wanted to review before merge.

Thanks for your contribution and thanks for you patience!

ElfenLiedGH:
@chdh 

> @ElfenLiedGH
> 
> i have taken lasted version and "Timeout: Request failed to complete in 15000ms" on paused request
> I have extended the test cases and cannot reproduce this problem. Did you include #530? Could it be that your query takes longer than 15 seconds before the first data package is sent from the server to the client? Could you provide a test case to reproduce the problem?

Oops. i took this only, now it works, thz

arthurschreiber:
As mentioned in https://github.com/tediousjs/tedious/pull/547, I'm trying to release `2.0.0` next week.

This would also be a great feature to include soon after! 👍

tvrprasad:
@arthurschreiber Now that we got the 2.0.0 released, I'll go ahead and merge this unless you have more comments.

tangkhaiphuong:
@arthurschreiber  Please merge this pull request and release this feature. I wait a long time this feature for resolve back-pressure. Thanks!!!

arthurschreiber:
@tvrprasad @tangkhaiphuong @chdh Just a quick update here: I'd _really_ like to have #528 merged before this and have this based on top of it. I believe the two changes fit nicely together and will also pave the way for other streaming changes like #523 or the new token stream reader and writer I've been working on over in https://github.com/tediousjs/node-tds-token-stream.

I'll try to get #528 in shape by the end of the week! 👍

ilusharulkov:
Hi! 
Sorry for my impatience, but I realy need this for sync large amount of data. Is there some forecast
when this feauture will be merged into master?

ilusharulkov:
Guys!
Any news ?

tangkhaiphuong:
@arthurschreiber  It's over 3 months but this feature still not merge, so what problem?

arthurschreiber:
Heya @chdh,

thank you so much for this contribution! I cleaned up the test cases a bit and fixed one tiny issue with paused requests and connection closing, but I'm really happy to have these changes be finally part of `tedious`.

I'll review the other contributions that were made since the last release and push a new version including these changes as soon as possible. I'll also try to get streaming bulk data loads into another release shortly after.

Again, thank you so much for this contribution and sorry that it has taken so long to get this merged.

❤️

chdh:
@arthurschreiber
Thanks for your cleanup, extended testing and improvements. I'm glad that this PR will make it into the next Tedious release.

jschell12:
Hey @arthurschreiber. Any word on when you'll release to npm?

v-suhame:
@chdh `pause-resume-test` fails occasionally in CI, I have restarted the travis build couple of times on contributor's PR, forgot to copy the error stack though. I will share the message, if I notice it again for you to fix.

v-suhame:
Here is the stack
```
AssertionError: false == true
    at Object.ok (/home/travis/build/tediousjs/tedious/node_modules/nodeunit/lib/types.js:83:39)
    at Timeout._onTimeout (/home/travis/build/tediousjs/tedious/test/integration/pause-resume-test.js:114:14)
    at ontimeout (timers.js:386:14)
    at tryOnTimeout (timers.js:250:5)
    at Timer.listOnTimeout (timers.js:214:5)
```

arthurschreiber:
@v-suhame this is actually my fault. 😥 I’ll open a pull request to fix the flaky tests soon.

chdh:
@arthurschreiber @v-suhame I had similar effects with my version of the tests. I suggest to repeat the isPaused() test for a couple of times with a delay for each retry.

hosseinGanjyar:
Is there a way for my problem?
https://github.com/patriksimek/node-mssql/issues/471
thanks a lot

chdh:
@hosseinGanjyar If pause/resume was implemented in node-mssql, you could call `Request.pause()` to stop the "row" events and `Request.resume()` to resume them. I think it would be easy to implement. Node-mssql could just forward the pause/resume calls to the Tedious driver. Without that, I can't see a way to call the pause/resume methods of the underlying Tedious driver.

gsamal:
@chdh I am using the upper layer library `node-mssql` and they have added the pause/resume on the tedious driver. It all seems to be working until I noticed that even if request is paused, it keeps fetching rows and adds it in memory. 
An easy way to reproduce this is after n rows, hit pause, then do nothing. The memory consumption keeps growing. Does this pause/resume work on the socket/db-connection level? or how does it work? Any ides why memory keeps rows as though it looks the record are still stacking up in memory even if it's paused.

chdh:
@gsamal The pause mechanism works by applying back-pressure on the internal chain of streams. Each stream continues to buffer data until the high water mark is reached. Then it applies back-pressure to the next preceding stream in the chain.

Could you open a new issue and supply some more detailed information about the effect you are investigating? How much does the memory grow after pause() was called. The best way would be to provide an isolated test case that allows us to reproduce the problem,

gsamal:
@chdh Okay. But it seems to be keep adding records in a buffer somewhere, only pause the internal chain of streams. I can see a similar issue reported here - [node-mssql#832](https://github.com/tediousjs/node-mssql/issues/832). Can you please take a look at it? 

In a simpler sense, if I pause the request after getting n records, then I monitor the network and memory, it keeps increasing.

arthurschreiber:
@gsamal This was fixed via #878.

arthurschreiber:
The fix for this is part of the `6.0.1` release. 🎉

Issue #523, opened 2017-03-02
Streaming BulkLoad for large table imports


chdh:
The current BulkLoad implementation requires that all data rows have to be stacked in memory before they can be sent to the database server. This is not suitable for large table imports.

A streaming bulk insert implementation must use data flow control to apply back pressure when the data arrives faster than it can be sent to the database. This could be modeled after the standard Node.js stream.Writable interface: Return false from write() to pause, emit a 'drain' event to resume.

It has to be determined whether a future BulkLoad streaming class should extend the stream.Writable class of the Node.js API (or the readable-stream package), or just implement the same interface, The objective is to make it compatible with stream.Readable.pipe().

arthurschreiber:
I'll take another look soon and will let you know of how I think this would be best implemented.

PR #528, opened 2017-03-05
WIP/POC: Streamify internals

arthurschreiber:
This is a very rough Proof-of-Concept I build based on some refactorings I've wanted to do since quite some time.

The general idea is to convert all the internals of tedious to make much more use of streams, which in turn will allow us to expose useful functionality like pausable/resumable requests while improving performance and massively reducing memory usage.

The naming and the code itself are very much work-in-progress, but this shows the general idea shown in these refactorings works.

In c9c6480, I replaced the `ReadablePacketStream` with a `IncomingMessageStream` that transforms the raw bytes we read over a socket into `IncomingMessage` objects. These `IncomingMessage` are streams again, but will only emit packets that belong to a single TDS Message.

In ca97ddb, I also added an `OutgoingMessage` class that basically acts the other way around - raw data that is written into it will be transformed into a series of TDS packets.

In 2e92e8e, I pulled the creation of these `OutgoingMessage` and refactored the `MessageIO#sendMessage` function to take instances of `OutgoingMessage`. The long term goal here is to be able to create an (empty) `OutgoingMessage`, pass it to `MessageIO#sendMessage`, and then write out message data progressively into the `OutgoingMessage` object. This will be especially helpful with the Bulk Inserts.

In 41d8169, I changed the way the data flows from the socket into the token parser to not use `data` events, but to use pipes instead.

42d0c73 contains a bogus change to show that in-flight requests can now be paused by pausing the `IncomingMessage` stream and the token parser.

---

@chdh @tvrprasad What do y'all think? These changes are obviously experimental, but we could polish this up for the release after `1.15.0`. Would one of you be willing to take the ideas here and work on this?


chdh:
I understand the idea, but I think an architecture that uses streams of streams is too tricky and too complex. The main stream may accept or emit the next substream while the previous substream is sill active. It's too complicated to master the asynchronicity that is happening here.

arthurschreiber:
> I understand the idea, but I think an architecture that uses streams of streams is too tricky and too complex.

Unless I'm mistaken, this is following e.g. the implementation of HTTP in Node.js where each incoming request from a socket is parsed in a streaming fashion and exposed to the user as a `IncomingMessage` (which is a subclass of `ReadableStream`).

> The main stream may accept or emit the next substream while the previous substream is sill active.

This should not happen, and if it does, it's a bug in the implementation. As you mention yourself, TDS is a ping-pong of messages, with the exception of the Out-of-band attention message. This means that again, with the exception of attention messages, no more than one message is ever going to be active at the same time.

In the case of an attention message coming in, we will need special handling for this anyway, and we really want those to be handled as soon as possible.

> Only one incoming and one outgoing message per connection (or SMUX channel) is active at a time. So the architecture should not use a stream of messages.

This is again true for the HTTP implementation shipping with Node.js - on a given socket only a single request / response can be active at a time. Both are still implemented as Streams on top of a Socket.

> If for each message a new stream object is used, a new TokenStreamParser instance could be created for each message and the packetUnwrapper could be eliminated

The packetUnwrapper is just a hack for now to make sure debugging/logging can continue to work with `Packet` objects. I'd like to move away from even having a `Packet` class at all. The streams in `tedious` should work on Buffer instances only, with the exception of the token stream parser which will continue to emit parsed token objects.

But yes, we should use a new TokenStreamParser for each Message, otherwise we run into a high risk that when we receive an attention message and unpipe whatever message is currently being processed, we will end up in an undefined token parsing state.

Thanks for taking a look @chdh! I hope I could address some of your concerns. 😄

tvrprasad:
@arthurschreiber I'll review this in the next few days.

tvrprasad:
> But yes, we should use a new TokenStreamParser for each Message, otherwise we run into a high risk that when we receive an attention message and unpipe whatever message is currently being processed, we will end up in an undefined token parsing state.

I looked at the TDS spec and below is based on my read....

Client only needs to handle one message at any time, it should never get an out of band message from Server, which means there should only be one instance of a TokenStreamParser active at any time. Client can get an attention acknowledgement but that's a status in the DONE token. From the spec:

_2.2.2.9 Attention Acknowledgment 
After a client has sent an interrupt signal to the server, the client MUST read returning data until the interrupt has been acknowledged. Attention messages are acknowledged in the DONE token data stream. or more details, see section 2.2.7.5._

Same thing on the send side. The "out of band" Attention message is a special packet in the context of the current message. Again, from the spec:

_2.2.1.7 Attention 
....
If a complete request has not been sent to the server, the client MUST send the next packet with both ignore bit (0x02) and EOM bit (0x01) set in the status to cancel the request._

arthurschreiber:
@tvrprasad You're right. There is no such thing as out-of-band messages, as the attention signal is sent via a `done` token while a message is in-flight.

There is still the issue that when a packet signals attention, the token parser can be left in an undefined state because tokens can span multiple packets, and the spec does not define whether all token data will have been transferred over before the packet with the attention bit gets send.

If I can find some free time, I'll see if I can test this out.

tvrprasad:
> There is still the issue that when a packet signals attention, the token parser can be left in an undefined state because tokens can span multiple packets, and the spec does not define whether all token data will have been transferred over before the packet with the attention bit gets send.

If you only receive a partial token followed by DONE token, I'd think that would constitute a structurally invalid response. I'm not sure how that can even be parsed, I didn't see any mention that Attention Acknowledgement needs to come in it's own packet from the server.

On the client side though the TDS spec does make it clear that the current packet needs to be fully sent before sending Attention message. Here is the excerpt:

_The client can interrupt and cancel the current request by sending an Attention message. This is also known as out-of-band data, but any TDS packet that is currently being sent MUST be finished before sending the Attention message._

We currently send Attention message from the client in cancel() and requestTimeout() and we don't pay much attention to the current state of any message being sent. This works fine in current code as we always send the message as a single chunk.

But once we start streaming the outgoing message, this will be a problem as we could be in the middle of sending a packet when, say, timeout happens. The resume/pause PR (https://github.com/tediousjs/tedious/pull/518) from @chdh this issue I believe, ATTENTION message could be sent while the request is paused in the middle of sending a message. I'm not sure your PR handles this correctly, I haven't looked at that detail yet.

arthurschreiber:
@tvrprasad Thanks again for your inputs. ❤️ I'll be taking a deep look into the poits you raised tomorrow, as I want to get this merged to allow and implementation of #523 and #518 on top of this. 

I think I might be missing some details. 🤔

test/unit/message/outgoing-message-test.js
```diff
@@ -0,0 +1,186 @@
...
+    const message = new OutgoingMessage(1, false, 20, debug);
+    const data = Buffer.from('12345');
+
+    message.on('data', function(packetData) {
+      const packet = new Packet(packetData);
+      test.strictEqual(1, packet.type());
+      test.strictEqual(1, packet.packetId());
+      test.ok(packet.isLast());
+      test.strictEqual(13, packet.length());
+      test.deepEqual(data, packet.data());
```
tvrprasad: Validate resetconnection flag being set correctly on packet.

test/unit/message/outgoing-message-test.js
```diff
@@ -0,0 +1,186 @@
...
+      test.ok(packets[2].isLast());
+      test.strictEqual(13, packets[2].length());
+      test.deepEqual(Buffer.from('56789'), packets[2].data());
+
+      test.done();
+    });
+  },
+
+  'transforms the data as soon as possible': function(test) {
+    const debug = new Debug();
```
tvrprasad: Unclear what this test is about. Please add a few comments.

test/unit/message/outgoing-message-test.js
```diff
@@ -0,0 +1,186 @@
...
+
+      test.strictEqual(3, packets[2].packetId());
+      test.strictEqual(1, packets[2].type());
+      test.ok(packets[2].isLast());
+      test.strictEqual(13, packets[2].length());
+      test.deepEqual(Buffer.from('56789'), packets[2].data());
+
+      test.done();
+    });
+  },
```
tvrprasad: You may add one corner case test - data size multiple of packet size so that the last packet is a full packet.

src/connection.js
```diff
@@ -550,8 +552,16 @@ class Connection extends EventEmitter {
       this.socket.on('close', this.socketClose);
       this.socket.on('end', this.socketEnd);
       this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-      this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
-      this.messageIo.on('message', () => { this.dispatchEvent('message'); });
+      this.messageIo.incomingMessageStream.on('data', (message) => {
+        if (this.state.name == 'SentClientRequest') {
```
arthurschreiber: This needs to be cleaned up. 👍 

According to section `2.2.4 - Packet Data Token and Tokenless Data Streams` of the TDS spec, I believe the decision on whether the message contents is a token stream or not should not be based on the current state of the connection, but on the message type instead.

Piping directly into `this.tokenStreamParser.parser` is not great from a coupling perspective, but the token stream parser is a bit of a mess right now and needs to be cleaned up anyway, so I'm willing to compromise here.

> A cleaner option may be to capture this logic in a helper method, let this flow through dispatch event and invoke the helper in the 'data' methods of the respective states.

This is what I want to get away from here. 😄 I'll see if I can clean this up.

src/connection.js
```diff
@@ -585,8 +595,12 @@ class Connection extends EventEmitter {
 
   requestTimeout() {
     this.requestTimer = undefined;
-    this.messageIo.sendMessage(TYPE.ATTENTION);
-    return this.transitionTo(this.STATE.SENT_ATTENTION);
+
+    const message = new OutgoingMessage(TYPE.ATTENTION, false, this.messageIo.packetSize(), this.debug);
+    this.messageIo.sendMessage(message);
```
arthurschreiber: Right. My plan is to implement this via a `cancel` method on `OutgoingMessage`, which ends the stream and `_flush` could then handle sending a packet with the attention bit set.

src/connection.js
```diff
@@ -738,7 +759,11 @@ class Connection extends EventEmitter {
       ntlmpacket: this.ntlmpacket,
       additional: this.additional
     });
-    this.messageIo.sendMessage(TYPE.NTLMAUTH_PKT, payload.data);
+
+    const message = new OutgoingMessage(TYPE.NTLMAUTH_PKT, false, this.messageIo.packetSize(), this.debug);
```
arthurschreiber: Good idea!

src/message-io.js
```diff
@@ -1,74 +1,22 @@
...
-    this.packetStream.on('data', (packet) => {
-      this.logPacket('Received', packet);
-      this.emit('data', packet.data());
-      if (packet.isLast()) {
-        this.emit('message');
-      }
-    });
 
-    this.socket.pipe(this.packetStream);
+    this.incomingMessageStream = new IncomingMessageStream(this.debug);
```
arthurschreiber: This was the "simplest" way to implement this without shuffling too many things around, but you're right, it's not clean. I'll think about how to improve this. 👍 

Long term, I want to get completely rid of `MessageIO`, and replace it with a custom Duplex stream so we can move away from using `tls.createSecurePair` and use the preferred `TLSSocket` class instead.

src/message/incoming-message-stream.js
```diff
@@ -0,0 +1,86 @@
...
+          this.currentMessage.end(packet);
+          this.currentMessage = undefined;
+        } else {
+          // If too much data is buffering up in the
+          // current message, wait for it to drain.
+          if (!this.currentMessage.write(packet)) {
+            this.currentMessage.once('drain', () => {
+              this.processBufferedData();
+            });
+            return;
```
arthurschreiber: That is not quite right! 😄 

The `return` here will break the outer `while` loop to make sure we don't continue writing data into `this.currentMessage` once it signalled that data pressure is building up and needs to drain first.

src/message/incoming-message-stream.js
```diff
@@ -0,0 +1,86 @@
...
+
+    // Not enough data to read the next packet. Stop here and wait for
+    // the next call to `_transform`.
+    this.waitForNextChunk();
+  }
+
+  _transform(chunk, encoding, callback) {
+    this.bl.append(chunk);
+
+    this.waitForNextChunk = () => { process.nextTick(callback); };
```
arthurschreiber: I had this implemented like that initially, but there was a bit of a performance difference here, but I don't remember whether it was significant or not. I'll test again and change it back if it does not matter! 👍

test/unit/message-io-test.coffee
```diff
@@ -14,74 +15,21 @@ class Connection extends Duplex
 packetType = 2
 packetSize = 8 + 4
 
-exports.sendSmallerThanOnePacket = (test) ->
```
arthurschreiber: Good catch! 👍

test/unit/message/outgoing-message-test.js
arthurschreiber: Good idea! 👍

src/message-io.js
tvrprasad: > This was the "simplest" way to implement this without shuffling too many things around

Doing a "new IncomingMessageStream(this.debug)" in Connection class and passing it to new MessageIO() shouldn't shuffle too many things :-)

Agree on the longer term, MessageIO does seem like a Duplex stream to me.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,85 @@
+const Transform = require('readable-stream').Transform;
+
+const { Packet, HEADER_LENGTH, TYPE } = require('../packet');
+
+const BufferList = require('bl');
+
+function isValidType(type) {
+  switch (type) {
```
tvrprasad: I think this function should go into Packet.js. And a corresponding unit test would be nice.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,85 @@
...
+  OutgoingMessage
+
+  This class can be used to transform the raw contents of a single
+  TDS message into a stream of TDS packets.
+*/
+module.exports = class OutgoingMessage extends Transform {
+  constructor(type, resetConnection, packetSize, debug) {
+    super();
+
+    if (typeof type !== 'number' || !isValidType(type)) {
```
tvrprasad: _typeof type !== 'number'_ could be folded into isValidType()

src/message/outgoing-message.js
```diff
@@ -0,0 +1,85 @@
...
+      throw new TypeError('"type" must be a a supported TDS message type');
+    }
+
+    if (typeof packetSize !== 'number' || packetSize <= HEADER_LENGTH) {
```
tvrprasad: This can be isValidPacketSize method in Packet.js... if you like :-)

src/connection.js
```diff
@@ -550,8 +550,19 @@ class Connection extends EventEmitter {
...
       this.socket.on('end', this.socketEnd);
       this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-      this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
-      this.messageIo.on('message', () => { this.dispatchEvent('message'); });
+      this.messageIo.incomingMessageStream.on('data', (message) => {
+        // Pre-Login responses from the server are sent in Tabular Result messages,
+        // but do not contain a token stream. All other Tabular Result messages
+        // contain a token stream.
+        if (message.type === TYPE.TABULAR_RESULT && this.state.name !== 'SentPrelogin') {
+          message.pipe(this.tokenStreamParser.parser, { end: false });
```
arthurschreiber: Haha, true. 😄

src/connection.js
```diff
@@ -550,8 +550,19 @@ class Connection extends EventEmitter {
...
+        } else {
+          message.on('data', (chunk) => { this.dispatchEvent('data', chunk); });
```
arthurschreiber: Good idea!

test/unit/message/incoming-message-stream-test.js
```diff
@@ -0,0 +1,64 @@
...
+    packet = new Packet(TYPE.TABULAR_RESULT);
+    packet.addData(payload.slice(0, 3));
+    incomingMessageStream.write(packet.buffer);
+
+    packet = new Packet(TYPE.TABULAR_RESULT);
+    packet.last(true);
+    packet.addData(payload.slice(3));
+    incomingMessageStream.write(packet.buffer);
+  }
+};
```
tvrprasad: Need a test to validate that back pressure is working correctly in the interaction with IncomingMessage.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,67 @@
...
+*/
+module.exports = class OutgoingMessage extends Transform {
+  constructor(type, resetConnection, packetSize, debug) {
+    super();
+
+    if (!isValidType(type)) {
+      throw new TypeError('"type" must be a a supported TDS message type');
+    }
+
+    if (typeof packetSize !== 'number' || packetSize <= HEADER_LENGTH) {
```
tvrprasad: Replace with isValidSize().

src/message/incoming-message-stream.js
```diff
@@ -0,0 +1,69 @@
...
+          this.push(this.currentMessage);
+        }
+
+        if (packet.isLast()) {
+          this.currentMessage.end(packet.data());
+          this.currentMessage = undefined;
+        } else {
+          // If too much data is buffering up in the
+          // current message, wait for it to drain.
+          if (!this.currentMessage.write(packet.data())) {
```
tvrprasad: Isn't the pattern to setup a pipe to currentMessage and then do push()? Curious about the choice of write() instead.

test/unit/message/outgoing-message-test.js
```diff
@@ -0,0 +1,219 @@
...
+            test.ok(!packets[0].isLast());
+            test.ok(!packets[0].resetConnection());
+            test.strictEqual(20, packets[0].length());
+            test.deepEqual(Buffer.from('123456789|12'), packets[0].data());
+
+            test.strictEqual(2, packets[1].packetId());
+            test.strictEqual(1, packets[1].type());
+            test.ok(packets[1].isLast());
+            test.ok(!packets[1].resetConnection());
+            test.strictEqual(14, packets[1].length());
```
tvrprasad: > You may add one corner case test - data size multiple of packet size so that the last packet is a full packet.

Is this test meant to address this comment? If so, the last packet needs to be full, i.e. same size as the first packet.

test/unit/message-io-test.coffee
```diff
@@ -12,17 +13,18 @@ class Connection extends Duplex
     @emit('packet', packet)
     callback()
 
-packetType = 2
+packetType = TYPE.SQL_BATCH
 packetSize = 8 + 4
```
arthurschreiber: Done.

src/message/incoming-message-stream.js
arthurschreiber: We can't pipe from the `IncomingMessageStream` directly into the `this.currentMessage`, because we need to make sure we stop writing once we reach the end of the current message.

src/message/incoming-message-stream.js
```diff
@@ -0,0 +1,69 @@
...
+            this.currentMessage.once('drain', () => {
```
arthurschreiber: No, that can't happen here. As long as `callback` does not get executed, `_transform` will not be called again.

This is documented in the Node.js API documentation for Transform streams.

src/message/incoming-message-stream.js
```diff
@@ -0,0 +1,69 @@
...
+          }
+        }
+      } else {
+        break;
+      }
+    }
+
+    // Not enough data to read the next packet. Stop here and wait for
+    // the next call to `_transform`.
+    process.nextTick(callback);
```
arthurschreiber: Each `_transform` call gets passed in a chunk of data. We add this chunk to our `BufferList` object and start consuming that data. Once we're done consuming the data, we need to execute the `callback` that was also passed to `_transform` to signal that we're done processing the current chunk and can start processing the next chunk.

`callback` does not strictly need to be executed asynchronously (via `process.nextTick`), but doing so is more consistent to how other streams are implemented and is generally "good form".

src/message/outgoing-message.js
arthurschreiber: I don't think we currently have any use for an extracted method in any other place, so I'll leave this in here for now.

test/unit/message/incoming-message-stream-test.js
arthurschreiber: Good idea!

src/message/incoming-message-stream.js
tvrprasad: For my edification, I'd be curious to know why this is good form. Sample code on stream documentation page invokes callback() synchronously. Not to say that makes it right :-)

I see why callback needs to be invoked asynchronously in the case where the data is consumed asynchronously - to ensure callback is invoked after data is consumed. In our case, we're consuming the data synchronously.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,67 @@
...
+      throw new TypeError(`"packetSize" must be a number greater than ${HEADER_LENGTH}`);
```
tvrprasad: String is in reverse quotes.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  _transform(chunk, encoding, callback) {
+    this.bl.append(chunk);
+
+    while (this.packetDataSize < this.bl.length) {
+      this.transformNextPacket();
+    }
+
+    process.nextTick(callback);
```
tvrprasad: You can invoke callback() directly.

src/message/outgoing-message.js
```diff
@@ -0,0 +1,67 @@
...
+    this.bl.consume(this.packetDataSize);
+
+    this.logPacket(packet);
+
+    this.push(packet.buffer);
+  }
+
+  _flush(callback) {
+    this.transformNextPacket(true);
+    process.nextTick(callback);
```
tvrprasad: You may invoke callback() directly.

src/message/incoming-message-stream.js
tvrprasad: Pending.

src/message/outgoing-message.js
tvrprasad: Pending.

src/packet.js
```diff
@@ -202,12 +202,26 @@ class Packet {
 }
 module.exports.Packet = Packet;
 
-module.exports.isPacketComplete = isPacketComplete;
-function isPacketComplete(potentialPacketBuffer) {
-  if (potentialPacketBuffer.length < HEADER_LENGTH) {
+module.exports.isValidType = isValidType;
+function isValidType(type) {
```
tvrprasad: Pending.

test/unit/message/outgoing-message-test.js
tvrprasad: Pending.

test/unit/message/outgoing-message-test.js
tvrprasad: Pending.

test/unit/message/outgoing-message-test.js
tvrprasad: Pending.

tvrprasad:
@arthurschreiber Do you need help getting this committed? I can pitch in some time to wrap this up.

src/message/outgoing-message.js
arthurschreiber: This is on purpose, it's a ES6 template string. Notice the use of `${...}` inside the string.

v-suhame:
@arthurschreiber I can help resolve conflicts if you're ready to merge this PR.

arthurschreiber:
The core idea of this pull request was implemented via #768, but in a much cleaner fashion - so I'll close this one! 🎉

PR #529, opened 2017-03-06
Avoid unnecessary allocation in WritableTrackingBuffer

chdh:
In the current implementation of WritableTrackingBuffer, a new internal buffer of initialSize bytes is allocated each time the data property of the WritableTrackingBuffer is accessed. This is because the getter method of WritableTrackingBuffer.data calls newBuffer(0) and newBuffer() sets the size parameter to initialSize if it is zero.

### DIFF
--- a/src/tracking-buffer/writable-tracking-buffer.js
+++ b/src/tracking-buffer/writable-tracking-buffer.js
@@ -7,6 +7,7 @@ require('../buffertools');
 const SHIFT_LEFT_32 = (1 << 16) * (1 << 16);
 const SHIFT_RIGHT_32 = 1 / SHIFT_LEFT_32;
 const UNKNOWN_PLP_LEN = new Buffer([0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
+const ZERO_LENGTH_BUFFER = new Buffer(0);
 
 /*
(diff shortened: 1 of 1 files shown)


src/tracking-buffer/writable-tracking-buffer.js
```diff
@@ -52,14 +53,13 @@ module.exports = class WritableTrackingBuffer {
...
   newBuffer(size) {
-    size || (size = this.initialSize);
     const buffer = this.buffer.slice(0, this.position);
     if (this.compositeBuffer) {
       this.compositeBuffer = Buffer.concat([this.compositeBuffer, buffer]);
     } else {
       this.compositeBuffer = buffer;
     }
-    this.buffer = new Buffer(size);
+    this.buffer = (size === 0) ? ZERO_LENGTH_BUFFER : new Buffer(size);
```
tvrprasad: Wouldn't this cause the same Buffer object to be shared across multiple instances of WritableTrackingBuffer?

chdh: Yes, ZERO_LENGGTH_BUFFER is common for all instances. But it's a zero length buffer and therefore immutable.

tvrprasad: Ah. That's clever :-)

PR #530, opened 2017-03-07
Stop request timer when first packet of response message is received

chdh:
This behavior is according to the TDS standard.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -523,7 +523,6 @@ class Connection extends EventEmitter {
     this.socket.on('end', this.socketEnd);
     this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
     this.messageIo.on('data', (data) => {
-      this.clearRequestTimer();
       this.dispatchEvent('data', data);
     });
     this.messageIo.on('message', () => {
@@ -543,6 +542,7 @@ class Connection extends EventEmitter {
   }
 
   createRequestTimer() {
+    this.clearRequestTimer();                              // release old timer, just to be safe
     if (this.config.options.requestTimeout) {
       return this.requestTimer = setTimeout(this.requestTimeout, this.config.options.requestTimeout);
     }
@@ -557,7 +557,7 @@ class Connection extends EventEmitter {
   }
 
   requestTimeout() {
-    this.requestTimer = void 0;
+    this.requestTimer = undefined;
     this.messageIo.sendMessage(TYPE.ATTENTION);
     return this.transitionTo(this.STATE.SENT_ATTENTION);
   }
@@ -1005,7 +1005,6 @@ class Connection extends EventEmitter {
       this.debug.log(message);
       return false;
     } else {
-      this.clearRequestTimer();
       this.request.canceled = true;
       this.messageIo.sendMessage(TYPE.ATTENTION);
       this.transitionTo(this.STATE.SENT_ATTENTION);
@@ -1244,6 +1243,9 @@ Connection.prototype.STATE = {
   },
   SENT_CLIENT_REQUEST: {
     name: 'SentClientRequest',
+    exit: function() {
+      this.clearRequestTimer();
+    },
     events: {
       socketError: function(err) {
         const sqlRequest = this.request;
@@ -1252,10 +1254,10 @@ Connection.prototype.STATE = {
         return this.transitionTo(this.STATE.FINAL);
       },
       data: function(data) {
+        this.clearRequestTimer();                          // request timer is stopped on first data package
         return this.sendDataToTokenStreamParser(data);
       },
       message: function() {
-        this.clearRequestTimer();
         this.transitionTo(this.STATE.LOGGED_IN);
         const sqlRequest = this.request;
         this.request = void 0;
--- a/src/connection.js
+++ b/src/connection.js
@@ -522,7 +522,10 @@ class Connection extends EventEmitter {
     this.socket.on('close', this.socketClose);
     this.socket.on('end', this.socketEnd);
     this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-    this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
+    this.messageIo.on('data', (data) => {
+      this.clearRequestTimer();
+      this.dispatchEvent('data', data);
+    });
     this.messageIo.on('message', () => {
       return this.dispatchEvent('message');
     });
@@ -567,7 +570,8 @@ class Connection extends EventEmitter {
 
   clearRequestTimer() {
     if (this.requestTimer) {
-      return clearTimeout(this.requestTimer);
+      clearTimeout(this.requestTimer);
+      this.requestTimer = undefined;
     }
   }
 


src/connection.js
```diff
@@ -567,7 +570,8 @@ class Connection extends EventEmitter {
 
   clearRequestTimer() {
     if (this.requestTimer) {
-      return clearTimeout(this.requestTimer);
+      clearTimeout(this.requestTimer);
+      this.requestTimer = undefined;
```
tvrprasad: For consistency with the rest of the file, please change this to:
```JavaScript
this.requestTimer = void 0;
```

tvrprasad: Remove the call to clearTimeout under SENT_CLIENT_REQUEST:message. We don't need that now.

tvrprasad: We have a requestTimeout integration test in connection-test.coffee that tests the old behavior. It's still relevant, so we can keep it.

To test the new behavior, we can add a test that'll sleep for a little over the requestTimeout amount in the first invocation of 'row' event handler. With old behavior that would cause a timeout, whereas with the new behavior it should not.

arthurschreiber: 'undefined' is preferred over 'void 0'. 'void 0' is a relic of the CoffeeScript to JavaScript conversion. 😔

tvrprasad: This thread seems to suggest 'void 0' is safer - http://stackoverflow.com/questions/5716976/javascript-undefined-vs-void-0.

If we like 'undefined' I'll send a PR to change all 'void 0' s to undefined for consistency. Otherwise it'll keep bugging me :-)

chdh: > Remove the call to clearTimeout under SENT_CLIENT_REQUEST:message. We don't need that now.

I would like to keep the call to clearRequestTimer() there, to guarantee that the timer is also stopped when a 'message' event is received without a 'data' event. It doesn't hurt to have a redundant call to clearRequestTimer(). Another solution would be to place it in the exit event of the SENT_CLIENT_REQUEST state.

> This thread seems to suggest 'void 0' is safer

This StackOverflow entry is old. Nobody uses "void 0" any more with JavaScript ES6. The initial value of this.requestTimer is `undefined`, so it's appropriate to reset it to `undefined`.

> If we like 'undefined' I'll send a PR to change all 'void 0' s to undefined for consistency. Otherwise it'll keep bugging me :-)

It would be good to clean up the source code, replace `void 0`, remove unnecessary return statements which probably also stem from the CoffeScript conversion, replace `inherit` calls with the new class syntax, convert the remaining CoffeScript files, etc. etc. But keep in mind that the Tedious package consists of 15,300 lines of source code. Until after a complete overhaul, we will have a mix of old and new language constructs. To really clean up the code base, we probably need a sponsor...

tvrprasad: > It doesn't hurt to have a redundant call to clearRequestTimer().

Fair enough.

> Another solution would be to place it in the exit event of the SENT_CLIENT_REQUEST state.

That would be too late, isn't it? I think this only gets invoked after all the rows are consumed. You could add another redundant call on 'exit' to account for the case, if there is one, where neither 'data' nor 'message' is received.

chdh: > We have a requestTimeout integration test in connection-test.coffee that tests the old behavior. It's still relevant, so we can keep it.
> To test the new behavior, we can add a test that'll sleep for a little over the requestTimeout amount in the first invocation of 'row' event handler. With old behavior that would cause a timeout, whereas with the new behavior it should not.

The existing test in connection-test.coffee uses the SQL statement `select 1 as C1;waitfor delay '00:00:05';select 2 as C2`. One would expect that the database server sends the first row immediately, waits for 5 seconds, and then sends the second row. But this is not the case. The first row is only received after the 1s request timeout has expired and the client has sent the ATTENTION message. While the data packet with the first row is received, the driver is in the SENT_ATTENTION state.

tvrprasad: Right, that is the behavior I was expecting with the current test and we should keep it.

My suggestion is to add a new test like this.

- Send any query that returns multiple rows.
- In the first invocation of 'row' event handler, sleep for a little over the request timeout period.
   o Something like child_process.execSync('sleep 2')

In the old code this should cause the request to timeout. With your fix, this should work because you clear the request timer already. Does that make sense?

chdh: Without calling Request.pause() it makes no sense.

tvrprasad: > I think it's not possible to design a reasonable and better test case without calling Request.pause(). But you could try yourself.

I tried and you're right. Sleep in 'row' does not help. 'message' event gets handled before timeout callback is invoked and clears the timer.

> I meant to move the existing call of clearRequestTimer() from STATE.SENT_CLIENT_REQUEST.events.message to STATE.SENT_CLIENT_REQUEST.exit, in addition to the new call. That way we would have a fail-safe for future code modifications. 

That makes sense. Please go ahead and make that update.

src/connection.js
```diff
@@ -522,7 +522,10 @@ class Connection extends EventEmitter {
     this.socket.on('close', this.socketClose);
     this.socket.on('end', this.socketEnd);
     this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-    this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
+    this.messageIo.on('data', (data) => {
+      this.clearRequestTimer();
```
tvrprasad: This technically belongs under SENT_CLIENT_REQUEST.data. Can't come up with a case where this breaks anything though...

src/connection.js
arthurschreiber: > If we like 'undefined' I'll send a PR to change all 'void 0' s to undefined for consistency. Otherwise it'll keep bugging me :-)

This would be great. There's quite a bit of leftovers from that conversion, and getting those cleaned up would be 💖.

`void 0` was safer in ye olde days of ES3 without the `'use strict'`, where you could do `var undefined = true` and break all code that used `undefined`. With the `'use strict'` directive, redefinition of `undefined` is no longer allowed and will throw an error.

arthurschreiber: I think we could test a timeout by pausing one of the underlying streams after receiving the first row (and before receiving it). It's pretty ugly test case and I'm not sure whether testing this is worth it.

tvrprasad: Probably not worth it, considering the real pause/resume is coming soon and we'll have clean tests.

src/connection.js
```diff
@@ -1248,10 +1254,10 @@ Connection.prototype.STATE = {
         return this.transitionTo(this.STATE.FINAL);
       },
       data: function(data) {
+        this.clearRequestTimer();                          // request timer is stopped on first data package
         return this.sendDataToTokenStreamParser(data);
       },
       message: function() {
-        this.clearRequestTimer();
```
chdh: This is now done in the exit function of the SENT_CLIENT_REQUEST state.

src/connection.js
```diff
@@ -1001,7 +1005,6 @@ class Connection extends EventEmitter {
       this.debug.log(message);
       return false;
     } else {
-      this.clearRequestTimer();
```
chdh: This is now done in the exit function of the SENT_CLIENT_REQUEST state.

Issue #373, opened 2016-03-23
Failed to connect message doesn't take into account default port


jcollum:
src/connection.js line 430

`const message = 'Failed to connect to ' + this.config.server + ':' + this.config.options.port + ' in ' + this.config.options.connectTimeout + 'ms';`

I'm using the default port, which doesn't seem to have made it into to the config.options. Results in an error that is misleading: `Failed to connect to DEF.abc.xyz.com:undefined in 15000ms (err.code=ETIMEOUT)` -- note the 'undefined'.

tvrprasad:
Can you share the full script where you're running into the issue?

Anthropic:
I run the following and get the same error, hope it helps.
node: 5.1.0
npm: 3.3.12
tedious: 1.14.0
tedious-connection-pool: 1.0.5
@jcollum are you using tedious-connection-pool too?

I use **instance name** so port isn't required, if I add port I get a mutual exclusivity error.

There are two separate issues here, the error message could be cleaned up, I have added PR #535 to tackle that issue.

However the main issue is the connection error which is something different, it appears to work for a while **then the timeout occurs**, so it seems like it may be failing even after successful connection.

```bash
ConnectionError: Failed to connect to UMELDT03.corp.uecomm.com.au:undefined in 15000ms
    at ConnectionError (/media/sf_ue/hg/PTN/node/node_modules/tedious/lib/errors.js:12:12)
    at Connection.connectTimeout (/media/sf_ue/hg/PTN/node/node_modules/tedious/lib/connection.js:467:28)
    at Timer.listOnTimeout (timers.js:92:15)
```

Here is my connection script:
```js
  function MssqlDataAccess(config) {
    var mssqlCon = {
      'userName': config.username,
      'password': config.password,
      'server': config.host,
      'options': { 'instanceName': config.instanceName }
    };

    var mssqlPool = { max: config.max, idelTimeOutMills: config.timeout };

    this.pool = new ConnectionPool(mssqlPool, mssqlCon);
    this.sqlLinkedServerName = config.linkServer;
  };

  MssqlDataAccess.prototype._execProc = function(name, params, callback) {
    var self = this;

    this.pool.acquire(function(err, connection) {
      if (err) { return callback(err); }

      var result = {};
      result.records = [];
      result.parameters = {};
      var request = new Request(name, function(err /*, rowCount*/) {
        if (err) { return callback(err); }
        callback(null, result);
        connection.release();
      });

[...]

PR #535, opened 2017-03-09
Connection error to use port or instance name

Anthropic:
Related to and fixes one of two issues raised by #373
Current connection error shows port as undefined wasting time debugging an issue that doesn't exist if the user instead defined instance name. This will show the instanceName if it is defined and there is no port defined.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -549,7 +549,14 @@ class Connection extends EventEmitter {
   }
 
   connectTimeout() {
-    const message = 'Failed to connect to ' + this.config.server + ':' + this.config.options.port + ' in ' + this.config.options.connectTimeout + 'ms';
+    let tail = '';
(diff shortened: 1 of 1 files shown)


Anthropic:
I believe the error in CI is due to #502 not going through CI as it alters connections.js also?

tvrprasad:
I tried the timeout test case without specifying port and I'm getting the right error message:
__ConnectionError: Failed to connect to localhost:1433 in 10ms__

The config code has been reworked quite a bit since the npm published version of Tedious. Try your test case with the latest bits, you shouldn't see the undefined.

> I believe the error in CI is due to #502 not going through CI as it alters connections.js also?

That change is not merged into master yet.

tvrprasad:
Closing the PR as the issue does not exist in the current bits.

PR #537, opened 2017-03-10
EventEmitter.listenerCount is not supported in 0.10 and 0.12.

tvrprasad:
Replaced with EventEmitter.listeners.

@arthurschreiber This'll fix the test failures in https://github.com/tediousjs/tedious/pull/362 for Node 0.10 and 0.12. Please merge.


tvrprasad:
Thanks!

PR #538, opened 2017-03-10
Work-around socket.connect bug in Node 7.7.2

tvrprasad:
Node 7.7.2 introduced a bug in socket.connect, where you get an error:
typeerror: "listener" argument must be a function

Switched to net.connect which does not have this bug.

CI tests are failing for https://github.com/tediousjs/tedious/pull/518 due to this bug.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -14,7 +14,7 @@ const Request = require('./request');
 const RpcRequestPayload = require('./rpcrequest-payload');
 const SqlBatchPayload = require('./sqlbatch-payload');
 const MessageIO = require('./message-io');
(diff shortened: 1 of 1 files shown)


arthurschreiber:
This actually is a bug with Node 7.7.2, so not something we necessarily would have needed to fix in `tedious` itself. 😉

tvrprasad:
That's why I called it a work-around :-) Couple of reasons why I did this:
1) net.connect() is equivalent of what we do now, so this does not introduce a hack.
2) CI tests are failing now due to 7.7.2 upgrade, so we have to do something to unblock merging PRs.

Hope that makes sense.

PR #546, opened 2017-03-16
Fix debug options

chdh:
Since 8b70dbdb37e73eba80c3c39719ce96fee9dedac7 the debug options had no longer an effect.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -63,6 +63,12 @@ class Connection extends EventEmitter {
         cryptoCredentialsDetails: {},
         database: undefined,
         datefirst: DEFAULT_DATEFIRST,
+        debug: {
+          data: false,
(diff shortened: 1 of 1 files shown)


src/connection.js
```diff
@@ -138,6 +144,21 @@ class Connection extends EventEmitter {
         this.config.options.datefirst = config.options.datefirst;
       }
 
+      if (config.options.debug) {
+        if (config.options.debug.data != undefined) {
```
tvrprasad: Thanks for fixing this. Too bad, our recent release does not have this fix.

arthurschreiber:
We can release 1.15.1 once I find some time.
tvrprasad <notifications@github.com> schrieb am Do. 16. März 2017 um 15:57:

PR #547, opened 2017-03-16
feat: Implements streaming BulkLoad

chdh:
This PR extends the public Tedious API with the method `BulkLoad.getRowStream()`. Calling `getRowStream()` switches the `BulkLoad` object into streaming mode and returns a `stream.Writable` that can be used to stream a large amount of rows to the server.

This implementation uses data flow control to prevent memory overload.

The BulkLoad test module has been converted from CoffeeScript to JavaScript and a new test case was added to test the streaming mode.

Implements #523


tvrprasad:
> The BulkLoad test module has been converted from CoffeeScript to JavaScript and a new test case was added to test the streaming mode.

Please send a separate PR for the CoffeeScript to JavaScript conversion. Helps with code review as well as much cleaner history for later. Thanks.

chdh:
> Please send a separate PR for the CoffeeScript to JavaScript conversion. 

OK, it's in #548

arthurschreiber:
Awesome! 😍 I looked over this, and I think this is great and heading in the right direction! 👍 

One thing I'd like to be changed is `BulkLoad#getRowStream`: Couldn't the `BulkLoad` class itself be a `Transform` stream that transforms row objects into the row token data? `BulkLoad#addRow` could then be implemented in terms of `BulkLoad#write`. `Connection#makeRequest` could then pipe the `BulkLoad` object no matter whether it's set in streaming mode or not, and explicitly call `#end` on it if it is not. If it is in streaming mode, it would be the user's task to end the `BulkLoad` stream.

I think I'll be able to release `2.0.0` the coming week (basically `1.15.0` without any of the code supporting non-maintained Node version), and we will hopefully have this as part of `2.1.0`!

Thank you so much for suggesting this change!

chdh:
@arthurschreiber
Thanks for your support.

I understand your idea of using `BulkLoad` as a `Transform` object. But there are some reasons against that:

- The primary purpose of the `BulkLoad` class is to set up the table structure (`addColumn()`) and to provide the `CREATE TABLE` SQL statement (`getTableCreationSql()`).
- To use it as a `Transform`, we would have to make BulkLoad a subclass of `stream.Transform`. `stream.Transform` is quite complex. The `BulkLoad` class would inherit all the methods and events of `stream.Writable` and `stream.Readable`. There are currently 9 events and 14 methods in the public API of `stream.Transform` and there could be more in the future. It's better to separate this complexity from the `BulkLoad` class.
- In the `Connection` class, the `BulkLoad` class is used in lieu of the `Request` class. This is already delicate now, because the `Connection` class adds properties to `this.request` dynamically. While the stream classes keep most of their internal state in `_writableState`, `_readableState` and `_transformState`, they still set some properties like `writable` and `readable`.
- One of the basic principles of modular software design is: When in doubt, separate!
- There is no disadvantage from separating the `BulkLoad` class and the stream. `BulkLoad.addRow()` could still be implemented through `Transform.write()` internally, as you propose.

With this PR, I tried not to touch the existing program logic and guarantee that the previous behaviour in non-streaming mode is preserved.

arthurschreiber:
Hey @chdh (and everyone else who is looking forward to this feature 😄),

just wanted to let you know I've not forgotten about this. ❤️ I still believe this is a great PR, and even though I might have a few reservations about the API / implementation, I think it's a good enough solution that warrants to be merged and iterated upon later.

I'll try and see if I can resolve the current conflicts and then get it merged and released over the next few days.

Again, my sincerest apologies on this taking so long, and thank you for all your work. It is very much appreciated. 😍

chdh:
@arthurschreiber Thanks for your work. Let me know when I can help with anything to finalize this PR.

chdh:
@arthurschreiber 
I have rebased and adapted this PR to the current master branch.
The Travis Lint run failed, but I don't understand why. Can you help?

arthurschreiber:
@chdh The lint checks are failing as we now require commits to stick to the "[conventional](https://conventionalcommits.org/)" commit messages, and enforce this via `commitlint`.

If you could rebase and squash together all the changes under a single commit, and change the message to

```
feature: streaming bulkloads
```

or similar, that'd be great! 🙇

test/integration/bulk-load-test.js
```diff
@@ -248,3 +251,91 @@ exports['bulkLoad - verify null value'] = function(test) {
...
+        }
+      }
+      this.push(null);
+    }
+  }
+};
+
+exports.dummyToSwitchToEncryptedConnection = function(test) {
+  // (Please let me know if you find a better solution for this.)
+  useEncryptedConnection = true;
```
arthurschreiber: I'm wondering why this line would be necessary? 🤔

arthurschreiber: Oh, I see, you wanted to execute one time with an encrypted connection, and one time without. I don't think that is strictly necessary, as the encryption happens on a totally different level in the network stack and I can't imagine any reason why the encryption would have any effect on the Bulkload functionality.

chdh: In message-io.cs, flow-control takes a different path when the connection is encrypted (drain event of securePair.cleartext).

chdh:
@arthurschreiber 
Thanks for your explanation about commitlint. I have updated the commit message.

arthurschreiber:
Hey @chdh,

I updated your branch with the latest changes from `master`, added `flow` type information, and refactored some of the implementation internals. Would you mind taking a look and letting me know what you think? I feel that thanks to the changes made in #768, the bulk stream code got quite a bit easier to wrap your head around.

Regarding the `flow` type changes, I'm planning to extract them into a separate PR (because on their own, they have nothing to do with the streaming). I just felt it's a bit easier to perform some of the refactorings I made with type information being present. 🤷‍♂️

src/bulk-load.js
```diff
@@ -260,4 +275,87 @@ module.exports = class BulkLoad extends EventEmitter {
...
+    this.columnMetadataWritten = false;
+  }
+
+  _transform(row, encoding, callback) {
+    if (!this.columnMetadataWritten) {
+      this.push(this.bulkLoad.getColMetaData());
+      this.columnMetadataWritten = true;
+    }
+
+    const buf = new WritableTrackingBuffer(10, 'ucs2', true);
```
chdh: 10 bytes are a bit small for the start. Even when writing only 2 integer values, we already need 11 bytes and `WritableTrackingBuffer.makeRoomFor()` will copy and expand the buffer to 128 bytes.

arthurschreiber:
> Is there any hard protection against the case, that an old stream pipe from an old request is still active and sending data that interfears with the currently active request?

Yes - this can never happen. `Message` instances are written to the `OutGoingMessageStream` via a call to `.write`, and when other `.write` calls happen before that first `Message` is fully consumed, the message object will be buffered internally and not consumed.

You can see this in the `._write` implementation here:

https://github.com/tediousjs/tedious/blob/master/src/outgoing-message-stream.js#L30-L77

Until `callback` is called, no further `._write` call will be scheduled, so there is no risk of different messages interfering. 👍

chdh:
@arthurschreiber 
You are right. Thanks for the explanation.

src/bulk-load.js
arthurschreiber: I changed this to 64 byte. How do you feel about that size? 🙇

BrunoLopes:
Please, can anyone give me an example of how to implement streaming BulkLoad?! 
I'm trying to use it with the code below:

...

const bulkLoad = connection.newBulkLoad('dbo.Adjustments', options, function (error, rowCount) {
	 if(error){
			context.log('bulkloadError=>', error);
		}
	context.log('inserted %d rows', rowCount);
});

bulkLoad.addColumn('XPTO', TYPES.VarChar, { length: 11, nullable: false });
bulkLoad.addColumn('XPTY', TYPES.NVarChar, { length: 7, nullable: true });
        
const rowStream = bulkLoad.getRowStream();

//geting csv records to streaming to the table        
csv()
.fromString(myBlob.toString())
.subscribe(function(jsonObj) { //single json object will be emitted for each csv line
	return new Promise(function(resolve, reject) {
		resolve()
		// bulkLoad.addRow({
		//     CPF: jsonObj.CPF, MesReferencia: jsonObj.MesReferencia, ValorTv: jsonObj.ValorTv, ValorBandaLarga: jsonObj.ValorBandaLarga, ValorVoz: jsonObj.ValorVoz,
		//     ValorOutros: jsonObj.ValorOutros, PercentualTv: jsonObj.PercentualTv, PercentualBandaLarga: jsonObj.PercentualBandaLarga, PercentualVoz: jsonObj.PercentualVoz, PercentualOutros: jsonObj.PercentualOutros
		// });
		rowStream.write({
			CPF: jsonObj.CPF, MesReferencia: jsonObj.MesReferencia, ValorTv: jsonObj.ValorTv, ValorBandaLarga: jsonObj.ValorBandaLarga, ValorVoz: jsonObj.ValorVoz,
			ValorOutros: jsonObj.ValorOutros, PercentualTv: jsonObj.PercentualTv, PercentualBandaLarga: jsonObj.PercentualBandaLarga, PercentualVoz: jsonObj.PercentualVoz, PercentualOutros: jsonObj.PercentualOutros
		},"utf-8", function(){
			context.log('Escreveu dados.');
		});
	})
}, function (eror){ context.log('erro subscribe-> ', error) },function(){});

connection.execBulkLoad(bulkLoad);
rowSource.pipe(rowStream);

What am I doing wrong? Please help me.

chdh:
@BrunoLopes Don't call `rowSource.pipe(rowStream)` when you fill the stream yourself by calling `rowStream.write()`. Call `rowStream.end()` when there is no more data.

If that doesn't help, construct a simpler test program without the csv reading part. Just write a few constant rows to the stream.

BrunoLopes:
Thanks so much @chdh !!!

PR #548, opened 2017-03-16
Convert BulkLoad test from CoffeeScript to JavaScript

chdh:
For this conversion from CoffeScript to JavaScript, the test cases have been restructured with separate setUp/tearDown for nodeunit, but the functionality of the tests has not been changed.


test/integration/bulk-load-test.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+exports.setUp = function(setUpDone) {
+  const connection = new Connection(getConfig());
+  connection.on('connect', (err) => {
+    if (err) {
+      setUpDone(err);
+      return;
+    }
+    this.connection = connection;
```
tvrprasad: Can you please sprinkle some new lines in the code at places you consider logical? It's much easier to read the code in an editor when there is that white space separation. Thanks.

tvrprasad: I think one is ok actually, the functions are not too long. For longer functions, we should have some white space separation between logical chunks of code.

PR #559, opened 2017-03-28
Use `babel-eslint` as the parser for `eslint`

arthurschreiber:
This basically allows `eslint` to parse any source code that can also be parsed by `babel`. This is  mainly in preparation for adding type checking with `flow`.

It also removes the `use strict` declarations from all files, because these are automatically added by `babel` anyway.


tvrprasad:
@arthurschreiber Thanks for the flurry of house cleaning tasks and releasing 2.0!
🥇

PR #564, opened 2017-04-09
ci: Run tests against sqlserver on linux when building on Travis.

arthurschreiber:


arthurschreiber:
Once this is merged, we have full integration test builds running both on AppVeyor and Travis. 👍

PR #565, opened 2017-04-10
Automatic releases via `semantic-release`

arthurschreiber:
Now that we run the full integration test suite across all supported Node.JS versions on travis (#564), we can use `semantic-release` to automatically publish new versions of `tedious` after each merge to `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ node_js:
   - "6"
   - "7"
 
+branches:
+  except:
+    - /^v\d+\.\d+\.\d+$/
+
 services:
   - docker
 
@@ -32,6 +36,9 @@ before_install:
       }
     }' > ~/.tedious/test-connection.json
 
+before_script:
+  - npm prune
+
 script:
   - npm run test
   - TEDIOUS_TDS_VERSION=7_4   npm run test-integration
@@ -39,3 +46,6 @@ script:
   - TEDIOUS_TDS_VERSION=7_3_A npm run test-integration
   - TEDIOUS_TDS_VERSION=7_2   npm run test-integration
   - TEDIOUS_TDS_VERSION=7_1   npm run test-integration
+
+after_success:
+  - npm run semantic-release
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "homepage": "https://github.com/tediousjs/tedious",
   "bugs": "https://github.com/tediousjs/tedious/issues",
   "license": "MIT",
-  "version": "2.0.0",
+  "version": "0.0.0-dev",
   "main": "./lib/tedious.js",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint": "^3.18.0",
     "mitm": "^1.3.2",
     "nodeunit": "^0.11.0",
+    "semantic-release": "^6.3.2",
     "sinon": "^1.17.5"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "test-all": "nodeunit --reporter minimal test/setup.js test/unit/ test/unit/token/ test/unit/tracking-buffer test/integration/",
     "pretest-integration": "eslint src test",
     "test-integration": "nodeunit --reporter minimal test/setup.js test/integration/",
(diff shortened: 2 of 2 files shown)


arthurschreiber:
If we want to go forward with this (which I really do, it will make releases a much more frequent thing, and will require less time), there's a few caveats. `semantic-release` depends on the commit message contents to figure out what the version number for the next release will be. This means that going forward, commit messages should be formatted according to the rules specified here: https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit

I understand that not all external contributions will be following this style and those contributions' commit messages will need to be fixed/cleaned up before they can be merged, but this should not be an insurmountable task. 😄

tvrprasad:
@arthurschreiber Currently we have some stabilization period before we do a release, gives us a chance to identify and fix issues. Should it be a concern that we lose that which could result in buggy releases?

Another question: If a commit message does not have the right format and the next commit has the right format, would the release include both the commits or just the second one?

arthurschreiber:
> @arthurschreiber Currently we have some stabilization period before we do a release, gives us a chance to identify and fix issues. Should it be a concern that we lose that which could result in buggy releases?

We could use npm `dist-tags` to still support "stable" and "unstable" releases. We would push releases automatically via `semantic-release`, but only tag them as `latest` whenever we see fit.

Users could then either follow the `latest` (default) or the `unstable` dist-tag, based on their preference.

If we can, I'd try without this first. I don't think we really ever had such a "stabilization" period, as almost no-one installs tedious from source in their application, so the occurrence of someone finding a bug before we pushed a release was either very low or zero.

> Another question: If a commit message does not have the right format and the next commit has the right format, would the release include both the commits or just the second one?

Yes. All commits (whether their message is correct or not) are part of the release. The commit messages only determine whether it will be a major, minor or bugfix release. Major releases only happen if a commit is marked a `BREAKING CHANGE`, minor releases happen if at least one commit is labeled as a `feat` (new feature), all other commits (whether they follow the format or not) will only cause a bump in the patch version.

We could just go ahead and give this a try, and recalibrate if we run into major issues. What do you think?

tvrprasad:
> If we can, I'd try without this first. I don't think we really ever had such a "stabilization" period, as almost no-one installs tedious from source in their application, so the occurrence of someone finding a bug before we pushed a release was either very low or zero.

My impression is that there are enough beta users in the community checking out the latest bits that we identify any major issues before release. This is based on what I see with folks checking out PRs while in flight as well as those who ask for PRs to be pushed to master so they can pick it up.

With adequate level of automated tests, I can see semantic release working well. I'm not sure we have that level of testing. And without that, it makes me nervous for pushing out a release on every commit.
 
"stable"/"unstable" sounds a bit more comforting. Of course the question is how we go from "unstable" to "stable". Here is one approach:
- Agree on a release cadence, say once every 2 months.
- We tag a version as pre-release, say on 6/1/17.
- Let it sit there for 2 months for beta users to try.
- If there are no major issues reported on the pre-release, we release that on 8/1/17.

Something like that. Unstable (daily), Pre-release (agreed upon cadence), Release (same cadence).

arthurschreiber:
> My impression is that there are enough beta users in the community checking out the latest bits that we identify any major issues before release. This is based on what I see with folks checking out PRs while in flight as well as those who ask for PRs to be pushed to master so they can pick it up.

That might be true, but the involvement here has been pretty low. Most times, people check out PRs and ask for them being included without testing, and only if people really need some functionality, they'll go through the trouble of testing the PR branch. But even then, I can't remember really fixing any functionality after it was merged to master because a user ran into an issue. I'm not saying this is something that will or can never happen, but it didn't happen in any way that left any impression on me in the two or so years I've been maintaining `tedious`.

My plan for now is to have every push to `master` result in a release, but be tagged as `next` and not latest. Then we can set the `latest` tag whenever we see fit. This will actually make it a **lot** simpler for people to test-drive new features in `tedious` and report back any issues.

tvrprasad:
@arthurschreiber How exactly is this supposed to work? I just did an <code>npm install tedious@next</code> and <code>npm install tedious@latest</code> and did a diff between the tedious\lib directories of both installs. I don't see any of the recent commits showing up in <code>@next</code> install. Am I missing something here? Thanks.

Issue #357, opened 2016-01-28
Unexpected behavior during request timeout


gshively11:
Hi, I'm seeing some unexpected behavior during request timeouts while using the mssql library.  I've traced the code down to the tedious module.  Here's my scenario:
1. Open a connection to the database
2. Make a request to call a stored procedure
3. Get result, everything is good.
4. Go blackhole the database IP, so that no outgoing connections are allowed from my box to that IP.
5. Initiate another request.
6. After the configured requestTimeout (3 seconds), I see the [request timeout](https://github.com/pekim/tedious/blob/master/src/connection.js#L433) triggering, which sets a 'SentAttention' state.
7. At this point, things just hang for a little bit, until what I assume is the socket KeepAlive notices that it can't get a response and triggers an error on the socket.  At this point, the connection reports back something is wrong.

The unexpected bit is where the request timeout doesn't immediately trigger an error event on the connection and cancel the request, so that the mssql module can report back to the consumer that the request failed due to timeout.  Is this behavior intentional?  If so, can you explain a little bit of the rationale?  Is there another way I should be setting a hard limit on request time?  Thanks in advance.

arthurschreiber:
Hey @gshively11,

We've not implemented the "Cancel Timeout" correctly in Tedious. See https://msdn.microsoft.com/en-us/library/dd341108.aspx

This is definately something we should get fixed and support correctly.

arthurschreiber:
Here's another line from the specification (https://msdn.microsoft.com/en-us/library/dd356487.aspx):

> If a TDS client implements the Client Request Timer and the timer times out, then the TDS client MUST send an Attention message to the server and enter the "Sent Attention" state.

So we're handling the request timeout correctly.

But on the next lane, it says:

>  If a TDS client implements the Cancel Timer and the timer times out, then the TDS client MUST close the underlying connection, indicate the error to the upper layer, and enter the "Final State" state.

I also checked the code again to be sure. Although we have a cancelTimeout option, it's not honored anywhere. Adding the timeout should not be too hard.

gshively11:
@arthurschreiber Thanks for clarifying, that makes sense now.  I'll take a crack at a PR if I have time, but it may not be for a little while.  I'd love to see this feature if you have the bandwidth!  In the mean time, we can just rely on the socket error for now.

IanChokS:
Feature merged in #846

PR #566, opened 2017-04-11
Feature/cancellation timer

gsterjov:


gsterjov:
Mainly to address the issue at: https://github.com/tediousjs/tedious/issues/357

Adds a cancellation timeout which will close the connection once elapsed. Request timeouts will also call `this.cancel()` rather than going straight to `SentAttention`, so a request timeout with a failed cancellation will close the connection

arthurschreiber:
@gsterjov I just fixed an issue with stack traces being incorrect for tests written in Coffee Script (see #567). Would you mind merging in the latest `master` changes or rebasing your changes on top of `master`? That would help figure out what test cases exactly are failing the Travis and AppVeyor builds.

src/connection.js
```diff
@@ -585,8 +594,12 @@ class Connection extends EventEmitter {
...
   requestTimeout() {
     this.requestTimer = undefined;
-    this.messageIo.sendMessage(TYPE.ATTENTION);
-    return this.transitionTo(this.STATE.SENT_ATTENTION);
+    this.cancel();
+  }
+
+  cancelTimeout() {
+    this.cancelTimer = undefined;
+    return this.close();
```
arthurschreiber: I don't think the `return` here is necessary.

Also, on top of that, I'm not sure whether just calling `this.close` is the best course of action here. The in-flight request that did not get cancelled in time will get a `RequestError`, but it will not be clear that this error was passed because cancellation timed out.

src/connection.js
```diff
@@ -585,8 +594,12 @@ class Connection extends EventEmitter {
 
   requestTimeout() {
     this.requestTimer = undefined;
-    this.messageIo.sendMessage(TYPE.ATTENTION);
-    return this.transitionTo(this.STATE.SENT_ATTENTION);
+    this.cancel();
```
tvrprasad: This will set canceled=true which in turn will cause the error message to be Canceled instead of Timeout. You should have three different error messages for the three different cases - user initiated cancel(), request timer timeout and cancel timer timeout (i.e. timeout waiting for Attention acknowledgement)

tvrprasad: There are actually two flavors of cancel timer timeout error messages:
1) Attention sent in response to request timeout.
2) Attention sent in response to user initiated cancel()

src/connection.js
```diff
@@ -249,6 +249,7 @@ class Connection extends EventEmitter {
     this.socketError = this.socketError.bind(this);
     this.requestTimeout = this.requestTimeout.bind(this);
     this.connectTimeout = this.connectTimeout.bind(this);
+    this.cancelTimeout = this.cancelTimeout.bind(this);
```
tvrprasad: A test would be nice though I can't think of a way to add it...

src/connection.js
gsterjov: Good point. I'll try get it to return an appropriate error message

src/connection.js
gsterjov: Great, thanks for the help! I'm thinking its possible to still have the single cancel timer but to stick some bool flags to determine what kind of cancellation it is.

src/connection.js
gsterjov: I'll see what I can do. Might be a bit contrived but at the very least it would be good to test the various types of cancellation

ihgrant:
As far as I can tell, this patch fixes a serious bug that my team was experiencing in production with the `mssql` package that depends on this library.

Is there anything I can do to help get this PR merged?

arthurschreiber:
@ihgrant Thanks for the ping. I'll take a look and will try to get this merged asap! ❤️

arthurschreiber:
The cancellation rework in https://github.com/tediousjs/tedious/pull/846 also implements a proper request cancellation timer. 👍

gsterjov:
Nice!

PR #574, opened 2017-05-08
Transient failure retries

tvrprasad:
@arthurschreiber @chdh This adds retries for transient failures that all Microsoft drivers do to improve reliability. This could address some of the connection failures issues that have been reported over time.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -358,16 +358,16 @@ class Connection extends EventEmitter {
           this.request.error['class'] = token['class'];
           this.request.error.serverName = token.serverName;
           this.request.error.procName = token.procName;
-          return this.request.error.lineNumber = token.lineNumber;
+          this.request.error.lineNumber = token.lineNumber;
         }
       } else {
         const isLoginErrorTransient = this.transientErrorLookup.isTransientError(token.number);
         if (isLoginErrorTransient && this.curTransientRetryCount !== this.config.options.maxRetriesOnTransientErrors) {
           this.debug.log('Initiating retry on transient error = ', token.number);
-          return this.transitionTo(this.STATE.TRANSIENT_FAILURE_RETRY);
+          this.transitionTo(this.STATE.TRANSIENT_FAILURE_RETRY);
+        } else {
+          this.loginError = ConnectionError(token.message, 'ELOGIN');
         }
-
-        return this.loginError = ConnectionError(token.message, 'ELOGIN');
       }
     });
 
--- a/src/connection.js
+++ b/src/connection.js
@@ -313,9 +313,7 @@ class Connection extends EventEmitter {
       this.closeConnection();
       if (cleanupTypeEnum === this.cleanupTypeEnum.REDIRECT) {
         this.emit('rerouting');
-      } else if (cleanupTypeEnum === this.cleanupTypeEnum.RETRY) {
-        this.emit('retry');
-      } else {
+      } else if (cleanupTypeEnum !== this.cleanupTypeEnum.RETRY) {
         this.emit('end');
       }
       if (this.request) {
@@ -634,6 +632,7 @@ class Connection extends EventEmitter {
 
   retryTimeout() {
     this.retryTimer = undefined;
+    this.emit('retry');
     this.transitionTo(this.STATE.CONNECTING);
   }
 
--- a/src/connection.js
+++ b/src/connection.js
@@ -2,6 +2,7 @@ const BulkLoad = require('./bulk-load');
 const Debug = require('./debug');
 const EventEmitter = require('events').EventEmitter;
 const InstanceLookup = require('./instance-lookup').InstanceLookup;
+const TransientErrorLookup = require('./transient-error-lookup.js').TransientErrorLookup;
 const TYPE = require('./packet').TYPE;
 const PreloginPayload = require('./prelogin-payload');
 const Login7Payload = require('./login7-payload');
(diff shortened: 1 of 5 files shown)


tvrprasad:
Some tests failing on Linux for NodeJS 4. Error suggests issue with machine state than code:
> Failed to connect to localhost:1433 in 15000ms

Is there some way to retrigger the test without submitting a dummy change?

tvrprasad:
I made a dummy PR (https://github.com/tediousjs/tedious/pull/575) with the same changes which triggered a re-run and tests pass now.

src/transient-error-lookup.js
```diff
@@ -0,0 +1,17 @@
+class TransientErrorLookup {
```
chdh: Why is the isTransientError() function within a class?

src/transient-error-lookup.js
```diff
@@ -0,0 +1,17 @@
...
+  isTransientError(error) {
+    // This list of transient errors comes from Microsoft implementation of SqlClient:
+    //  - https://github.com/dotnet/corefx/blob/master/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlInternalConnectionTds.cs#L115
+    const transientErrors = [4060, 10928, 10929, 40197, 40501, 40613];
+
+    for (let j = 0; j < transientErrors.length; j++) {
```
chdh: Why not use Array.indexOf()? It's available in Node 4.8.3.

src/connection.js
```diff
@@ -331,6 +361,12 @@ class Connection extends EventEmitter {
           return this.request.error.lineNumber = token.lineNumber;
         }
       } else {
+        const isLoginErrorTransient = this.transientErrorLookup.isTransientError(token.number);
+        if (isLoginErrorTransient && this.curTransientRetryCount !== this.config.options.maxRetriesOnTransientErrors) {
+          this.debug.log('Initiating retry on transient error = ', token.number);
+          return this.transitionTo(this.STATE.TRANSIENT_FAILURE_RETRY);
```
chdh: The "return" statements with a value are an artefact of the CoffeeScript to JavaScript conversion and should not be used whereit makes no sense.

src/connection.js
```diff
@@ -331,6 +361,12 @@ class Connection extends EventEmitter {
           return this.request.error.lineNumber = token.lineNumber;
         }
       } else {
+        const isLoginErrorTransient = this.transientErrorLookup.isTransientError(token.number);
+        if (isLoginErrorTransient && this.curTransientRetryCount !== this.config.options.maxRetriesOnTransientErrors) {
```
chdh: Better use `<` than `!==` for comparing a retry count.

src/connection.js
```diff
@@ -663,6 +717,12 @@ class Connection extends EventEmitter {
     if (this.state === this.STATE.REROUTING) {
       this.debug.log('Rerouting to ' + this.routingData.server + ':' + this.routingData.port);
       return this.dispatchEvent('reconnect');
+    } else if (this.state === this.STATE.TRANSIENT_FAILURE_RETRY) {
+      const server = this.routingData ? this.routingData.server : this.server;
+      const port = this.routingData ? this.routingData.port : this.config.options.port;
+      this.debug.log('Retry after transient failure connecting to ' + server + ':' + port);
+
+      return this.dispatchEvent('retry');
```
chdh: Why not just call `createRetryTimer()` here? `dispatchEvent()` is executed synchronously.

test/unit/transient-error-lookup-test.js
```diff
@@ -0,0 +1,20 @@
+const TransientErrorLookup = require('../../src/transient-error-lookup').TransientErrorLookup;
+
```
chdh: This test case seems very superfluous to me for such a simple function.

test/integration/connection-retry-test.js
```diff
@@ -0,0 +1,104 @@
...
+    this.sinon.restore();
+    done();
+  },
+
+  'retry specified number of times on transient errors': function(test) {
+    const config = getConfig();
+    test.expect(config.options.maxRetriesOnTransientErrors + 1);
+
+    this.sinon.stub(TransientErrorLookup.prototype, 'isTransientError', () => {
+      return true;
```
chdh: Why not test for the error code for "invalid password" here?

src/connection.js
tvrprasad: Done.
Added a clean-up item to remove these across the board - https://github.com/tediousjs/tedious/issues/576

src/connection.js
tvrprasad: I think that'd make it harder to follow the logic. Now it's easy to see what's happening by looking at code under <code>TRANSIENT_FAILURE_RETRY</code> and back track from there.

src/transient-error-lookup.js
tvrprasad: This makes it easier and cleaner to stub it out in the tests. Add a comment to clarify that.

src/transient-error-lookup.js
tvrprasad: Done.

test/integration/connection-retry-test.js
tvrprasad: Done.

test/unit/transient-error-lookup-test.js
tvrprasad: Makes it even more silly now that I changed the code to indexOf :-)
This is just an assertion to make sure any additions or deletion from list of transient errors is intentional.

tvrprasad: btw, this test actually caught a bug when I switched to Array.indexOf per your suggestion. I had <code>return transientErrors.indexOf(error) === -1;</code> where I should have <code>return transientErrors.indexOf(error) !== -1;</code> and this test failed.

tvrprasad:
@chdh Addressed/responded to your comments. If you don't have more comments I'll commit this. Thanks!

PR #584, opened 2017-05-18
Adding survey link to README.md

LuisBosquez:


tvrprasad:
@LuisBosquez I tried rendering this using a visual studio extension and the image does not align correctly. Also, I think the image does not quite gel with the content of the page. Would just a link do?


image

LuisBosquez:
@tvrprasad updated to just a link to avoid rendering issues

PR #586, opened 2017-05-25
chore: Only run `eslint` once.

arthurschreiber:
Instead of running `eslint` before `test`, `test-integration` and
`test-all` scripts, we only run it before `test`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,9 +66,7 @@
   "scripts": {
     "pretest": "eslint src test",
     "test": "nodeunit --reporter minimal test/setup.js test/unit/ test/unit/token/ test/unit/tracking-buffer",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -66,9 +66,7 @@
   "scripts": {
     "pretest": "eslint src test",
     "test": "nodeunit --reporter minimal test/setup.js test/unit/ test/unit/token/ test/unit/tracking-buffer",
-    "pretest-all": "eslint src test",
     "test-all": "nodeunit --reporter minimal test/setup.js test/unit/ test/unit/token/ test/unit/tracking-buffer test/integration/",
```
tvrprasad: It may make sense to run eslint for test-all. It's _all_, right? :-)

PR #596, opened 2017-06-27
Supports sql_varaint for variable length data types

### LINKED ISSUE #593: SELECT ServerProperty('Edition') not working
Coming from there https://github.com/patriksimek/node-mssql/issues/496 but it seems a tedious bug (or _missing feature_?)

This query just works

```sql
select @@version as Version
```
This doesnt:

```sql
select SERVERPROPERTY('Edition') as Version
```

Can it be fixed/added to the library?

v-suhame:
also fixes #593

### DIFF
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -335,7 +335,8 @@ function valueParse(parser, metaData, options, callback) {
                     valueMetaData.collation = collation;
                     if (baseType === 0xA5 || baseType === 0xAD || baseType === 0xA7 || baseType === 0xAF || baseType === 0xE7 || baseType === 0xEF) {
                       return readDataLength(parser, valueMetaData.type, {}, null, (maxDataLength) => {
-                        valueMetaData.dataLength = maxDataLength;
+                        // skip the 2-byte max length sent for BIGVARCHRTYPE, BIGCHARTYPE, NVARCHARTYPE, NCHARTYPE, BIGVARBINTYPE and BIGBINARYTYPE types 
+                        // and parse based on the length of actual data 
                         return valueParse(parser, valueMetaData, options, callback);
                       });
                     } else {
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -335,8 +335,8 @@ function valueParse(parser, metaData, options, callback) {
                     valueMetaData.collation = collation;
                     if (baseType === 0xA5 || baseType === 0xAD || baseType === 0xA7 || baseType === 0xAF || baseType === 0xE7 || baseType === 0xEF) {
                       return readDataLength(parser, valueMetaData.type, {}, null, (maxDataLength) => {
-                        // skip the 2-byte max length sent for BIGVARCHRTYPE, BIGCHARTYPE, NVARCHARTYPE, NCHARTYPE, BIGVARBINTYPE and BIGBINARYTYPE types 
-                        // and parse based on the length of actual data 
+                        // skip the 2-byte max length sent for BIGVARCHRTYPE, BIGCHARTYPE, NVARCHARTYPE, NCHARTYPE, BIGVARBINTYPE and BIGBINARYTYPE types
+                        // and parse based on the length of actual data
                         return valueParse(parser, valueMetaData, options, callback);
                       });
                     } else {
--- a/test/integration/datatypes-in-results-test.js
+++ b/test/integration/datatypes-in-results-test.js
@@ -506,6 +506,10 @@ exports.variantVarChar = function(test) {
   execSql(test, "select cast('abc' as sql_variant)", 'abc', '7_2');
 };
 
+exports.variantVarChar2 = function(test) {
+  execSql(test, "select SERVERPROPERTY('LicenseType') as LicenseType", 'DISABLED', '7_2');
+};
+
 exports.variantVarBin = function(test) {
   execSql(
     test,
--- a/test/integration/datatypes-in-results-test.js
+++ b/test/integration/datatypes-in-results-test.js
@@ -527,6 +527,25 @@ exports.variantDateTimeOffset = function(test) {
   );
 };
 
+
+exports.variantMultipleDatatypes = function(test) {
+  const sql = `\
+  create table #tab1 (
+  [c0] [int] IDENTITY(1,1),
+  [c1] [sql_variant] NULL);
+  insert into #tab1 ([c1]) values (N'abcdШ');
+  insert into #tab1 ([c1]) select cast(3148.29 as decimal(20,8));	
+  insert into #tab1 ([c1]) select cast(0x1234 as varbinary(16));
+  insert into #tab1 ([c1]) select cast('01234567-89AB-CDEF-0123-456789ABCDEF' as uniqueidentifier); 
+  insert into #tab1 ([c1]) values (0.00000090000000000);	--decimal(38,17);
+  insert into #tab1 ([c1]) select cast('2011-12-4 10:04:23' as datetime);
+  insert into #tab1 ([c1]) select cast('abcde' as varchar(10));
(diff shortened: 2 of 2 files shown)


v-suhame:
Current tests for variant are using `cast`, so this issue wasn't captured in them. SQL Server returns max length as exact data length while using `cast`, or other keywords that might involves processing of data. 

For regular `SELECT` queries, max length(in `VARIANT_PROPERTIES`) is sent as max length of data type.

`select cast('abc' as sql_variant);`  -- max length is 3
`select [variantColumn] from table where  [variantColumn]='hello' ;`  --  max length is 8000

src/value-parser.js
```diff
@@ -335,7 +335,8 @@ function valueParse(parser, metaData, options, callback) {
                     valueMetaData.collation = collation;
                     if (baseType === 0xA5 || baseType === 0xAD || baseType === 0xA7 || baseType === 0xAF || baseType === 0xE7 || baseType === 0xEF) {
                       return readDataLength(parser, valueMetaData.type, {}, null, (maxDataLength) => {
-                        valueMetaData.dataLength = maxDataLength;
+                        // skip the 2-byte max length sent for BIGVARCHRTYPE, BIGCHARTYPE, NVARCHARTYPE, NCHARTYPE, BIGVARBINTYPE and BIGBINARYTYPE types
+                        // and parse based on the length of actual data
```
tvrprasad: I just looked at the TDS spec for variant specification and it doesn't seem to match the code, at least as I read it. See '2.2.5.5.4 SQL_VARIANT Values' in the spec. Perhaps this code needs a bigger overhaul?

v-suhame: I did look into '2.2.5.5.4 SQL_VARIANT Values' while investigating. It isn't clear why `max length` is passed and what it should be used for, as it is not the actual data length in TDS packet. 

Say for this query that returns variant, `select SERVERPROPERTY('Edition') as Version` , in TDS stream the max length portion of `VARIANT_PROPERTIES` is `8000`(wtr varchar), so there is a wait for all `8000` bytes to arrive, whereas the actual data has already arrived. This wait might also  consume data of different column or row(to fill 8k buffer). So, it is safe to read based on actual length.

I stepped into SqlClient to double check, and noticed those bytes were skipped with the comment 'skip over unknown properties'.
I did ad hoc testing with multiple rows and columns of variant and non-variant data mixture, dint run into any issue. Let me know if there is any scenario that you think might cause problem or needs to be tested.

P.S: Is the Travis failure intermittent?

tvrprasad: The parsing code seems to assume there is always baseType (1 byte), propBytes (1 byte), precision(1 byte), scale(1 byte), collation(5 bytes), dataLength(2 bytes) and then data. See https://github.com/tediousjs/tedious/blob/master/src/value-parser.js#L333.

Whereas the spec seems to say something else, at least as I read it. The layout is more dynamic. For example, propBytes value is 0 for GUIDTYPE per spec, which means there is no precision or scale and the above code would break. See below:
image

The code doesn't seem to match anything in the spec. Either that or I'm missing something big time :-)

Perhaps one test could be a Variant with GUIDTYPE contained in it. You can try a few other Variant base types from the table and see if they all work.

And yes, there is some flakiness with Travis test I've seen before.

v-suhame: Now I get what you are referring to! Calls are made to readPrecision/readScale/readCollation functions, within these functions bytes are read only if datatype supports precision/scale/collation respectively. I tested all the groups in `VARIANT_BASETYPE`, they are working fine ✌️

tvrprasad: That makes sense. Thanks! Can you please add the tests you wrote to the set of automated tests?

This is still a deviation from spec though. You're ignoring maxDataLength and instead relying on the globally defined MAX value. If you go by spec, you'd pass maxDataLength as a parameter to valueParser and use that instead of MAX in valueParser implementation. So the code would look something like this:

```JavaScript
                    if (baseType === 0xA5 || baseType === 0xAD || baseType === 0xA7 || baseType === 0xAF || baseType === 0xE7 || baseType === 0xEF) {
                      return readDataLength(parser, valueMetaData.type, {}, null, (maxDataLength) => {
                        return valueParse(parser, valueMetaData, options, maxDataLength + 1, callback);
                      });
                    } else {
                      return valueParse(parser, valueMetaData, options, MAX, callback);
                    }
```

Can you see how this is done SqlClient? Perhaps, we could mimic that here.

v-suhame: Sure, I'll push the tests after writing them in nodeunit. 

Some clarification - this PR is not using the globally defined MAX value. It uses [valueMetaData.dataLength](https://github.com/tediousjs/tedious/pull/596/files#diff-6ee9b395d11ea3cffaecde0296efd5aaR328), this is the data length derived from column header of the row.

Details on why this is preferred over the maxDataLength is in https://github.com/tediousjs/tedious/pull/596#discussion_r124680598. SqlClient too reads the maxDataLength(just to skip over it, never uses it), actual data is read using the length in column header.

This PR essentially would be something like this, if it was in structured programming

```node
if (baseType in BIGVARCHRTYPE || BIGCHARTYPE || NVARCHARTYPE || NCHARTYPE || BIGVARBINTYPE || BIGBINARYTYPE) {
  // ignore maxDataLength - we already have all necessary detail to call valueParse
  maxDataLength  = readDataLength(parser, valueMetaData.type, {}, null) 
}
valueParse(parser, valueMetaData, options, callback);
```

tvrprasad: > Some clarification - this PR is not using the globally defined MAX value. It uses valueMetaData.dataLength, this is the data length derived from column header of the row.

Sorry I was not clear. I understand the code uses valueMetaData.dataLength for data length. What I'm referring is the globally used MAX, for example <code>if (metaData.dataLength === MAX) {</code> under <code>case 'Char'</code> in <code>valueParse</code>. Per spec it seems like for Variant case instead of MAX, the max length read from the stream should be used.

If SqlClient is ignoring that, I think it's fine if we're doing the same here.

test/integration/datatypes-in-results-test.js
```diff
@@ -523,6 +527,25 @@ exports.variantDateTimeOffset = function(test) {
...
+  insert into #tab1 ([c1]) values (N'abcdШ');
+  insert into #tab1 ([c1]) select cast(3148.29 as decimal(20,8));	
+  insert into #tab1 ([c1]) select cast(0x1234 as varbinary(16));
+  insert into #tab1 ([c1]) select cast('01234567-89AB-CDEF-0123-456789ABCDEF' as uniqueidentifier); 
+  insert into #tab1 ([c1]) values (0.00000090000000000);	--decimal(38,17);
+  insert into #tab1 ([c1]) select cast('2011-12-4 10:04:23' as datetime);
+  insert into #tab1 ([c1]) select cast('abcde' as varchar(10));
+  select [c1] from #tab1 ORDER BY [c0];
+  `;
+  const expectedValues = ['abcdШ', 3148.29, new Buffer([0x12, 0x34]), '01234567-89AB-CDEF-0123-456789ABCDEF', 0.00000090000000000, new Date('December 4, 2011 10:04:23 GMT'), 'abcde'];
```
tvrprasad: Can you re-format this so the line is not so long? This'll also retrigger the flaky TravisCI tests and hopefully pass this time!

Thanks for adding the tests. We can merge this once the tests pass.

src/value-parser.js
v-suhame: Thanks for clarifying that! 

Passing maxDataLength and using it as MAX doesn't always work. For instance, `select cast('abc' as sql_variant);` returns maxDataLength as 3, in this case `(metaData.dataLength === MAX)` so flow will enter into `readMaxChars` to read PLP stream. `2-byte max length` mentioned in the TDS is flaky, changes based on SQL.

PR #597, opened 2017-06-30
nanasecondDelta rounded sometimes incorrect

peter-int-veld:
NanasecondDelta rounded sometimes incorrect, for example with this date/time: ‘2017-06-29 17:20:03.5036264’. This was caused by an addition with more than 6 decimal digits, which is not exact in JS. Fixed by executing the addition first and then the division.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "babel-register": "^6.24.0",
     "benchmark": "^2.1.0",
     "eslint": "^3.19.0",
+    "eslint-plugin-react": "^7.1.0",
     "mitm": "^1.3.2",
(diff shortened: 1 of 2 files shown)


peter-int-veld:
Forget this solution, with different numbers the problem is still there. The problem is that multiplying with floating point gives rounding errors. The nanosecond delta is therefore not precise. This gives problems in our customer project with WHERE conditions.

peter-int-veld:
Have tried it with a lib like bignumber.js. This was also not successful. The issue is that the time with nanoseconds has more than 15 significant digits.

peter-int-veld:
Will create a new pull request with the correct solution

Issue #599, opened 2017-07-03
Rounding error on nanosecondDelta


peter-int-veld:
NanosecondDelta  is rounded incorrect sometimes for the date/time datatype. For example with this date/time: ‘2017-06-29 17:20:03.5036264’. This is caused multiplication rounding errors.

peter-int-veld:
I added a pull request with a fix.

LuisBosquez:
@peter-int-veld, can you open a PR against this project? I think you linked to a commit on your fork and I can't find the request on this project

tvrprasad:
@LuisBosquez There are actually 2 links. https://github.com/tediousjs/tedious/issues/599 is the link to the PR.

tvrprasad:
Never mind :-) Here is the link to the PR - https://github.com/tediousjs/tedious/pull/600

arthurschreiber:
This was fixed by #600. `tedious@2.1.3` was released on the `next` channel containing the fix for this. You can install this version via `npm install tedious@2.1.3` or `npm install tedious@next`. Could you double check and let me know if that solves your problem?

PR #600, opened 2017-07-03
Fix for issue #599 (Rounding error on nanosecondDelta)

peter-int-veld:
Fixed by first executing the multiplication and then the addition.

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1286,8 +1286,9 @@ const TYPE = module.exports.TYPE = {
         } else {
           time = ((time.getHours() * 60 + time.getMinutes()) * 60 + time.getSeconds()) * 1000 + time.getMilliseconds();
         }
-        time = Math.round(time * Math.pow(10, parameter.scale - 3));
-        time += Math.round((parameter.value.nanosecondDelta != null ? parameter.value.nanosecondDelta : 0) * Math.pow(10, parameter.scale));
+        time = time * Math.pow(10, parameter.scale - 3);
+        time += (parameter.value.nanosecondDelta != null ? parameter.value.nanosecondDelta : 0) * Math.pow(10, parameter.scale);
+        time = Math.round(time);
         switch (parameter.scale) {
           case 0:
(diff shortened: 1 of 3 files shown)


src/data-type.js
```diff
@@ -1280,13 +1280,14 @@ const TYPE = module.exports.TYPE = {
...
-        let ref, time = new Date(+parameter.value);
+        let time = new Date(+parameter.value);
         if (options.useUTC) {
           time = ((time.getUTCHours() * 60 + time.getUTCMinutes()) * 60 + time.getUTCSeconds()) * 1000 + time.getUTCMilliseconds();
         } else {
           time = ((time.getHours() * 60 + time.getMinutes()) * 60 + time.getSeconds()) * 1000 + time.getMilliseconds();
         }
-        time = (time / 1000 + ((ref = parameter.value.nanosecondDelta) != null ? ref : 0)) * Math.pow(10, parameter.scale);
+        time = Math.round(time * Math.pow(10, parameter.scale - 3));
+        time += Math.round((parameter.value.nanosecondDelta != null ? parameter.value.nanosecondDelta : 0) * Math.pow(10, parameter.scale));
```
tvrprasad: Please do the <code>Math.round()</code> after the addition. Like below:
```javascript
time = time * Math.pow(10, parameter.scale - 3);
time += (parameter.value.nanosecondDelta != null ? parameter.value.nanosecondDelta : 0) * Math.pow(10, parameter.scale);
time = Math.round(time);
```
Also, can you please add tests?

Thanks for fixing this!

peter-int-veld:
Added a unit test for the fix.

ErikJansenIRefact:
When will this fix be made available?

ErikJansenIRefact:
Hi. 

We currently have another module: tedious-connectionpool which depends on tedious. We cannot deploy to production due to the issue with the nanoseconds. 

What should we do to get a review on this fix?
Thanks in advance for your suggestions.

PR #606, opened 2017-07-31
If encrypt not specified, escalate if required by default

ciaranj:
Changes the behaviour of the pre-login to automatically try again if the server requires
encryption to be enabled but the user hasn't configured the client in that way.

Still allows for the user to explicitly state that they do not ever want to use
encryption by specifying the value as false.  Or to always use and require encryption by setting the value as true.

This commit partially undoes the work done in ceaa8dc15994fe600b01c5e0d84949f3be0984a0 and is a  change in the behaviour, but brings this driver more inline with other TDS drivers such as jTDS.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -69,7 +69,7 @@ class Connection extends EventEmitter {
         },
         enableArithAbort: false,
         enableAnsiNullDefault: true,
-        encrypt: false,
+        encrypt: undefined,
         fallbackToDefaultDb: false,
         instanceName: undefined,
         isolationLevel: ISOLATION_LEVEL.READ_COMMITTED,
@@ -715,7 +715,9 @@ class Connection extends EventEmitter {
   socketClose() {
     this.debug.log('connection to ' + this.config.server + ':' + this.config.options.port + ' closed');
     if (this.state === this.STATE.REROUTING) {
-      this.debug.log('Rerouting to ' + this.routingData.server + ':' + this.routingData.port);
+      if (typeof this.routingData !== 'undefined') {
+        this.debug.log('Rerouting to ' + this.routingData.server + ':' + this.routingData.port);
+      }
       return this.dispatchEvent('reconnect');
     } else if (this.state === this.STATE.TRANSIENT_FAILURE_RETRY) {
       const server = this.routingData ? this.routingData.server : this.server;
@@ -753,12 +755,15 @@ class Connection extends EventEmitter {
     });
 
     if (preloginPayload.encryptionString === 'ON' || preloginPayload.encryptionString === 'REQ') {
-      if (!this.config.options.encrypt) {
+      if (this.config.options.encrypt === false) {
         this.emit('connect', ConnectionError("Server requires encryption, set 'encrypt' config option to true.", 'EENCRYPT'));
         return this.close();
+      } else if (typeof this.config.options.encrypt === 'undefined') {
+        this.config.options.encrypt = true;
+        return this.transitionTo(this.STATE.REROUTING);
+      } else {
+        return this.dispatchEvent('tls');
       }
-
-      return this.dispatchEvent('tls');
     } else {
       return this.dispatchEvent('noTls');
     }
--- a/examples/minimal.js
+++ b/examples/minimal.js
@@ -14,8 +14,7 @@ var config = {
       token: false,
       log: true
     },
-    database: 'DBName',
-    encrypt: true // for Azure users
+    database: 'DBName'
   }
   */
 };


ciaranj:
That thing where you've not worked on a project for so long you discover you're no longer a maintainer ;) doh!

v-suhame:
@ciaranj Did you get to test if retry works? Just noticed that driver sends [`ENCRYPT_NOT_SUP`](https://github.com/tediousjs/tedious/blob/master/src/prelogin-payload.js#L107) instead of `ENCRYPT_OFF`, if encrypt is not enabled. 
As per TDS doc (in 2.26.5 Prelogin section), for client with `ENCRYPT_NOT_SUP` and server with `ENCRYPT_ON` combination, server response is `ENCRYPT_REQ (connection terminated)`. Curious if this impacts retry.
It would be good if driver could send `ENCRYPT_OFF` in first attempt and then `ENCRYPT_ON` on receiving `ENCRYPT_REQ`, it would make it inline with TDS spec.

ciaranj:
Retry does work definitely.  However the server does have errors in the event view, I wonder if that's related :)

v-suhame:
Most likely they are related :) Can you try sending `ENCRYPT_OFF` in prelogin? Hopefully that will clear the error. I don't foresee any side effects if we make this switch from `ENCRYPT_NOT_SUP`.

src/connection.js
```diff
@@ -715,7 +715,9 @@ class Connection extends EventEmitter {
   socketClose() {
     this.debug.log('connection to ' + this.config.server + ':' + this.config.options.port + ' closed');
     if (this.state === this.STATE.REROUTING) {
-      this.debug.log('Rerouting to ' + this.routingData.server + ':' + this.routingData.port);
+      if (typeof this.routingData !== 'undefined') {
```
arthurschreiber: This change here seems unrelated? 🤔 If it's not related to the other changes, could you revert this?

src/connection.js
```diff
@@ -69,7 +69,7 @@ class Connection extends EventEmitter {
         },
         enableArithAbort: false,
         enableAnsiNullDefault: true,
-        encrypt: false,
+        encrypt: undefined,
```
arthurschreiber: 👍 for sane defaults!

src/connection.js
arthurschreiber: Ah, no, I see that you're re-using the re-routing logic here. I'm not sure how I feel about this, but if that's working, why not? 👍

src/connection.js
```diff
@@ -753,12 +755,15 @@ class Connection extends EventEmitter {
     });
 
     if (preloginPayload.encryptionString === 'ON' || preloginPayload.encryptionString === 'REQ') {
-      if (!this.config.options.encrypt) {
+      if (this.config.options.encrypt === false) {
         this.emit('connect', ConnectionError("Server requires encryption, set 'encrypt' config option to true.", 'EENCRYPT'));
         return this.close();
+      } else if (typeof this.config.options.encrypt === 'undefined') {
```
arthurschreiber: What do you think about changing the conditions here to be a bit more explicit?

```js
if (this.config.options.encrypt === true) {
  ...
} else if (this.config.options.encrypt === false) {
  ...
} else {
  ...
}
```

arthurschreiber:
Also, as @v-suhame already suggested, can you change the prelogin payload to send `ENCRYPT_OFF`?

src/connection.js
ciaranj: Yeah I had the same feeling if I'm honest, but I traded that against the scale of change I'd need to do to encapsulate a 'reconnect' state change :/

src/connection.js
ciaranj: I don't mind tbh, from what I recall (this was a while ago!) Semantically I guess it would still behave the same?

ciaranj:
Yeah, it's on my list :/ ( to change the payload)

arthurschreiber:
Do you think just changing the default for the `encrypt` option to `true` would be a simpler and more straightforward fix for this? I don't really see any negative effect of doing this, and I even believe we could do this without bumping the major version. 🤔 

@v-suhame What do you think about this ☝️ proposal? It'd solve the same issue as the code in this pull request, but without adding more code to Tedious.

ciaranj:
Sure thing, it was only done this way so as to not negatively impact any existing consumers of the driver (but you’re right we could rely on semver for that)

v-suhame:
I was on same page as @ciaranj regarding backward compatibility. After some reading, figured its version older than SQL Server 2000 that doesn't support SSL, that hopefully has a very small customer base(or none at all).
It's a thumbs-up for making encrypt true by default.

arthurschreiber:
This will be solved by #694, as we'll default to an encrypted connection instead. #694 will be part of the next semver-major release of `tedious`.

Issue #379, opened 2016-04-19
 request.on('done',  ) Not working


hasan3ysf:
Hi,
I wrote the below statement to be executed once the SQlexec is DONe, but got nothing!

`request.on('done', function (rowCount, more, rows) {console.log(rowCount + ' rows final'); });`

my configuration is defined as:

```
    var config = { 
       userName: 'sa',
       password: 'myPswd',
       server: 'myIP',
       options: {
            database: 'myDB',
                rowCollectionOnRequestCompletion: true,
                  rowCollectionOnDone: true
         },
    };
```

seanpoulter:
I can recreate this problem running the [minimal example](https://github.com/pekim/tedious/blob/master/examples/minimal.js#L58) connecting to SQL Server 2012.

codyaweber:
I'm fairly new to tedious, so forgive me if this is way off. But I think this (from [docs](http://pekim.github.io/tedious/api-connection.html#function_execSql)) could be causing a problem:

"As `sp_executesql` is used to execute the SQL, if the same SQL is executed multiples times using this function, the SQL Server query optimizer is likely to reuse the execution plan it generates for the first execution."

I think 'reusing the execution plan' means the query gets treated as a stored procedure, so instead of the `done` event being called upon completion, `doneProc` and `doneInProc` get called. Using `execSqlBatch` as opposed to `execSql` would fix this. Not sure if there's a more appropriate way to prevent the server from treating the query as a stored procedure though.

RaduGrama:
@codyaweber, you are correct and all workarounds you suggested address the issue, with `doneProc` or `doneInProc` probably being the better workarounds. This should probably be considered a documentation bug.

arthurschreiber:
Yes, this is a documentation bug. The `done` event is not always fired. There's been quite a few issues regarding this. Patches for fixing the documentation would be very welcome.

veleek:
Excuse me if this seems like a ridiculous question, but where IS the documentation located?  I was going to try and update this, but I don't know where the update should go.

codyaweber:
It seems the link I posted earlier has changed.

http://tediousjs.github.io/tedious/api-request.html

PR #607, opened 2017-08-18
Add clarifying details around done events.

veleek:
See #379.  Attempting to make the documentation a bit more clear.

I may do a quick run through all the documentation later and make sure that all <code> elements also link to their respective items in the documentation.


v-suhame:
@veleek Thanks for updating the documentation !

PR #612, opened 2017-09-14
Kerberos Integrated auth

v-suhame:
- Server name should be FQDN - security context is generated based on this.
- Tested with keytab and cached Kerberos ticket.
- Uses https://www.npmjs.com/package/node-kerberos/ as dependency. This is based on https://github.com/mongodb-js/kerberos , couldn't get the necessary changes for Kerberos merged in there due to inactivity.
 
Tests to be added soon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "readable-stream": "^2.2.6",
     "sprintf": "0.1.5",
     "sspi-client": "^0.1.0",
-    "kerberos": "git://github.com/v-suhame/kerberos.git#krbv5Mech"
+    "node-kerberos": "^0.0.25"
   },
   "devDependencies": {
     "async": "^1.4.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "iconv-lite": "^0.4.11",
     "readable-stream": "^2.2.6",
     "sprintf": "0.1.5",
-    "sspi-client": "^0.1.0"
+    "sspi-client": "^0.1.0",
+    "kerberos": "git://github.com/v-suhame/kerberos.git#krbv5Mech"
   },
   "devDependencies": {
     "async": "^1.4.0",
--- a/src/connection.js
+++ b/src/connection.js
@@ -322,23 +322,34 @@ class Connection extends EventEmitter {
 
   cleanupConnection(cleanupTypeEnum) {
     if (!this.closed) {
-      this.clearConnectTimer();
-      this.clearRequestTimer();
-      this.clearRetryTimer();
-      this.closeConnection();
-      if (cleanupTypeEnum === this.cleanupTypeEnum.REDIRECT) {
-        this.emit('rerouting');
-      } else if (cleanupTypeEnum !== this.cleanupTypeEnum.RETRY) {
-        this.emit('end');
+      const cleanConnection = () => {
+          this.clearConnectTimer();
+          this.clearRequestTimer();
+          this.clearRetryTimer();
+          this.closeConnection();
+
+          if (cleanupTypeEnum === this.cleanupTypeEnum.REDIRECT) {
+          this.emit('rerouting');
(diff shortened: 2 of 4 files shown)


v-suhame:
@tvrprasad @arthurschreiber Can you review it when possible?

arthurschreiber:
Heya @v-suhame! 👋

I just opened #624, which proposes a better API for adding authentication providers (like the Kerberos provider you are working on here). Please check it out and let me know what you think! ❤️

package.json
```diff
@@ -46,7 +46,8 @@
     "iconv-lite": "^0.4.11",
     "readable-stream": "^2.2.6",
     "sprintf": "0.1.5",
-    "sspi-client": "^0.1.0"
+    "sspi-client": "^0.1.0",
+    "node-kerberos": "^0.0.25"
```
arthurschreiber: As mentioned in #624, I don't think having all these different libraries be dependencies of `tedious` is a scalable approach.

src/connection.js
```diff
@@ -409,23 +414,34 @@ class Connection extends EventEmitter {
...
+          this.request = undefined;
+        }
+        this.closed = true;
+        this.loggedIn = false;
+        return this.loginError = null;
+      };
+
+      // clean kerberos security context if exists
+      if (this.kerberos && this.context) {
+        this.kerberos.authGSSClientClean(this.context, () => { cleanConnection(); });
```
arthurschreiber: What do you think about cleaning the kerberos context right after the security handshake was performed successfully? 🤔

src/connection.js
```diff
@@ -1533,6 +1616,39 @@ Connection.prototype.STATE = {
       }
     }
   },
+  SENT_LOGIN7_WITH_KERBEROS: {
```
arthurschreiber: The state machine is starting to get really complicated. Do you think kerberos authentication could be implemented without adding a new state? I know that there is also a special state for NTLM authentication defined, but while working on #624, I was thinking if we maybe could get rid of authentication specific states somehow. 🤔

src/connection.js
```diff
@@ -1057,6 +1101,41 @@ class Connection extends EventEmitter {
     }
   }
 
+  processLogin7KerberosResponse() {
+    if (this.loggedIn) {
+      return this.dispatchEvent('loggedIn');
```
arthurschreiber: Maybe I'm missing something, but is this really the right location for this call? Where is `this.loggedIn` set to true for kerberos authentication? 🤔

arthurschreiber: Ok, I think I figured out where this happens. 😓

src/connection.js
```diff
@@ -1057,6 +1101,41 @@ class Connection extends EventEmitter {
...
+          //verify if kerberos auth was successful, ie, GSS_C_COMPLETE flag returned
+          if (!((null != result) && ('number' === typeof (result)) && (1 === result /* GSS_C_COMPLETE */))) {
+            this.emit('error', new Error('Expected GSS_C_COMPLETE flag not received, kerberos authentication failed'));
+            return this.close();
+          }
+
+          this.sspiClientResponsePending = false;
+
+          // clear the ntlmpacket
+          delete this.ntlmpacketBuffer;
```
arthurschreiber: Deleting properties has negative performance implications in V8, if I remember correctly. 🤔

v-suhame:
@arthurschreiber Thanks for reviewing it :) I'll keep this PR open for now, will make the necessary changes after pluggable-auth-providers gets merged.

src/connection.js
v-suhame: sure, I'll double check that.

Fadoli:
Hey just asking, not work is planned on this anymore ? (since the pluggable auth providers is okay ish ?)

David-Engel:
@Fadoli Changes from this PR look to be incorporated in #624 which is still open.

Issue #608, opened 2017-08-24
sql.DateTime data type does not cast as per sql.DateTime2


agarstang:
Copying [#185](https://github.com/patriksimek/node-mssql/issues/185) from [node-mssql](https://github.com/patriksimek/node-mssql).

```
days = Math.floor((parameter.value.getTime() - UTC_EPOCH_DATE.getTim
                                    ^
TypeError: undefined is not a function
```
In data-type.js you have missed to convert date string to object as you have done for DateTime2
```
if (parameter.value != null) {
      parameter.value = new Date(+parameter.value);  // this line is missing
       if (options.useUTC) {
```

arthurschreiber:
The way we do type conversion in `tedious` is a mess, and really inconsistent.

I _think_ this here is an actual bug (because it's raising a `TypeError`), and that's bad and needs to be fixed. But I'm not sure what a "good" fix is going to look like.

Long term (in `tedious@3.x` or even later), I want us to move away from doing any type conversion at all - if your application defines a parameter to be a `DateTime`, it should be your application's responsibility to pass a `Date` object and not a `String` or something else. And if you want to make use of SQLServer's type conversion, you should've specified a `String` parameter anyway.

chdh:
@arthurschreiber 
>  I want us to move away from doing any type conversion at all

This answers my [question]( https://github.com/tediousjs/tedious/pull/623#issuecomment-336662934) in #623.
What do you think about conversion from float to integer (same data type in JS, but not for SQL Server)? I think it's better not to convert at all.

> Long term (in tedious@3.x or even later),

I think it's not such a big change and it's important. I could provide a PR if you like.

arthurschreiber:
> I think it's not such a big change and it's important. I could write a PR if you like.

It's a backwards incompatible change if we stop converting these values, so it should be a major version bump. We should not be afraid of major version bumps. 😄

PR #623, opened 2017-10-11
TypeError: parameter.value.getTime is not a function

atsage:
Fixed an error TypeError: parameter.value.getTime is not a function.  Parameter needs to be converted to a date to avoid errors.

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -295,7 +295,7 @@ const TYPE = module.exports.TYPE = {
     writeParameterData: function(buffer, parameter, options) {
       if (parameter.value != null) {
         let days, dstDiff, milliseconds, seconds, threeHundredthsOfSecond;
-		var time = new Date(+parameter.value);
+        var time = new Date(+parameter.value);
         if (options.useUTC) {
           days = Math.floor((time.getTime() - UTC_EPOCH_DATE.getTime()) / (1000 * 60 * 60 * 24));
           seconds = time.getUTCHours() * 60 * 60;
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1,15 +1,17 @@
-const guidParser = require('./guid-parser');
+'use strict';
 
-const NULL = (1 << 16) - 1;
-const EPOCH_DATE = new Date(1900, 0, 1);
-const UTC_EPOCH_DATE = new Date(Date.UTC(1900, 0, 1));
-const YEAR_ONE = new Date(2000, 0, -730118);
-const UTC_YEAR_ONE = Date.UTC(2000, 0, -730118);
-const MAX = (1 << 16) - 1;
+var guidParser = require('./guid-parser');
 
-const typeByName = module.exports.typeByName = {};
+var NULL = (1 << 16) - 1;
+var EPOCH_DATE = new Date(1900, 0, 1);
+var UTC_EPOCH_DATE = new Date(Date.UTC(1900, 0, 1));
+var YEAR_ONE = new Date(2000, 0, -730118);
+var UTC_YEAR_ONE = Date.UTC(2000, 0, -730118);
+var MAX = (1 << 16) - 1;
 
-const TYPE = module.exports.TYPE = {
+var typeByName = module.exports.typeByName = {};
+
+var TYPE = module.exports.TYPE = {
   0x1F: {
     type: 'NULL',
     name: 'Null'
@@ -19,16 +21,16 @@ const TYPE = module.exports.TYPE = {
     type: 'INT1',
     name: 'TinyInt',
 
-    declaration: function() {
+    declaration: function declaration() {
       return 'tinyint';
     },
 
-    writeTypeInfo: function(buffer) {
+    writeTypeInfo: function writeTypeInfo(buffer) {
       buffer.writeUInt8(typeByName.IntN.id);
       return buffer.writeUInt8(1);
     },
 
-    writeParameterData: function(buffer, parameter) {
+    writeParameterData: function writeParameterData(buffer, parameter) {
       if (parameter.value != null) {
         buffer.writeUInt8(1);
         return buffer.writeUInt8(parseInt(parameter.value));
@@ -37,7 +39,7 @@ const TYPE = module.exports.TYPE = {
       }
     },
 
-    validate: function(value) {
+    validate: function validate(value) {
       if (value == null) {
         return null;
       }
@@ -56,16 +58,16 @@ const TYPE = module.exports.TYPE = {
     type: 'BIT',
     name: 'Bit',
 
-    declaration: function() {
+    declaration: function declaration() {
       return 'bit';
     },
 
-    writeTypeInfo: function(buffer) {
+    writeTypeInfo: function writeTypeInfo(buffer) {
       buffer.writeUInt8(typeByName.BitN.id);
       return buffer.writeUInt8(1);
     },
 
-    writeParameterData: function(buffer, parameter) {
+    writeParameterData: function writeParameterData(buffer, parameter) {
       if (typeof parameter.value === 'undefined' || parameter.value === null) {
         return buffer.writeUInt8(0);
       } else {
@@ -74,7 +76,7 @@ const TYPE = module.exports.TYPE = {
       }
     },
 
-    validate: function(value) {
+    validate: function validate(value) {
       if (value == null) {
         return null;
       }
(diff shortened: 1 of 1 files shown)


atsage:
There are a few issues fixed with this PR:

https://github.com/patriksimek/node-mssql/issues/377
https://github.com/patriksimek/node-mssql/issues/185
https://github.com/tediousjs/tedious/issues/608

I've created a node that depends on node-mssql and so my node also has these datetime errors.

I also notice that the node on npm is not the same code that's in this repository.  How often do you publish to npm?

src/data-type.js
```diff
@@ -295,19 +295,20 @@ const TYPE = module.exports.TYPE = {
     writeParameterData: function(buffer, parameter, options) {
       if (parameter.value != null) {
         let days, dstDiff, milliseconds, seconds, threeHundredthsOfSecond;
+        var time = new Date(+parameter.value);
```
chdh: What is the rationale behind this conversion?
If `parameter.value` is already a `Date`, the `+` operator converts it to an integer and the `new Date()` converts it back to a `Date`. 
If `parameter.value` is a string that does not contain a number, the `+` operator produces a NaN.
e.g.: `+'2017-10-11'` ==> `NaN`
or: `+'Wed Oct 11 2017 19:49:46 GMT+0200'` ==> `NaN`

chdh:
The root problem of this issue is that sometimes the parameter values are passed through `Type.validate()` (in `Request.validateParameters()`) and converted, but sometimes they are not passed through `Type.validate()` (in `BulkLoad.addRow()`).

A validation function like `Request.validateParameters()` should not have the side effect of converting the parameter values and `Type.validate()` should not return the converted value.

Parameter value conversion should be moved to a separate function which is then called from both `Type.validate()` and `Type.writeParameterData()`.

Another solution would be to rename `validate()` to e.g. `validateAndConvert()` and also call it in BulkLoad.

atsage:
What file are you talking about?

atsage:
So is this something that will be changed?

chdh:
I think the whole conversion logic in the data-type.js module has to be revised.

Conversion to integer is also faulty. `parseInt()` is used even if the value is already a number and that can produce strange effects, e.g.:

`console.log(parseInt(0.000000006));`   ==> Output is 6!

Maybe it would be better not to convert at all and just throw a `TypeError` in `writeParameterData()` when the data type does not match? Why is it important to automatically convert a `string` to a `Date`?

atsage:
I don't think that it's important to convert, the values should be passed in the correct format to begin with.  However, I've seen this issue a couple places and that is the fix.  It was initially posted here https://github.com/patriksimek/node-mssql/issues/185 and line 1413 of data-type.js has the same code to convert DATETIME2 time = new Date(+parameter.value);  I'm not a professional programmer, so if there is something that needs to be addressed in general with data-types.js then I would be glad to see it put in place.

atsage:
After some more testing I have found that the problem is the node-mssql node converts the datetime into a number.  See here starting on line 166 https://github.com/patriksimek/node-mssql/blob/master/lib/datatypes.js  I will raise an issue there as well as this seems incorrect.

atsage:
https://github.com/patriksimek/node-mssql/issues/546 Issue raised here

chdh:
I don't agree. Starting at line 166 in https://github.com/patriksimek/node-mssql/blob/master/lib/datatypes.js, `Date` values are not converted to `number` but to a string in the format `N'yyyy-mm-dd hh:mm:ss'`.

atsage:
You are right, I wasn't looking at it very hard.  It must get converted somewhere else to a number because when it gets to the section of this file it is indeed a number type.  Either way, neither of those types are a date and in node-mssql there is no possible way to send a date type so it fails.

chdh:
@arthurschreiber @v-suhame what do you think about automatic value conversion for SQL statement parameters and BulkLoad field values? Should Tedious try to convert the data type of values that do not match, or just throw a TypeError?

Examples:

- Convert a string to an integer.
- Round a float value to an integer.
- Convert a string to a `Date`.

The current implementation converts values for SQL statement parameters but not for BulkLoad field values. There are some errors and discrepancies in `data-type.js` that should be fixed. How can we proceed?

arthurschreiber:
> Maybe it would be better not to convert at all and just throw a TypeError in writeParameterData() when the data type does not match? Why is it important to automatically convert a string to a Date?

Yeah, so I totally think we should not be doing any conversion between datatypes. 👍 

> I don't think that it's important to convert, the values should be passed in the correct format to begin with.

Yup, again, this is the way to go. 😅 

---

I think long-term, here's what we need to do (and what probably is the only sane approach to this):

`tedious` should provide a system to plug in custom data-type support modules. Each module would handle parsing of TDS data values into JavaScript objects and converting JavaScript objects back into TDS data values.

Out-of-the box, `tedious` should only support TDS data types that can be converted to equivalent JavaScript data types without a loss of precision or other information. E.g. we would support "basic" data types like `varchar`, `char`, `binary`, `varbinary`, `int`, etc.

If a user of `tedious` needs to work with other data types that can't be represented in plain JavaScript (like `bigint`), they'd have to provide a data-type module. These could be either write by them (to satisfy their needs) or written by us (to provide a base set).

Let's use the `bigint` example again: There's currently a bunch of arbitrary precision integer libraries available for JavaScript (https://www.npmjs.com/browse/keyword/biginteger). Whichever one we pick, someone will be unhappy because we didn't use the library they use in their application. By having a pluggable system, we can provide a base set of types for "popular" libraries, and if they need something special, they can go and build their own.

This would also solve the issue of providing conversion logic - if someone believes they really need to be able to convert String objects to DateTime values, they should be able to hook up their own data-type handler and do whatever they want. 😄 

@chdh What do you think?

chdh:
@arthurschreiber A plug-in architecture would be a very good and flexible solution. But in the short term, we should fix the problems with the "basic" data types. This would be a small change.

arthurschreiber:
@chdh You're right, we need a fix in the interim and can't leave this broken. As you've worked on the BulkLoad code before, what would be you preferred way to fix this?

chdh:
@arthurschreiber I would just clean up the existing `validate()` and `writeParameterData()` functions for each data type in data-type.js and add `validate()` calls to `BulkLoad`.

If we want backward compatibility, so that the PR can already be included in a Tedious 2.x release, I would add `convert()` functions for each data type and call these in `Request.validateParameters()` just after the `validate()` calls. In this case, `validateParameters()` should be renamed to `validateAndConvertParameters()`. We could add an option to enable/disable parameter conversion. For Tedious 2.x, the option would be enabled by default. For 3.x it would be disabled.

niwsa:
Any update on the fix? I am facing the same issue.

arthurschreiber:
Hey there,

I've been spending a lot of time going through the internals of tedious and it's data type system. There's a lot of weird code in there (as pointed out by @chdh).

I'll be opening a PR shortly that will fix this issue here by throwing a more descriptive error message when an unexpected type is passed to the different `writeParameterData` functions for different data types. As that is not a breaking change and will clearly signal to our users why they're seeing an error (which is much better than just throwing some obscure javascript error).

We'll have to find a better method of handling types long term, but that's a separate discussion from the error and fix proposed here. 😞 

Thank you for opening this PR and getting the discussion started. ❤️  I'm really sorry this has taken a long time to start moving to a solution 😞.

I really appreciate the time you spent and hope the solution I'm suggesting above is satisfactory. If not, please let us know!

atsage:
Any progress is good, I'll take it!

Shaffuan:
Any Progress Guy???

esterkjuicepress:
Was this fix ever merged? I still have issues with bulk inserting DateTime types
```
TypeError: parameter.value.getTime is not a function
at Object.writeParameterData (\node_modules\tedious\lib\data-types\datetime.js:30:44)
```

stegel:
was this ever resolved?

thbl:
+1

MichaelSun90:
@arthurschreiber, I had a different PR #907 that holds a similar change, and it already updated to the latest master.  I am not sure whether that we should close this PR and get the other PR merged in?

arthurschreiber:
I think with the changes in https://github.com/tediousjs/tedious/pull/1065 and `validateBulkLoadParameters` set to `true` this is no longer necessary? Please feel free to reopen if you think I'm mistaken. 🙇‍♂️

PR #624, opened 2017-10-13
feat: Introduce pluggable authentication providers.

arthurschreiber:
In preparation of a new `tedious` release, I was going through all changes since `v2.0.1`. One thing I want to cleanup before the release is the integration with `sspi-client`.

I'm not super happy about the current situation where `sspi-client` is a hard dependency of `tedious`, and seeing that SQL Server actually supports more authentication methods and that there is another authentication related pull request in flight (#612) that adds `node-kerberos` as another hard dependency, I feel that the current approach for authentication methods is not scalable.

In this pull request, I explore the addition of what I call "authentication" providers (they're called "Security Support Providers" in the MS-TDS specification). The whole pull request is still very much in flux and a work in progress, but I believe the direction proposed here to be sound.

Authentication providers are simply objects that implement a `.handshake(input, callback)` method that is responsible for taking in authentication handshake data coming from SQL Server, and calling the passed `callback` with the handshake data that should be sent back to the SQL Server.

The long-term goal is to ship tedious with only a "default" dummy provider that does not actually do anything. Other providers like for NTLM, Kerberos or the native SSPI support will be extracted into separate npm modules organization. This way, we can extract both the implementation of these providers out of `tedious`, and keep `tedious` free from dependencies that are not required for all users.

The authentication provider to use for connecting to SQL Server will be specifiable via a new option when establishing a connection, and each provider will be able to handle options that are specific to it in it's own way.

For example, when using NTLM (domain) authentication I imagine connection creation could look like this:

```js
new Connection({
  "server": "localhost",
  "userName": "sa",
  "password": "yourStrong(!)Password",
  "authProvider": require("tedious-auth-ntlm")({
    "domain": "MYDOMAIN.COM",
    "workstation": "FOOBARBAZ"
  }),
  "options": {
    "port": 1433,
    "database": "master"
  }
});
```

While native SSPI could look like this:

```js
new Connection({
  "server": "localhost",
  "userName": "sa",
  "password": "yourStrong(!)Password",
  "authProvider": require("tedious-auth-native")({
    "securityProvider": "negotiate"
  }),
  "options": {
    "port": 1433,
    "database": "master"
  }
});
```

@tvrprasad @v-suhame What do you think?

### DIFF
--- a/src/auth/ntlm.js
+++ b/src/auth/ntlm.js
@@ -1,0 +1,271 @@
+const WritableTrackingBuffer = require('./tracking-buffer/writable-tracking-buffer');
+const crypto = require('crypto');
+const BigInteger = require('big-number').n;
+
+const hex = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'];
+
+const NTLMFlags = {
+  NTLM_NegotiateUnicode: 0x00000001,
+  NTLM_NegotiateOEM: 0x00000002,
+  NTLM_RequestTarget: 0x00000004,
+  NTLM_Unknown9: 0x00000008,
+  NTLM_NegotiateSign: 0x00000010,
+  NTLM_NegotiateSeal: 0x00000020,
+  NTLM_NegotiateDatagram: 0x00000040,
+  NTLM_NegotiateLanManagerKey: 0x00000080,
+  NTLM_Unknown8: 0x00000100,
+  NTLM_NegotiateNTLM: 0x00000200,
+  NTLM_NegotiateNTOnly: 0x00000400,
+  NTLM_Anonymous: 0x00000800,
+  NTLM_NegotiateOemDomainSupplied: 0x00001000,
+  NTLM_NegotiateOemWorkstationSupplied: 0x00002000,
+  NTLM_Unknown6: 0x00004000,
+  NTLM_NegotiateAlwaysSign: 0x00008000,
+  NTLM_TargetTypeDomain: 0x00010000,
+  NTLM_TargetTypeServer: 0x00020000,
+  NTLM_TargetTypeShare: 0x00040000,
+  NTLM_NegotiateExtendedSecurity: 0x00080000,
+  NTLM_NegotiateIdentify: 0x00100000,
+  NTLM_Unknown5: 0x00200000,
+  NTLM_RequestNonNTSessionKey: 0x00400000,
+  NTLM_NegotiateTargetInfo: 0x00800000,
+  NTLM_Unknown4: 0x01000000,
+  NTLM_NegotiateVersion: 0x02000000,
+  NTLM_Unknown3: 0x04000000,
+  NTLM_Unknown2: 0x08000000,
+  NTLM_Unknown1: 0x10000000,
+  NTLM_Negotiate128: 0x20000000,
+  NTLM_NegotiateKeyExchange: 0x40000000,
+  NTLM_Negotiate56: 0x80000000
+};
+
+class NTLMResponsePayload {
+  constructor(loginData) {
+    this.data = this.createResponse(loginData);
+  }
+
+  toString(indent) {
+    indent || (indent = '');
+    return indent + 'NTLM Auth';
+  }
+
+  createResponse(challenge) {
+    const client_nonce = this.createClientNonce();
+    const lmv2len = 24;
+    const ntlmv2len = 16;
+    const domain = challenge.domain;
+    const username = challenge.userName;
+    const password = challenge.password;
+    let ntlmData = challenge.ntlmpacket;
+    const server_data = ntlmData.target;
+    const server_nonce = ntlmData.nonce;
+    const bufferLength = 64 + (domain.length * 2) + (username.length * 2) + lmv2len + ntlmv2len + 8 + 8 + 8 + 4 + server_data.length + 4;
+    const data = new WritableTrackingBuffer(bufferLength);
+    data.position = 0;
+    data.writeString('NTLMSSP\u0000', 'utf8');
+    data.writeUInt32LE(0x03);
+    const baseIdx = 64;
+    const dnIdx = baseIdx;
+    const unIdx = dnIdx + domain.length * 2;
+    const l2Idx = unIdx + username.length * 2;
+    const ntIdx = l2Idx + lmv2len;
+    data.writeUInt16LE(lmv2len);
+    data.writeUInt16LE(lmv2len);
+    data.writeUInt32LE(l2Idx);
+    data.writeUInt16LE(ntlmv2len);
+    data.writeUInt16LE(ntlmv2len);
+    data.writeUInt32LE(ntIdx);
+    data.writeUInt16LE(domain.length * 2);
+    data.writeUInt16LE(domain.length * 2);
+    data.writeUInt32LE(dnIdx);
+    data.writeUInt16LE(username.length * 2);
+    data.writeUInt16LE(username.length * 2);
+    data.writeUInt32LE(unIdx);
+    data.writeUInt16LE(0);
+    data.writeUInt16LE(0);
+    data.writeUInt32LE(baseIdx);
+    data.writeUInt16LE(0);
+    data.writeUInt16LE(0);
+    data.writeUInt32LE(baseIdx);
+    data.writeUInt16LE(0x8201);
+    data.writeUInt16LE(0x08);
+    data.writeString(domain, 'ucs2');
+    data.writeString(username, 'ucs2');
+    const lmv2Data = this.lmv2Response(domain, username, password, server_nonce, client_nonce);
+    data.copyFrom(lmv2Data);
+    const genTime = new Date().getTime();
+    ntlmData = this.ntlmv2Response(domain, username, password, server_nonce, server_data, client_nonce, genTime);
+    data.copyFrom(ntlmData);
+    data.writeUInt32LE(0x0101);
+    data.writeUInt32LE(0x0000);
+    const timestamp = this.createTimestamp(genTime);
+    data.copyFrom(timestamp);
+    data.copyFrom(client_nonce);
+    data.writeUInt32LE(0x0000);
+    data.copyFrom(server_data);
+    data.writeUInt32LE(0x0000);
+    return data.data;
+  }
+
+  createClientNonce() {
+    const client_nonce = new Buffer(8);
+    let nidx = 0;
+    while (nidx < 8) {
+      client_nonce.writeUInt8(Math.ceil(Math.random() * 255), nidx);
+      nidx++;
+    }
+    return client_nonce;
+  }
+
+  ntlmv2Response(domain, user, password, serverNonce, targetInfo, clientNonce, mytime) {
+    const timestamp = this.createTimestamp(mytime);
+    const hash = this.ntv2Hash(domain, user, password);
+    const dataLength = 40 + targetInfo.length;
+    const data = new Buffer(dataLength);
+    serverNonce.copy(data, 0, 0, 8);
+    data.writeUInt32LE(0x101, 8);
+    data.writeUInt32LE(0x0, 12);
+    timestamp.copy(data, 16, 0, 8);
+    clientNonce.copy(data, 24, 0, 8);
+    data.writeUInt32LE(0x0, 32);
+    targetInfo.copy(data, 36, 0, targetInfo.length);
+    data.writeUInt32LE(0x0, 36 + targetInfo.length);
+    return this.hmacMD5(data, hash);
+  }
+
+  createTimestamp(time) {
+    const tenthsOfAMicrosecond = new BigInteger(time).plus(11644473600).multiply(10000000);
+    const hexArray = [];
+
+    let pair = [];
+    while (tenthsOfAMicrosecond.val() !== '0') {
+      const idx = tenthsOfAMicrosecond.mod(16);
+      pair.unshift(hex[idx]);
+      if (pair.length === 2) {
+        hexArray.push(pair.join(''));
+        pair = [];
+      }
+    }
+
+    if (pair.length > 0) {
+      hexArray.push(pair[0] + '0');
+    }
+
+    return new Buffer(hexArray.join(''), 'hex');
+  }
+
+  lmv2Response(domain, user, password, serverNonce, clientNonce) {
+    const hash = this.ntv2Hash(domain, user, password);
+    const data = new Buffer(serverNonce.length + clientNonce.length);
+
+    serverNonce.copy(data);
+    clientNonce.copy(data, serverNonce.length, 0, clientNonce.length);
+
+    const newhash = this.hmacMD5(data, hash);
+    const response = new Buffer(newhash.length + clientNonce.length);
+
+    newhash.copy(response);
+    clientNonce.copy(response, newhash.length, 0, clientNonce.length);
+
+    return response;
+  }
+
+  ntv2Hash(domain, user, password) {
+    const hash = this.ntHash(password);
+    const identity = new Buffer(user.toUpperCase() + domain.toUpperCase(), 'ucs2');
+    return this.hmacMD5(identity, hash);
+  }
+
+  ntHash(text) {
+    const hash = new Buffer(21);
+    hash.fill(0);
+
+    const unicodeString = new Buffer(text, 'ucs2');
+    const md4 = crypto.createHash('md4').update(unicodeString).digest();
+    if (md4.copy) {
+      md4.copy(hash);
+    } else {
+      new Buffer(md4, 'ascii').copy(hash);
+    }
+    return hash;
+  }
+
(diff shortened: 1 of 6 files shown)


v-suhame:
@arthurschreiber sound like a good idea! Later we can also bring these `tedious` specific authentication providers dependencies under `tediousjs` organization. 
I'll start reviewing PR 👍

tvrprasad:
@arthurschreiber Couple of quick high level comments:
1) Further abstracting out Authentication sounds like general goodness.
2) How does this get rid of dependency on sspi-client? It seems to me like you still need to install it with Tedious. May be I'm missing something.

arthurschreiber:
@tvrprasad The idea would be to have at least 3 separate packages:

* `tedious-auth-native` to expose native system SSPI packages on Windows Systems.
* `tedious-auth-ntlm` to expose NTLM auth (regardless of operating system).
* `tedious-auth-kerberos` to expose Kerberos auth (regardless of operating system).

I think that'd cover all SSPI types SQL Server supports. I've seen that there is also some sort of Federated Authentication support documented in the latest TDS specifications, but I've not really looked into how it works - but it seems different enough from the SSPI authentication mechanisms that there's probably no sane way to reconcile the APIs.

`tedious` would not depend directly on any of these packages, so the dependencies like `node-kerberos` and `sspi-client` would not automatically be pulled in when installing `tedious` and using the standard TDS authentication mechanism.

arthurschreiber:
@v-suhame Thanks! I'd love to have tests for NTLM and Kerberos auth as part of the CI builds (first inside tedious itself, later as part of the individual packages). Is there some "straightforward" way to set up a CI system for this? Maybe via Azure SQL Server and Azure AD? 🤔 I do have an Azure account and could set everything up, but some pointers would be great.

(/cc @tvrprasad - I know you might not have time, but maybe you have a good idea here.)

v-suhame:
Federated Authentication was the former name for Azure AD, TDS docs still seems to use the old name. We have long term plan to add Azure AD support to Tedious 😉 
For CI, the only option I can think of is to set an Azure VM and test integrated auth and NTLM in there. 
@meet-bhagdev @tvrprasad Is there better alternatives?

src/connection.js
```diff
@@ -396,6 +399,8 @@ class Connection extends EventEmitter {
       REDIRECT: 1,
       RETRY: 2
     };
+
+    this.authProvider = new DefaultAuthProvider(this);
```
v-suhame: `this.authProvider` should be equal to the local `authProvider` here.

src/connection.js
```diff
@@ -51,12 +49,22 @@ class Connection extends EventEmitter {
       throw new TypeError('Invalid server: ' + config.server);
     }
 
+    let authProvider;
+    if (config.authProvider) {
+      authProvider = config.authProvider;
+    } else if (config.domain) {
+      authProvider = new NTLMAuthProvider(this, { domain: config.domain });
+    } else {
```
v-suhame: `NTLMAuthProvider` needs username, password, workstation along with domain.

meet-bhagdev:
@v-suhame that seems like the best idea right now. @Hadis-Fard any thoughts on this? How do we do this for PHP?

src/connection.js
```diff
@@ -867,8 +869,13 @@ class Connection extends EventEmitter {
...
-    const sendPayload = function(clientResponse) {
+  sendLogin7Packet(callback) {
+    this.authProvider.handshake(null, (error, data) => {
+      if (error) {
+        this.emit('error', error);
+        return this.close();
+      }
+
       const payload = new Login7Payload({
         domain: this.config.domain,
```
v-suhame: `domain` should be the value of `this.authProvider.options.domain`

src/connection.js
```diff
@@ -1462,74 +1344,29 @@ Connection.prototype.STATE = {
...
-          // call to complete on the client. We cannot process server data
-          // until this call completes as the state can change on completion of
-          // the call. Queue it for later.
-          const boundDispatchEvent = this.dispatchEvent.bind(this);
-          return setImmediate(boundDispatchEvent, 'data', data);
-        } else {
-          return this.sendDataToTokenStreamParser(data);
+        // Use SSPI blob if received
+        if (this.ntlmpacketBuffer) {
+          return this.authProvider.processChallenge(this.ntlmpacketBuffer, (error, responseBuffer) => {
```
v-suhame: Function call should be `this.authProvider.parseChallengeMessage( )`, authentication provider implementation uses the name `parseChallengeMessage`  😄

Hadis-Fard:
@meet-bhagdev +1 to @v-suhame's idea. As she said Tedious doesn't support AzureAD yet and is in our future plans to support it.
 In PHP, we are testing the Azure AD scenarios that can be tested using sql server docker (UID and PWD authentication), and added a variable to tests for other cases as we couldn't reveal credentials on GitHub , the variable is replaced with a valid credential when we pull tests from GitHub to internal test lab before connecting to sqlDB

tvrprasad:
> I'd love to have tests for NTLM and Kerberos auth as part of the CI builds (first inside tedious itself, later as part of the individual packages).

There are tests inside Tedious for NTLM, both for sspi-client and the non-sspi-client version. I think they run as part of CI but not 100% sure.

sspi-client is applicable only for Windows and that's where the tests run. sspi-client is not just NTLM, it's NTLM, Kerberos and Negotiate protocols. But since it uses Microsoft proprietary API, it'll only work on Windows.

Kerberos auth code path that @v-suhame has in PR, I think, should only be enabled for Linux.

v-suhame:
SQL Server image in docker only supports username/password auth, so we have to use Azure to test Integrated auth. Just realized that current CI setup, skips unit tests for NTLM and Windows auth 🤔 
I'll create an issue to track that.

And yes, Kerberos auth code path is reachable only for non-Windows based system.

meet-bhagdev:
@v-suhame how woulad Azure integrated auth work with the SQL Server docker image? I am not sure if that is a supported scenario. Am I missing something?

v-suhame:
Sorry! I meant SQL Server in Azure VM.

v-suhame:
@arthurschreiber I've committed changes, based on my review comments.

tvrprasad:
@v-suhame Could you install box version of SQL Server on Azure VM that's added to an AD domain to test for AD authentication, right? Or does that also only work with Azure AD?

tvrprasad:
> tedious would not depend directly on any of these packages, so the dependencies like node-kerberos and sspi-client would not automatically be pulled in when installing tedious and using the standard TDS authentication mechanism.

So, node-Kerberos and sspi-client would be external dependencies that will need to be installed independent of Tedious, may be globally. Is that the idea? Trying to understand when and how exactly these packages will be pulled in.

v-suhame:
I have pushed some more changes for Windows Integrated Auth.

@tvrprasad  `sspi-client` should be installed in client's application to use Windows Integrated Auth. 
```
new Connection({
  "server": "localhost",
  "domain": "local",
  "authProvider": require("sspi-client"),
  "options": {
    "port": 1433,
    "database": "test"
  }
});
```

v-suhame:
> Could you install box version of SQL Server on Azure VM that's added to an AD domain to test for AD authentication, right? Or does that also only work with Azure AD?

Yes, that setup should work without AAD.

src/connection.js
```diff
@@ -51,11 +50,24 @@ class Connection extends EventEmitter {
...
     }
 
+    let authProvider;
+    if (config.authProvider) {
+      if (config.authProvider.ModuleSupported) {
+        authProvider = new NativeAuthProvider(this);
+      }
+      //TODO: throw error if there is no authprovider to handle
+    } else if (config.domain) {
+      authProvider = new NTLMAuthProvider(this);
```
v-suhame: I've removed the `option` passed to authProviders. For each authentication mechanism, the options needed are different and huge.

arthurschreiber: @v-suhame I just added it back. I think it was my fault I didn't explain why I added this option in the first place. 😣

The idea behind provider specific options is to decouple the `Connection` class from the providers. Without a provider specific option set, each option that a provider could potentially support would need to be stored in the `Connection` class - and `Connection` is already doing and knowing too much.

If you've any concerns about this, please let me know! 🙇

arthurschreiber:
@v-suhame @chdh I think this is mostly ready.

I stepped back from my idea to extract the NTLM and native auth providers - I think for now we should keep them internal and only extract them in the next major `tedious` version, once we have a better understanding on whether a common interface can be shared between sspi auth and federated auth providers somehow.

arthurschreiber:
What's left here is:

- [ ] Extend the NTLM auth test cases to cover some missing cases.
- [ ] Extract https://github.com/tediousjs/tedious/pull/624/commits/168b845f12a283d0fdb13f5dfdc0690f29c5808a into a separate PR. This fix is unrelated to the auth provider changes, so I want this to be merged separately so.

I'll try to resolve those two open points today, and as soon as I have an approval for these changes in this pull request I'll get it merged and will release a new version of `tedious` to npm under the `next` tag. I'll open another issue to track feedback for `v2.1.0` and will promote it to `latest` after we gained a tiny bit of confidence.

How does this sound?

chdh:
@arthurschreiber Does this mean that the next Tedious release will have a dependency on the package `sspi-client` and require `node-gyp`?

src/connection.js
v-suhame: Got it! Sorry for making you re-work :pensive:

v-suhame:
I would prefer not having to require node-gyp, it brings in list of other dependencies, it can be challenging to get it going. 

Also, should revert commit [38ed20f ](https://github.com/tediousjs/tedious/pull/624/commits/38ed20fcedd58bc0a85dad09a967ca6481c23d39) - I haven't looked close into the details of sppi-client implementation, but without these checks we get `Single invocation of getNextBlob per instance of SspiClient may be in flight` error from sspi_client.

arthurschreiber:
> @arthurschreiber Does this mean that the next Tedious release will have a dependency on the package sspi-client and require node-gyp?

> I would prefer not having to require node-gyp, it brings in list of other dependencies, it can be challenging to get it going.

Thanks for the feedback! ❤️ Let's do the following then: we extract 2 new packages, `tedious-auth-ntlm` and `tedious-auth-native`. `tedious-auth-ntlm` will be included as a dependency of tedious (for backwards-compatibility), while `tedious-auth-native` will be available for everyone who wants to use tedious on windows and use native windows authentication.

In one of the next releases we will deprecate the `domain` configuration option, and then in the next major version, we will stop bundling `tedious-auth-ntlm` as part of tedious.

> Also, should revert commit 38ed20f - I haven't looked close into the details of sppi-client implementation, but without these checks we get Single invocation of getNextBlob per instance of SspiClient may be in flight error from sspi_client.

Thanks for letting me know. To be honest, I've not done a lot of testing on the native auth code yet - but if we extract this into a separate module anyway, we can make sure this works properly before releasing it (independent of any tedious release).

Thanks @v-suhame and @chdh!

src/auth/ntlm.js
```diff
@@ -0,0 +1,276 @@
...
+    } else {
+      process.nextTick(callback, null, this.buildAuthenticateMessage(this.parseChallengeMessage(data)));
+    }
+  }
+
+  buildNegotiateMessage() {
+    const messageBuffer = Buffer.alloc(40);
+    const domainBuffer = Buffer.from(this.options.domain || '', 'ascii');
+    const workstationBuffer = Buffer.from(this.options.workstation || '', 'ascii');
+
```
v-suhame: `config.workstation` is not re-assigned to `this.config.workstation` in connection.js (referring to master branch), was it missed somehow in the early PRs? or was that intentional? Can you give me some context?

arthurschreiber: The `workstation`  option was never exposed via the options `Connection`, so there is no need to expose it now.

If a user wants to specify this option, they will be able to install the `tedious-auth-ntlm` module and pass that option manually, like this:

```js
const ntlmAuthProvider = require('tedious-auth-ntlm');

  ...
  authProvider: ntlmAuthProvider({
    domain: 'DOMAIN',
    workstation: 'WORKSTATION',
    username: 'username',
    password: 'password'
  }),
  ...
});
```

v-suhame:
@arthurschreiber If there is anything I can help with to get the npm release going, I'll be glad to help :) 
If you have a stub in place, I can help with `tedious-auth-native` and some testing.

arthurschreiber:
Ugh, it looks like I accidentally release `2.1.0`. 😞 I won't mention the addition of integrated authentication in the release notes and we will just keep quiet about it until `2.2.0` ships with the pluggable authentication support finished.

Sorry about this! 🙇

v-suhame:
Node.js web app in Azure fails with `2.1.0`, due to sspi-client dependencies :worried: 

I did a quick test by moving native.js from here into sspi-client, it is working. What do you think about publishing `2.1.1` maybe, with just this pluggable-auth in it?

v-suhame:
@arthurschreiber @tvrprasad It's right time to transfer `sspi-client` repository to `tediousjs` organization, what do you think?

tvrprasad:
> It's right time to transfer sspi-client repository to tediousjs organization, what do you think?

I agree. It doesn't make sense to keep making commits under my account as I won't be tracking the repository.

src/connection.js
```diff
@@ -1462,74 +1343,31 @@ Connection.prototype.STATE = {
...
-          return this.sendDataToTokenStreamParser(data);
+        // Use SSPI blob if received
+        if (this.sspiBuffer) {
+          return this.authProvider.handshake(this.sspiBuffer, (error, responseBuffer) => {
+            if (error) {
+              this.emit('error', error);
+              return this.close();
+            }
+            this.messageIo.sendMessage(TYPE.NTLMAUTH_PKT, responseBuffer);
+
```
v-suhame: Send `NTLMAUTH_PKT` only if `responseBuffer` is valid, as not all auth mechanism require this packet. For instance Kerberos handshake is complete in one round trip and doesn't involve `NTLMAUTH_PKT`.

arthurschreiber: During kerberos auth, does the server send back a SSPI token? 🤔

If not, then this code should be okay, as it only sends another `NTLMAUTH_PKT` in case a SSPI token was received. 🤷‍♂️

v-suhame: Server does send SSPI token for Kerberos. Unlike NTLM and Windows integrated auth, Client/Server validation is complete in one round-trip for Kerberos.
 
TDS document in section `2.2.6.8`  says `The client MUST respond with another SSPI message`, but it is not applicable for Kerberos, as it contradicts the handshake explained in RFC4120 😕. You can refer to exchange `(5)` and `(6)` of handshake diagram in https://msdn.microsoft.com/en-us/library/cc233863.aspx, for extended reading you can check section `3.2` in [RFC4120](https://www.ietf.org/rfc/rfc4120.txt).

arthurschreiber: @v-suhame Unfortunately, I can't double check the actual behaviour under kerberos, as I don't have a test environment with domain support set up (and kerberos does not support workstation based login). 😞 

Could you double check this that the server really sends a `SSPI` token during kerberos authentication?

v-suhame: No worries :) 
In Kerberos, after sending `Login7` packet, packet with `SSPI` token and a packet with `ENVCHANGE` token are received one after another. This avoids the need for `NTLMAUTH_PKT` packet as Server has already authenticated the Client.

arthurschreiber: Oh, okay. 🤔 I'll fix that up then! 👍

Hadis-Fard:
@arthurschreiber I have finished AAD (fedauth) authentication based on current master, and able to connect. I was wondering when would you merge this PR, so I can adapt this design, or should i just send the AAD PR as is? 😕

arthurschreiber:
@Hadis-Fard Would be great if you could send the current version of your PR. 👍 This would allow us to take a look and see whether we can integrate the different auth approaches somehow.

v-suhame:
@arthurschreiber Is there anything you need assistance with to wrap up this PR? 😄  Looks like it is almost ready to merge.

arthurschreiber:
:wave: Hey @David-Engel,

I've picked up work on this pull request and #707 - I'm trying right now to come up with an authentication provider interface that will support both integrated and federated authentication schemes. 👍

v-suhame:
> I've picked up work on this pull request and #707 - I'm trying right now to come up with an authentication provider interface that will support both integrated and federated authentication schemes.

Hey @arthurschreiber, any update on this? Is there anything I can help with to get this going?

fenngineering:
When will this be worked on?

drc-schihak:
@arthurschreiber This feature sounds perfect, do you know if this will completed?

pradz85:
I am mainly looking for kerberos authentication support in tedious. Ticket #612 says it would be fixed as part of this ticket. Is there any update ? Thanks.

2160006332:
Has the tedious implementation of Windows authentication been accomplished?

2160006332:
Has the tedious implementation of Windows authentication been accomplished?

I am trying to make the following connection in tedious, however without success.

image

PR #628, opened 2017-10-19
Use Date constructors over unpredictable setters (fixes #625)

### LINKED ISSUE #625: Wrong datetime retrieved from database on DST day
Hello!

I've had a single issue on the day 15, and found out the problem lies in tedious. The issue is repeatable, and predictable, any query in a database on the date 2017-10-15 **in my timezone** returns a datetime with an hour offset (+1). useUTC is disabled.

The issue lies in this line: https://github.com/tediousjs/tedious/blob/d2cf7186c9d8eda4e530f42d5be8072eb6481e85/src/value-parser.js#L519-L521

When the line 520 changes the date to the faulty DST date 2017-10-15, it appropriately changes the time to 01:00:00, since 00:00:00 doesn't **exist**. So, in line 512, the Date object **sums** the 1 hour to the milliseconds from the database, which results in an offset.

The way to fix this is by removing the 1 hour difference from the Date object at that place, which isn't feasible directly because 00:00:00 doesn't exist. So we have to derive the current milliseconds offset in some way like calculating `getHours() * 60 * 60 * 1000` and then subtracting that from `milliseconds`.

I can submit a pull request later, but I really don't know how I would approach writing an unit test for this, since it seems dependent on the Date implementation and the timezone of the user.

fer22f:
As documented in bug #625, setters are a bit on the unpredictable side, `setDate` can, under certain circumstances, change the time of the `Date` object, preventing other calls from working correctly.

Missing unit tests of which I'm unsure of how to approach.

### DIFF
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -492,13 +492,9 @@ function readSmallDateTime(parser, useUTC, callback) {
     parser.readUInt16LE((minutes) => {
       let value;
       if (useUTC) {
-        value = new Date(Date.UTC(1900, 0, 1));
-        value.setUTCDate(value.getUTCDate() + days);
-        value.setUTCMinutes(value.getUTCMinutes() + minutes);
+        value = new Date(Date.UTC(1900, 0, 1 + days, 0, minutes));
       } else {
-        value = new Date(1900, 0, 1);
-        value.setDate(value.getDate() + days);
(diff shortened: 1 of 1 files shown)


fer22f:
Thanks! Yea, we really need to look for a good way to test this...

PR #630, opened 2017-10-23
fix: Do not send username and password via LOGIN7 when a domain is set.

arthurschreiber:
This fixes a potentially security relevant issue where `tedious` would send username and password in plaintext via the LOGIN7 token, even if the `domain` option was specified and the credentials were sent as hashes via NTLM.

The security impact of this is rather low, as NTLM hashes are not really much more secure than plain text credentials, but this is still something we should fix.


arthurschreiber:
@v-suhame Thanks for the review! ❤️

v-suhame:
+1 for the release 😄

Issue #632, opened 2017-10-25
request emits prepared even on error (sometimes multiple times)


andrewsensus:
When connection.prepare(request) encounters an error, the request emits the 'prepared' event rather than emitting an error. If the event-handler responds to the false 'prepared' event by calling connection.execute(request, params) and that also encounters an error, then a second 'prepared' event is emitted rather than calling the request callback (even though connection.prepare(request) was only called once).

steps to reproduce:
1. configure tedious to connect to the db via an ssh forwarded port (surely there are other ways, but ssh port forwarding provides an easy way to interrupt the tcp-connection)
2. setup ssh to forward a port, such as "ssh -L 1433:db-host:1433 localhost"
3. cause tedious to connect and perform some queries to verify the connection is working
4. stop/background ssh (example: enter, ~, control-Z; send it the corresponding signal if you prefer)
5. at this point tedious is still connected at the TCP level, but ssh is no longer forwarding bytes to/from the server
6. prepare a request
7. kill ssh (I used kill -KILL on the ssh pid)
8. observe that 'prepare' is emitted
9. in the handler for the 'prepare' event, attempt to execute the request
10. observe that the 'prepare' event is emitted a second time

v-suhame:
@andrewsensus Thanks for the detailed repro steps 👍 [ `errorMessage`](http://tediousjs.github.io/tedious/api-connection.html#event_errorMessage) event is emitted for all Server Errors. 
For the scenario you described, `errorMessage` will be emitted before `prepared`, application should listen to it and decide on how to handle further listeners.
That would be the generic solution to handle Server error any point in time during connection:)

andrewsensus:
@v-suhame so far my application has avoided handling connection events. It handles calls to the request callback and events emitted by the request but leaves tedious-connection-pool to handle connection events. Perhaps I need to look into that.
How do you feel about the pull request to avoid emitting a 'prepared' event when the prepare wasn't successful?

v-suhame:
I'm skeptical about that, driver already has a generic way to handle error, it would be inconvenient to handle individual cases. 
Do look into tedious-connection-pool and let us know how that goes.

andrewsensus:
@v-suhame I tested this, and I don't see an [errorMessage](http://tediousjs.github.io/tedious/api-connection.html#event_errorMessage) event, but I do see an [error](http://tediousjs.github.io/tedious/api-connection.html#event_error) event. It should be possible for the application to hook into the error event on the connection.

[...]

PR #633, opened 2017-10-25
fix: avoid (repeatedly) emitting 'prepared' event on error

andrewsensus:
'prepared' was emitted on error because the callback error argument was not checked.

'prepared' was emitted again if an error occurred again before the emit call returned
because the prepared flag was only updated after emitting the event.


andrewsensus:
fixes #632

PR #497, opened 2017-02-02
Implementation and tests for Windows Integrated Authentication.

tvrprasad:
- This is currently implemented for Windows only.
- No username/password needed when connecting as domain user.
- Leverages current implementation of NTLM authentication that requires
  username/password.
- Adds a dependency on https://www.npmjs.com/package/sspi-client package
  for implementation of SSPI protocol.
- sspi-client has native code which means the module will be built on
  client machines at Tedious installation time.

Address issue - https://github.com/tediousjs/tedious/issues/415

This is a squashed version of the work done under
https://github.com/tediousjs/tedious/pull/486

PR #635, opened 2017-10-28
Fix: Fix deployments to Azure.

arthurschreiber:
The `sspi-client` that was added to support native windows
authentication requires a small amount of C++ code to be compiled via
`node-gyp` during installation. This is not support on Azure out-of-the
box.

As the native windows authentication was never documented and only
accidentally included in the 2.1.0 release, we can remove the code to
support it to fix deployments for Azure users.

Native windows authentication will be brought back and supported in
future `tedious` releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,7 @@
     "bl": "^1.2.0",
     "iconv-lite": "^0.4.11",
     "readable-stream": "^2.2.6",
-    "sprintf": "0.1.5",
-    "sspi-client": "^0.1.0"
+    "sprintf": "0.1.5"
   },
   "devDependencies": {
     "async": "^1.4.0",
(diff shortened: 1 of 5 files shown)


arthurschreiber:
Hey @v-suhame,

this should fix deployment of node.js applications using tedious to Azure. I'm sorry I had to remove the code you added earlier, but I'll try to get #624 released soon after to bring back support for windows native authentication, in a way that does not break deployments to Azure. 🙇 

Thanks for letting me know about the issue! ❤️

v-suhame:
@arthurschreiber No worries! The changes are going to be back sooner or later 😉  

This PR is good to go if you remove [integration tests](https://github.com/tediousjs/tedious/pull/497/files#diff-33a6e1cce2a4742a36f4fcab71c12643) for windows auth.

arthurschreiber:
@v-suhame Thanks for pointing out these test cases! 🙇

As soon as the test cases pass and this gets merged, a `tedious@2.1.1` will be released. 😄

Issue #589, opened 2017-06-09
bulkLoad and Foreign Key


jllunell:
Hi, I'm trying to load data to Azure SQL Server, and when i use bulkLoad non of the Foreign Keys violations are fired, so i'm able to load corrupted data to the database.

When i use Request, the exception is fired and i can't load the same data that BulkLoad can. So Request is working fine with foreign keys, but I don't know why BulkLoad doesn't.

Anybody knows if it's my misconfiguration or something is wrong with this?
Thank you.

Edited:

I think here is an explanation:
https://dba.stackexchange.com/questions/103095/foreign-keys-become-untrusted-after-bulk-insert

Seems like when you do a bulkLoad the foreign keys are disabled unless you mark an option called CHECK_CONSTRAINTS.

v-suhame:
Fixed in #637

PR #637, opened 2017-11-03
Bulk load options

### LINKED ISSUE #438: Bulkload interface doesn't support FIRE_TRIGGERS, etc
Currently, bulk loading doesn't have the ability to issue these optional statements as part of the bulk load interface.  I've located mono's implementation:

https://github.com/mono/mono/blob/master/mcs/class/System.Data/System.Data.SqlClient/SqlBulkCopy.cs#L446

and in tedious:

https://github.com/tediousjs/tedious/blob/master/src/bulk-load.js#L120

Thanks!

ps - I can send a PR if interested

v-suhame:
fixes #438 and #589
Added options to honor constrains, triggers, null values and table lock during bulk load.

Couldn't add test for lock scenario, its hard to test as result might vary based on speed and network delay.


src/bulk-load.js
```diff
@@ -115,6 +117,16 @@ module.exports = class BulkLoad extends EventEmitter {
...
   }
 
+
+  /**
+   * @param {BULK_LOAD_OPTIONS} bulkOption - options to use during BulkLoad
+   */
+  addOptions(bulkOption) {
+    if (undefined !== BULK_LOAD_OPTIONS[bulkOption]) {
+      this.bulkOptions.push(bulkOption);
+    }
```
chdh:
```
else {
    throw new Error("Unknown bulk load option.");
}
```

src/bulk-load.js
```diff
@@ -115,6 +117,16 @@ module.exports = class BulkLoad extends EventEmitter {
     }
   }
 
+
+  /**
+   * @param {BULK_LOAD_OPTIONS} bulkOption - options to use during BulkLoad
+   */
+  addOptions(bulkOption) {
```
chdh: Check for duplicate options or use `Set` instead of array?

src/bulk-load-options.js
```diff
@@ -0,0 +1,15 @@
...
+ * Enum for BulkLoad options
+ * @readonly
+ */
+module.exports.BULK_LOAD_OPTIONS = {
+  /** Honors constraints during bulk load, it is disabled by default */
+  CHECK_CONSTRAINTS: 'CHECK_CONSTRAINTS',
+  /** Honors insert triggers during bulk load, it is disabled by default */
+  FIRE_TRIGGERS: 'FIRE_TRIGGERS',
+  /** Honors null value passed, ignores the default values set on table */
+  KEEP_NULLS: 'KEEP_NULLS',
```
chdh: The documentation sais that the option is named 'KEEPNULLS' not "KEEP_NULLS", but why did the test case not fail? Is "KEEP_NULLS" an alias?

test/integration/bulk-load-test.js
```diff
@@ -132,3 +133,124 @@ exports.bulkLoadError = function(test) {
...
+      test.ifError(err);
+      bulkLoad.addRow({
+        id: 555
+      });
+      connection.execBulkLoad(bulkLoad);
+    }
+  );
+
+  const request_verify = new Request(verifyTrigger, function(err) {
+    connection.execSql(request_dropTable);
```
chdh: `test.ifError(err);`

test/integration/bulk-load-test.js
```diff
@@ -132,3 +133,124 @@ exports.bulkLoadError = function(test) {
...
+  });
+
+  const request_dropTable = new Request(dropTable, function() {
```
chdh: test err

v-suhame:
@chdh Thanks for reviewing it :) I'll update them with your suggestions.

test/integration/bulk-load-test.js
```diff
@@ -132,3 +133,124 @@ exports.bulkLoadError = function(test) {
...
+        id: 555
+      });
+      connection.execBulkLoad(bulkLoad);
+    }
+  );
+  connection.execSqlBatch(request);
+};
+
+// Test if bulk load honours trigger dependent on destination table
+exports.bulkLoad_Triggers = function(test) {
```
arthurschreiber: If you think it makes sense, you can give test cases more descriptive names by using the `[]` syntax:

```
exports['BulkLoad with FIRE_TRIGGERS'] = ...
```

v-suhame:
Making bulk load options as object does sounds better :)

v-suhame:
@chdh @arthurschreiber Can you review the changes?

src/bulk-load.js
```diff
@@ -115,6 +127,33 @@ module.exports = class BulkLoad extends EventEmitter {
     }
   }
 
+  /**
+   * @param {{checkConstraints : boolean , fireTriggers : boolean , keepNulls : boolean, tableLock: boolean}} bulkLoadOptions
```
arthurschreiber: What do you think about changing this to be JSDoc compatible? Something like:

```js
/**
  @param {Object} bulkLoadOptions
  @param {boolean} bulkLoadOptions.checkConstraints - honors constraints during bulk load, it is disabled by default.
  @param {boolean} bulkLoadOptions.fireTriggers - honors insert triggers during bulk load, it is disabled by default.
  @param {boolean} bulkLoadOptions.keepNulls - honors null value passed, ignores the default values set on table.
  @param {boolean} bulkLoadOptions.tableLock- places a bulk update(BU) lock on table while performing bulk load. Uses row locks by default
*/
```

arthurschreiber: Note: Not putting a `*` in front of each line of a multiline doc comment is my personal preference.

src/bulk-load.js
```diff
@@ -115,6 +127,33 @@ module.exports = class BulkLoad extends EventEmitter {
...
+   *  checkConstraints - honors constraints during bulk load, it is disabled by default.
+   *
+   *  fireTriggers -  honors insert triggers during bulk load, it is disabled by default.
+   *
+   *  keepNulls - honors null value passed, ignores the default values set on table.
+   *
+   *  tableLock - places a bulk update(BU) lock on table while performing bulk load. Uses row locks by default
+   */
+  setOptions(bulkLoadOptions) {
+    if (bulkLoadOptions) {
```
arthurschreiber: Iterating over object properties using `for ... in` is problematic in JavaScript, mostly because _any_ object can be extended with random properties via extensions on `Object.prototype`. Also, I believe it's quite unexpected that an error gets thrown on "unknown" properties.

Another issue here is that `bulkOptions` is modified during the iteration of `bulkLoadOptions`, which can lead the bulk load options to be left in an unexpected state when one of the properties in `bulkLoadOptions` has an invalid value.

What do you think about changing this code to follow precedent set in the Node.js core codebase? E.g. if we take this snippet here https://github.com/nodejs/node/blob/556ebab30e171c2941deb529a42ad47301972af5/lib/child_process.js#L378-L489 as an inspiration, the `setOptions` method could look something like this:

```js
setOptions(options) {
  if (typeof options !== 'object') {
    throw new TypeError('"options" argument must be an object');
  }

  if (options.checkConstraints != undefined && typeof options.checkConstraints !== 'boolean') {
    throw new TypeError('"checkConstraints" must be a boolean');
  }

  if (options.fireTriggers != undefined && typeof options.fireTriggers !== 'boolean') {
    throw new TypeError('"fireTriggers" must be a boolean');
  }

  if (options.keepNulls != undefined && typeof options.keepNulls !== 'boolean') {
    throw new TypeError('"keepNulls" must be a boolean');
  }

  if (options.lockTable != undefined && typeof options.lockTable !== 'boolean') {
    throw new TypeError('"lockTable" must be a boolean');
  }

  if (options.checkConstraints != undefined) {
    this.bulkOptions.checkConstraints = options.checkConstraints;
  }

  if (options.fireTriggers != undefined) {
    this.bulkOptions.fireTriggers = options.fireTriggers;
  }

  if (options.keepNulls != undefined) {
    this.bulkOptions.keepNulls = options.keepNulls;
  }

  if (options.lockTable != undefined) {
    this.bulkOptions.lockTable = options.lockTable;
  }
}
```

This code is overall a bit longer, but fixes the aforementioned issues and is more straightforward (and thus overall easier to maintain). 👍

src/bulk-load.js
```diff
@@ -125,6 +164,17 @@ module.exports = class BulkLoad extends EventEmitter {
       sql += '[' + c.name + '] ' + (c.type.declaration(c));
     }
     sql += ')';
+
+    // filter bulkOptions that are true and get their corresponding T-SQL keyword
```
arthurschreiber: What do you think about removing the need to use `filter` and `map` by being a bit more explicit here?

```js
const addOptions = [];

if (this.bulkOptions.checkConstraints) {
  addOptions.push('CHECK_CONSTRAINTS');
}

if (this.bulkOptions.fireTriggers) {
  addOptions.push('FIRE_TRIGGERS');
}

if (this.bulkOptions.keepNulls) {
  addOptions.push('KEEP_NULLS');
}

if (this.bulkOptions.lockTable) {
  addOptions.push('TABLOCK');
}
```

src/bulk-load.js
```diff
@@ -125,6 +164,17 @@ module.exports = class BulkLoad extends EventEmitter {
...
+    const addOptions = Object.keys(this.bulkOptions)
+      .filter((key) => {
+        return this.bulkOptions[key] === true;
+      }).map((key) => {
+        return this.bulkOptionMapping.get(key);
+      });
+
+    if (addOptions.length !== 0)
+      sql += 'WITH (' + addOptions.join(',') + ')';
```
arthurschreiber: What do you think about using template literals here?

```js
sql += `WITH (${addOptions.join(',')})`;
```

src/bulk-load.js
arthurschreiber: This would also allow you to remove the `bulkOptionMapping` property.

src/bulk-load.js
chdh: I would completely replace the previous `this.bulkOptions` here, instead of modifying them:

```
setOptions(options) {
   const newOptions = {};
   newOptions.checkContraints = getBooleanOption('checkContraints', false);
   newOptions.fireTriggers = getBooleanOption('fireTriggers', false);
   newOptions.keepNulls = getBooleanOption('keepNulls', false);
   newOptions.lockTable = getBooleanOption('lockTable', false);
   this.bulkOptions = newOptions;

   function getBooleanOption(optionName, defaultValue) {
      const optionValue = options[optionName];
      if (optionValue == undefined) {
         return defaultValue;
      }
      if (typeof optionValue !== 'boolean')  {
         throw new TypeError(´"${optionName}" must be a boolean');
      }
      return optionValue;
   }
}
```

Then we can initialize `this.bulkOptions` within the class constructor with:

```
   this.setOptions({});
```

src/bulk-load.js
v-suhame: Somewhere along the line I got confused between DocumentJS and JSDoc and ended up with that comment 😅

src/bulk-load.js
v-suhame: Thank you for detailing it :)

src/bulk-load.js
v-suhame: Since its just 4 options, being explicit should be fine 👍

v-suhame:
@arthurschreiber @chdh Thanks a lot for taking time to review it and for the detailed feedback 😃 I have reused most of code snippet you guys shared, hope that's fine with you.

src/bulk-load.js
arthurschreiber: No worries! I think in general, documentation.js is compatible with JSDoc, so we're good here now! 🎉

src/bulk-load.js
```diff
@@ -115,6 +116,64 @@ module.exports = class BulkLoad extends EventEmitter {
...
+    @param {boolean} bulkLoadOptions.keepNulls - honors null value passed, ignores the default values set on table.
+    @param {boolean} bulkLoadOptions.tableLock- places a bulk update(BU) lock on table while performing bulk load. Uses row locks by default.
+  */
+  setOptions(bulkLoadOptions) {
+    if (typeof bulkLoadOptions !== 'object') {
+      throw new TypeError('"bulkLoadOptions" argument must be an object');
+    }
+
+    const bOptions = {};
+    bOptions.checkConstraints = getBooleanOption('checkConstraints', false);
```
arthurschreiber: I believe the current version of `setOptions` behaves differently than your initial version, and differently than how I as a user would expect it to behave.

I'd expect multiple `setOptions` calls to only change the options referred by the keys of the `options` object that is passed in, like this:

```
bulkLoad.setOptions({ fireTrigger: true });
bulkLoad.setOptions({ keepNulls: true });
// => bulkLoad should have both fireTriggers and keepNulls enabled
```

But I don't think this is how it works now. 😞 

@v-suhame What do you think would be the correct approach here?

chdh: @arthurschreiber I would expect that `setOptions()` replaces the previous options and that the missing options get the default value. Otherwise I would expect it to be named `addOptions()` or `updateOptions()`.

If the options could be passed with `connection.newBulkLoad()`, we would not need `setOptions()` at all.

v-suhame: >// => bulkLoad should have both fireTriggers and keepNulls enabled

Making this change would affect existing users. Its best to let the user decide if they want to enable these options, and moreover it is the same behavior as other SQL Server drivers.  

>I would expect that setOptions() replaces the previous options and that the missing options get the default value.

@arthurschreiber This was the intended behavior after we moved from `addOptions()` 👆.

arthurschreiber: @v-suhame What do you think about the suggestion from @chdh to move the options to the newbulkload method instead of having a setOptions method?

v-suhame: Are we looking for something like the below snippet after change? 

````node
  connection.newBulkLoad(tableName, {
      callback: cb,
      bulkLoadOptions : {
        keepNulls: true,
        checkConstraints: true
        ..
      }
  });
```

chdh: @v-suhame Yes, or simpler with all options mixed at the same level:

```
connection.newBulkLoad(tableName, {
      callback: cb,
      keepNulls: true,
      checkConstraints: true
      ..
  });
```

arthurschreiber: @v-suhame @chdh I'd prefer if we could stick to the convention of having the callback as the last argument of the function, instead of being part of an options object. 🙇

We'd also need to keep backwards compatibility. What about something like this:

```js
newBulkLoad(table, options, callback) {
  if (callback === undefined) {
     callback = options;
     options = undefined;
  }

  return new BulkLoad(table, this.config.options, options, callback);
}
```

and then in `bulk-load.js`:

```js
  constructor(table, connectionOptions, {
    checkConstraints = false,
    fireTriggers = false,
    keepNulls = false,
    lockTable = false,
  }, callback) {
    super();

    if (typeof checkConstraints !== 
'boolean') {
      throw new TypeError('The "options.checkConstraints" property must be of type boolean.');
}

    if (typeof fireTriggers !== 
'boolean') {
      throw new TypeError('The "options.fireTriggers" property must be of type boolean.');
}

    if (typeof keepNulls !== 
'boolean') {
      throw new TypeError('The "options.keepNulls" property must be of type boolean.');
}

    if (typeof lockTable !== 
'boolean') {
      throw new TypeError('The "options.lockTable" property must be of type boolean.');
}

    this.error = undefined;
    this.canceled = false;

    this.table = table;
    this.options = connectionOptions;
    this.callback = callback;
    this.columns = [];
    this.columnsByName = {};
    this.rowsData = new WritableTrackingBuffer(1024, 'ucs2', true);
    this.firstRowWritten = false;

    this.bulkOptions = { checkConstraints, fireTriggers, keepNulls, lockTable };
  }
```

arthurschreiber:
This is looking really great. ❤️

I especially like that the bulk load options are set are set on creation and can't be changed afterwards, leading to a simpler API overall!

I'll merge this a bit later today to trigger a new release. 👍

v-suhame:
@chdh @arthurschreiber Thank you for the review and feedback 😃

PR #640, opened 2017-11-07
chore: Add `flow`.

arthurschreiber:
This adds two new dependencies, `flow-bin` and `babel-preset-flow`, allowing us to add type information to `tedious`. Type information will only be "internal" for now, while we build out the use of types in the `tedious` code base.

Once we have enough type information defined to be useful for end users, we can use the `flow-copy-source` package to make sure types get packaged up and published in the `tedious` npm package as well. 👍

~~This also adds some very basic type information to the `Request` related code inside `src/request.js`, as a first step.~~


arthurschreiber:
I changed this pull request to only contain the addition of `flow` libraries and configuration. I'll open new pull requests to gradually add type information. 👍

PR #641, opened 2017-11-09
chore: Upgrade Node.js versions used for CI.

arthurschreiber:
This updates the Node.js versions that we use to run our tests on during CI on Travis and Appveyor. Node.js 7.x is no longer maintained or supported, 8.x became the new long-term support release, and new development happens on 9.x now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ cache:
     - node_modules
 
 before_install:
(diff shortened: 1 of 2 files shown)


arthurschreiber:
Thanks for the review @v-suhame! ❤️

PR #643, opened 2017-11-10
fix: connection failure in Azure SQL Data Warehouse

### LINKED ISSUE #631: Connection failure - Azure SQL Data Warehouse
Connection to Azure SQL Data Warehouse fails with `ERROR: 'CursorCloseOnCommit' is not a recognized option.`
[Cursor_close_on_commit](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-cursor-close-on-commit-transact-sql) is not supported by DW, since this is OFF by default, if user doesn't set it in `config`, avoid sending it in initial SQL.

v-suhame:
> Connection to Azure SQL Data Warehouse fails with ERROR: 'CursorCloseOnCommit' is not a recognized option.
Cursor_close_on_commit is not supported by DW, since this is OFF by default, if user doesn't set it in config, avoid sending it in initial SQL.

https://docs.microsoft.com/en-us/sql/t-sql/statements/set-cursor-close-on-commit-transact-sql

Fixes #631

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -962,7 +962,6 @@ class Connection extends EventEmitter {
     const enableAnsiWarnings = this.config.options.enableAnsiWarnings ? 'on' : 'off';
     const enableArithAbort = this.config.options.enableArithAbort ? 'on' : 'off';
     const enableConcatNullYieldsNull = this.config.options.enableConcatNullYieldsNull ? 'on' : 'off';
-    const enableCursorCloseOnCommit = this.config.options.enableCursorCloseOnCommit ? 'on' : 'off';
     const enableImplicitTransactions = this.config.options.enableImplicitTransactions ? 'on' : 'off';
     const enableNumericRoundabort = this.config.options.enableNumericRoundabort ? 'on' : 'off';
     const enableQuotedIdentifier = this.config.options.enableQuotedIdentifier ? 'on' : 'off';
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Once the current build on master finishes, this should be release as `tedious@2.1.4`. 🎉

arthurschreiber:
Now that I merged and released this, I'm wondering whether this is truly the right fix here. 😅 

If I understand https://docs.microsoft.com/en-us/sql/t-sql/statements/set-cursor-close-on-commit-transact-sql correctly, the server wide value for `cursor_close_on_commit` can be changed from the default `false` value. So if someone would do that, there would not be any way to set this value to `false` again on login.

This is not something we need to fix now, but should revisit some time later in the future.

@v-suhame What do you think?

v-suhame:
We sure need to fix that, sorry my bad, I should have tested that 😞

v-suhame:
How about setting `this.config.options.enableCursorCloseOnCommit` as `undefined` by default, 
and  in getInitialSql()
```node
getInitialSql() {
 ..
    const getCursorCloseOnCommit = () => {
      if (this.config.options.enableCursorCloseOnCommit === undefined)
        return '';
      else
        return 'set cursor_close_on_commit ' + (this.config.options.enableCursorCloseOnCommit ? 'on' : 'off') + '\n';
    };

...
return `...
${getCursorCloseOnCommit ()}
...
```

PR #649, opened 2017-11-13
refactor: Add type information to `src/request.js`

arthurschreiber:
This adds type information to the `src/request.js` file, mainly adding type definitions for the `Request` class.


src/request.js
```diff
@@ -1,34 +1,77 @@
...
     this.canceled = false;
     this.paused = false;
+    this.error = undefined;
+    this.connection = undefined;
     this.userCallback = callback;
     this.callback = function() {
       if (this.preparing) {
         this.emit('prepared');
-        return this.preparing = false;
+        this.preparing = false;
```
arthurschreiber: I removed the `return` statements here, because I don't think these make sense, really, and they conflict the `void` return type of the `CompletionCallback`.

src/request.js
```diff
@@ -1,34 +1,77 @@
...
+
+  callback: () => void;
+
+  constructor(sqlTextOrProcedure: ?string, callback: CompletionCallback) {
     super();
 
     this.sqlTextOrProcedure = sqlTextOrProcedure;
     this.parameters = [];
     this.parametersByName = {};
+    this.originalParameters = [];
```
arthurschreiber: I changed the constructor to initialize all properties that are used inside the `Request` class.

This simplifies the type signature of the class, and will help v8 to perform more/better optimizations for this code. 🎉

v-suhame:
It's a good start 🎉

arthurschreiber:
@v-suhame Thanks for taking a look! 👀 ❤️

PR #650, opened 2017-11-13
refactoring: Add type information to `src/tracking-buffer/writable-tracking-buffer.js`.

arthurschreiber:


src/tracking-buffer/writable-tracking-buffer.js
```diff
@@ -1,27 +1,40 @@
+// @flow
+
 const bigint = require('./bigint');
 
 const SHIFT_LEFT_32 = (1 << 16) * (1 << 16);
 const SHIFT_RIGHT_32 = 1 / SHIFT_LEFT_32;
 const UNKNOWN_PLP_LEN = new Buffer([0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
 const ZERO_LENGTH_BUFFER = new Buffer(0);
 
+export type Encoding = 'utf8' | 'ucs2' | 'ascii';
```
arthurschreiber: I _think_ these are all the encodings that we use together with `WritableTrackingBuffer`. If we find more later one, we can add them here.

src/tracking-buffer/writable-tracking-buffer.js
```diff
@@ -1,27 +1,40 @@
...
+  constructor(initialSize: number, encoding: ?Encoding, doubleSizeGrowth: ?boolean) {
     this.initialSize = initialSize;
-    this.encoding = encoding;
-    this.doubleSizeGrowth = doubleSizeGrowth;
-    this.doubleSizeGrowth || (this.doubleSizeGrowth = false);
-    this.encoding || (this.encoding = 'ucs2');
+    this.encoding = encoding || 'ucs2';
+    this.doubleSizeGrowth = doubleSizeGrowth || false;
     this.buffer = new Buffer(this.initialSize);
+    this.compositeBuffer = new Buffer(0);
```
arthurschreiber: I changed the `compositeBuffer` to be initialized in the constructor. This simplifies the `newBuffer` method a tiny bit and keeps us from having to check whether `this.compositeBuffer` is `undefined` or not before use.

src/tracking-buffer/writable-tracking-buffer.js
```diff
@@ -93,137 +102,140 @@ module.exports = class WritableTrackingBuffer {
...
 
-  writePLPBody(value, encoding) {
+  writePLPBody(value: any, encoding: ?Encoding) {
     if (encoding == null) {
       encoding = this.encoding;
     }
 
     let length;
-    if (Buffer.isBuffer(value)) {
+    if (value instanceof Buffer) {
```
arthurschreiber: Flow does not understand that the `Buffer.isBuffer` check refines the type of the `value` variable. But switching to an `instanceof` check here does the trick.

src/tracking-buffer/writable-tracking-buffer.js
```diff
@@ -239,11 +251,11 @@ module.exports = class WritableTrackingBuffer {
     if (length > 0) {
       // One chunk.
       this.writeUInt32LE(length);
-      if (Buffer.isBuffer(value)) {
+      if (value instanceof Buffer) {
         this.writeBuffer(value);
       } else {
         this.makeRoomFor(length);
-        this.buffer.write(value, this.position, encoding);
+        this.buffer.write(value, this.position, this.buffer.length - this.position, encoding);
```
arthurschreiber: Flow does not have all the different possible signatures for `Buffer.prototype.write` defined, so I had to add passing a `length` value here.

chdh: Would this work?

```javascript
this.buffer.write(value, this.position, undefined, encoding);
```

arthurschreiber: Unfortunately not. 😞

I'll change this to use the special `FlowFixMe` comment for now to ignore this and open a PR against flow to get it fixed there. 👍

Issue #576, opened 2017-05-09
Code cleanup: Remove unneeded return statements from code


tvrprasad:
There are a bunch of return statements across the code base that are artifacts of conversion from CoffeeScript to JavaScript. These should be cleaned up.

arthurschreiber:
I think #651 solved most of these, so I'm going to close this one! 👍 😄

PR #651, opened 2017-11-18
refactor: Reduce the number of explicit `return` statements.

arthurschreiber:
While working on adding flow type information to the `tedious` source code, some CoffeeScript leftovers started to get in the way. 😞 

One of the leftovers from the conversion are stray `return` statements. In CoffeeScript, the last statement in a function was implicitly and automatically used as the return value of the function. This is really annoying when adding types to functions, as that means functions can have different result types depending on the execution path taken through the function. Even more annoying when we consider the fact that many of these functions were not supposed to have a return value in the first place (and the `return` statement generated by CoffeeScript returned useless information).

This gets rid of many of these return statements, cleaning up the code and paving the way for adding type information later. 👍

### DIFF
--- a/src/bulk-load.js
+++ b/src/bulk-load.js
@@ -106,7 +106,7 @@ module.exports = class BulkLoad extends EventEmitter {
 
     this.columns.push(column);
 
-    return this.columnsByName[name] = column;
+    this.columnsByName[name] = column;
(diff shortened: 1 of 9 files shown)


chdh:
resolves #576

PR #652, opened 2017-11-20
Allow length of 65535 for images. 

mgetz:
readBinary would prevent image lengths of 65535 by doing a NULL check. This fix skips the NULL check and moves directly to readBuffer to retrieve the Image.
I am extracting 50k+ images from a database and a couple of the images had this issue and were throwing errors because the requested image data was assumed to be invalid and was not extracted and returned. This would then cause the stream parser (token/stream-parser.js) to throw errors on this line: 
`this.emit('error', new Error('Unknown type: ' + type));` 
because it did not know how to parse the remaining image data.

### DIFF
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -204,14 +204,14 @@ function valueParse(parser, metaData, options, callback) {
           if (textPointerNull) {
             return callback(null);
           } else {
-            return readChars(parser, dataLength, metaData.collation.codepage, callback);
+            return readText(parser, dataLength, metaData.collation.codepage, callback);
           }
 
         case 'NText':
           if (textPointerNull) {
             return callback(null);
           } else {
-            return readNChars(parser, dataLength, callback);
+            return readNText(parser, dataLength, callback);
           }
 
         case 'Image':
@@ -387,6 +387,30 @@ function readChars(parser, dataLength, codepage, callback) {
 
 function readNChars(parser, dataLength, callback) {
   if (dataLength === NULL) {
+    return callback(null);
+  } else {
+    return parser.readBuffer(dataLength, (data) => {
+      callback(data.toString('ucs2'));
+    });
+  }
+}
+
+function readText(parser, dataLength, codepage, callback) {
+  if (codepage == null) {
+    codepage = DEFAULT_ENCODING;
+  }
+
+  if (dataLength === PLP_NULL) {
+    return callback(null);
+  } else {
+    return parser.readBuffer(dataLength, (data) => {
+      callback(iconv.decode(data, codepage));
+    });
+  }
+}
+
+function readNText(parser, dataLength, callback) {
+  if (dataLength === PLP_NULL) {
     return callback(null);
   } else {
     return parser.readBuffer(dataLength, (data) => {
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -218,7 +218,7 @@ function valueParse(parser, metaData, options, callback) {
           if (textPointerNull) {
             return callback(null);
           } else {
-            return readBinary(parser, dataLength, callback);
+            return parser.readBuffer(dataLength, callback);
           }
 
         case 'Xml':
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -181,7 +181,7 @@ function valueParse(parser, metaData, options, callback) {
           if (metaData.dataLength === MAX) {
             return readMaxChars(parser, codepage, callback);
           } else {
-            return readChars(parser, dataLength, codepage, callback);
+            return readChars(parser, dataLength, codepage, NULL, callback);
           }
 
         case 'NVarChar':
@@ -189,7 +189,7 @@ function valueParse(parser, metaData, options, callback) {
           if (metaData.dataLength === MAX) {
             return readMaxNChars(parser, callback);
           } else {
-            return readNChars(parser, dataLength, callback);
+            return readNChars(parser, dataLength, NULL, callback);
           }
 
         case 'VarBinary':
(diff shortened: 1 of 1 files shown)


v-suhame:
@mgetz Can you share more detail about this issue? I tried to read a random image of 100K+ `DATALENGTH`, it was working fine. 
Are you referring to [`DATALENGTH`](https://docs.microsoft.com/en-us/sql/t-sql/functions/datalength-transact-sql) when you mention `image length`?

mgetz:
Suraiya,
Thank you for your response and taking the time to investigate this issue! Most if not all of this you most likely already know but since I was unfamiliar with the tedious code, this is how I found the issue.

My use case is extracting images and their corresponding records from MSSQL for a client. This batch is only ~50k images and of these images, they ranged in size from 30k to 300k. There were only a couple of images that would fail and this is what I discovered was happening.
In value-parse.js around line 85 `readDataLength(parser, type, metaData, textPointerNull, (dataLength) => {` would get the dataLength to extract for the image. In the case of these failing images, the dataLength is 65535. The `type.name` is 'Image' so it would drop down to the case on line 217 `        case 'Image':
          if (textPointerNull) {
            return callback(null);
          } else {
            return readBinary(parser, dataLength, callback);
          }`
textPointerNull is a falsy value so it goes into the readBinary call with a datalength of 65535.

readBinary is the following code: 
`function readBinary(parser, dataLength, callback) {
  if (dataLength === NULL) {
    return callback(null);
  } else {
    return parser.readBuffer(dataLength, callback);
  }
}`.
NULL has a definition of `const NULL = (1 << 16) - 1;` which equals 65535.
Since the length of this image was exactly 65535, it would return with a null value and not remove the image data from the buffer. 
Thus, I would not get my image and the buffer now held data that was unexpected and would not be understood by the next steps in the process, causing stream-parser.js to emit an error saying that it encountered an unknown type: 
`      if (tokenParsers[type]) {
        tokenParsers[type](this, this.colMetadata, this.options, doneParsing);
      } else {
        this.emit('error', new Error('Unknown type: ' + type));
      }
`
My fix was simply to call the readBuffer function with the correct dataLength (of 65535 in this case) and bypass the 'NULL' check that was in readBinary.

This fail case is very specific. Large and small image sizes work great. It is only when the image size is exactly 65535 in length that it will fail in this way.

v-suhame:
Thanks for explaining it in detail :) 
The problem must exist in all the char and binary types, they all refer to `NULL` aka `65535` 🙁 I will investigate a bit more and keep you posted.

mgetz:
Guess I should have read the documentation before submitting that change. :-(

> Char and binary data types have values that are either null or 0 to 65534 (0x0000 to 0xFFFE) bytes in
length. Null is represented by a length of 65535 (0xFFFF). A char or binary, which cannot be null, can
still have a length of zero (for example an empty value). A program that MUST pad a value to a fixed
length will typically add blanks to the end of a char and binary zeros to the end of a binary.
Text and image data types have values that are either null, or 0 to 2 gigabytes (0x00000000 to
0x7FFFFFFF bytes) in length. Null is represented by a length of -1 (0xFFFFFFFF). No other length
specification is supported.

v-suhame:
No worries :) It has many details, can be overwhelming. 

In value-parser, image and text are treated as binary and char respectively, so it checks for null value 0xFFFF
>Char and binary data types have values that either are null or are 0 to 65534 (0x0000 to 0xFFFE) bytes of data. Null is represented by a length of 65535 (0xFFFF).

We should update it to check for null 0xFFFFFFFF aka `PLP_NULL` instead of `NULL`
> Text and image data types have values that either are null or are 0 to 2 gigabytes (0x00000000 to 0x7FFFFFFF bytes) of data. Null is represented by a length of -1 (0xFFFFFFFF).

Can you update your PR and extend the fix for `text` and `ntext` types too.?

mgetz:
Yes! Thank you!

src/value-parser.js
```diff
@@ -218,7 +218,7 @@ function valueParse(parser, metaData, options, callback) {
           if (textPointerNull) {
             return callback(null);
           } else {
-            return readBinary(parser, dataLength, callback);
+            return parser.readBuffer(dataLength, callback);
           }
```
v-suhame: This fix doesn't handle `null` data, how about making call to something like

```node
function readImage(parser, dataLength, callback) {
  if (dataLength === PLP_NULL) {
    return callback(null);
  } else {
    return parser.readBuffer(dataLength, callback);
  }
}
```

mgetz:
Suraiya,
I implemented your suggested fix and then extended the PLP_NULL check to the Text and NText data types.
That should be more accurate this time.
Thanks!
Mark

src/value-parser.js
```diff
@@ -387,6 +395,30 @@ function readNChars(parser, dataLength, callback) {
   }
 }
 
+function readText(parser, dataLength, codepage, callback) {
```
v-suhame: @mgetz since `readNText`, `readText` and  `readImage` are exactly same as `readChars`, `readNChars` and `readBinary`, except for the null check, it would be good to get rid of redundant function. How about passing additional parameter to the original methods ?
```node
function readBinary(parser, dataLength, nullType /*receives PLP_NULL or NULL*/, callback) {
  if (dataLength === nullType) {..}
  ...
}
```
Sorry for asking you to update the PR again 😅

mgetz: No problem at all. Had considered it when I added those two extra functions. :-)

mgetz:
Removed duplicate functions for a cleaner solution. Thanks Suaiya!

v-suhame:
@mgetz Thanks for the PR :)

arthurschreiber:
I'm not really sure what's going on here, but it looks like the travis build for this branch is not getting triggered. 😞

arthurschreiber:
@mgetz According to the travis request log, the build on your pull request is not getting triggered because "Abuse detected". Any idea what the cause for this could be? I'll be merging this pull request anyway, because the fix looks good to me!

@v-suhame Thank you for guiding @mgetz on this! ❤️ 😍

mgetz:
@arthurschreiber No idea why it triggered an "Abuse detected" message. 
Thanks for getting this merged and thanks again to @v-suhame!

v-suhame:
@mgetz if you haven't already, you might want to mail support@travis-ci.com to get yourself out of that list.

PR #653, opened 2017-11-25
fix: Only set `cursor_close_on_commit` if explicitly specified.

arthurschreiber:
This is a follow up to #643 and fine-tunes the way the `enableCursorCloseOnCommit` connection option is handled.

Instead of only sending the `cursor_close_on_commit` option in the initial sql query payload if the `enableCursorCloseOnCommit` is explicitly set to `true` (without any option to disable `cursor_close_on_commit`), `tedious` now defaults to not sending `cursor_close_on_commit`. But if `enableCursorCloseOnCommit` is set to `true` or `false` explicitly, we either enable or disable this option in the initial sql query payload.

This way, users can decide whether they want to have the server's default `cursor_close_on_commit` value (by not specifying `enableCursorCloseOnCommit`), or explicitly set it to their required value.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -956,33 +956,81 @@ class Connection extends EventEmitter {
   }
 
   getInitialSql() {
-    const enableAnsiNull = this.config.options.enableAnsiNull ? 'on' : 'off';
-    const enableAnsiNullDefault = this.config.options.enableAnsiNullDefault ? 'on' : 'off';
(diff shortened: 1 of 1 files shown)


v-suhame:
@arthurschreiber Thanks for taking care of this option :)

arthurschreiber:
@v-suhame published a new version of tedious (`tedious@2.1.5`) that contains this fix. I also tagged it as `latest`, so all users should be able to connect to Azure SQL Data Warehouse now. Thank you so much! ❤️

PR #661, opened 2017-12-18
ActiveDirectoryPassword and sqlpassword authentication 

Hadis-Fard:
- wrote prelogin for fedauth 
- wrote login7 and response for fedauth
- added adal-node lib to generate token

- added `authentication` keyword to config options, the valid values are `ActiveDirectoryPassword` and `SQLPassword` (at the moment), where sqlpassword is added for consistency and is only used for secure connection to on-premise server. for AD password authentication, user should be a valid AAD user. You might want to set trustServerCertificate to true if you don't have certificate on client machine.

``` js
// ActiveDirectoryPassword authentication method
new Connection({
  "server": "yourADserver.database.windows.net",
  "userName": "yourADuser",
  "password": "yourStrong(!)Password",
  "options": {
    "database": "FederatedAuthDB",
   "authentication": "ActiveDirectoryPassword",
    "trustServerCertificate" : true,
  }
});
```

```js
// SQLPassword authentication method
new Connection({
  "server": "localhost",
  "userName": "sa",
  "password": "yourStrong(!)Password",
  "options": {
    "database": "yourDB",
    "authentication": "SQLPassword",
    "trustServerCertificate" : true
  }
});
```

- TODOs:
    - ActiveDirectoryIntegrated support
    - SecurityToken support
    - unit tests


arthurschreiber:
@Hadis-Fard Thank you for opening this pull request! ❤️ 😍 

I'll try to take a look over the next few days and see if I can leave a few comments.

src/connection.js
```diff
@@ -353,6 +371,29 @@ class Connection extends EventEmitter {
...
+        // check for valid options
+        if (!(config.options.authentication.toUpperCase() === this.fedAuthInfo.ValidFedAuthEnum.SqlPassword ||
+        config.options.authentication.toUpperCase() === this.fedAuthInfo.ValidFedAuthEnum.ActiveDirectoryPassword)) {
+          throw new Error('An invalid authentication method is specified');
+        }
+        if (this.config.options.tdsVersion < '7_4') {
+          throw new Error(`Azure Active Directory authentication is not supported in the TDS version ${this.config.options.tdsVersion}`);
+        }
+        this.config.options.encrypt = true;
+        this.fedAuthInfo.method = config.options.authentication;
```
v-suhame: Just an FYI, `encrypt` will be true by default once #606 is merged.

src/connection.js
```diff
@@ -353,6 +371,29 @@ class Connection extends EventEmitter {
...
+        if (config.options.trustServerCertificate != 'true') {
+          this.config.options.trustServerCertificate = false;
+        }
+      }
+    }
+
+    if (this.config.domain && !this.config.userName && !this.config.password) {
+      this.config.options.useWindowsIntegratedAuth = true;
```
v-suhame: Can you remove the reference to WindowsIntegratedAuth 😅

src/connection.js
```diff
@@ -353,6 +371,29 @@ class Connection extends EventEmitter {
...
+        config.options.authentication.toUpperCase() === this.fedAuthInfo.ValidFedAuthEnum.ActiveDirectoryPassword)) {
+          throw new Error('An invalid authentication method is specified');
+        }
+        if (this.config.options.tdsVersion < '7_4') {
+          throw new Error(`Azure Active Directory authentication is not supported in the TDS version ${this.config.options.tdsVersion}`);
+        }
+        this.config.options.encrypt = true;
+        this.fedAuthInfo.method = config.options.authentication;
+        if (config.options.trustServerCertificate != 'true') {
+          this.config.options.trustServerCertificate = false;
```
v-suhame: `this.config.options.trustServerCertificate` is already has a boolean value (set [here](https://github.com/tediousjs/tedious/pull/661/files#diff-875da84119066ec244742e95d1209f0cR363)), can you remove this validation?

src/connection.js
```diff
@@ -859,7 +918,18 @@ class Connection extends EventEmitter {
     this.debug.payload(function() {
       return preloginPayload.toString('  ');
     });
-
+    if (0 != preloginPayload.fedAuthRequired && 1 != preloginPayload.fedAuthRequired) {
+      this.emit('connect', ConnectionError('Server sent an unexpected value for FedAuthRequired PreLogin Option. Value was' + preloginPayload.fedAuthRequired, 'EFEDAUTH'));
+      return this.close();
```
v-suhame: This error would be triggered for all older versions of SQL Server that doesn't support fedAuth 🤔 How about changing the condition such that error is triggered only when client requests fedAuth and server responds with something other than 1 or 0?

src/connection.js
Hadis-Fard: it could be set to anything (for example it could be 'abcd') up there, and I wanna make sure unless it is explicitly set to true by user, it is set to false.

src/connection.js
Hadis-Fard: Isn't better to drop it after the PR is merged? cause without encryption fed auth would fail, so I think it's better to set it to true for now

src/connection.js
v-suhame: You're right! For most of the boolean options, error is thrown if value is not boolean (like [here](https://github.com/tediousjs/tedious/pull/661/files#diff-875da84119066ec244742e95d1209f0cR274)), looks like `config.options.trustServerCertificate` was missed. Can you do the same for `trustServerCertificate` too? It would be more generic.

src/prelogin-payload.js
```diff
@@ -63,7 +64,8 @@ module.exports = class PreloginPayload {
       this.createEncryptionOption(),
       this.createInstanceOption(),
       this.createThreadIdOption(),
-      this.createMarsOption()
+      this.createMarsOption(),
+      ...(this.options.fedAuthRequested ? [this.createFedAuthOption()] : [])
```
v-suhame: I'm trying to understand the need for spread operator here. `createFedAuthOption()` returns single object, so when is `...` put into use?

src/connection.js
```diff
@@ -860,7 +916,18 @@ class Connection extends EventEmitter {
...
       return preloginPayload.toString('  ');
     });
-
+    if (this.fedAuthInfo.method != undefined) {
+      if (0 != preloginPayload.fedAuthRequired && 1 != preloginPayload.fedAuthRequired) {
+        this.emit('connect', ConnectionError('Server sent an unexpected value for FedAuthRequired PreLogin Option. Value was' + preloginPayload.fedAuthRequired, 'EFEDAUTH'));
+        return this.close();
+      }
+      // fedAuthRequired is used for capability negotiation when choosing between SSPI and federated authentication
+      if (preloginPayload.fedAuthRequired === 1 && this.config.domain) {
```
v-suhame: What do you think about adding TODO/Note here so that we remember to extend this condition to include other SSPI auths (Windows integrated and Kerberos) in future.

src/connection.js
```diff
@@ -860,7 +916,18 @@ class Connection extends EventEmitter {
     this.debug.payload(function() {
       return preloginPayload.toString('  ');
     });
-
+    if (this.fedAuthInfo.method != undefined) {
+      if (0 != preloginPayload.fedAuthRequired && 1 != preloginPayload.fedAuthRequired) {
+        this.emit('connect', ConnectionError('Server sent an unexpected value for FedAuthRequired PreLogin Option. Value was' + preloginPayload.fedAuthRequired, 'EFEDAUTH'));
```
v-suhame: Can you use template strings for error message? It will be easier to keep track of spaces while appending variables. 
Also, I think this message is too technical for users to understand 😅 `FedAuthRequired PreLogin Option`

src/token/feature-ext-parser.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+      });
+    });
+  });
+};
+
+module.exports.fedAuthInfoParser = fedAuthInfoParser;
+function fedAuthInfoParser(parser, colMetadata, options, callback) {
+  parser.readUInt32LE((tokenLength) => {
+    parser.readUInt32LE((countOfInfoIDs) => {
```
arthurschreiber: Nesting many `parser.readXYZ` calls is really expensive and slow, due to callback nesting involved. 😞

Could we change this to instead just read `tokenLength` via `parser.readUInt32LE`, and then read the whole token into a `Buffer` in one go via `parser.readBuffer(tokenLength, (data) => { ... })`. All other data extraction could then happen using regular `Buffer` operations, without any callbacks being involved.

@Hadis-Fard What do you think? 😄

src/token/feature-ext-parser.js
```diff
@@ -0,0 +1,79 @@
...
+      async.times(countOfInfoIDs, () => {
+        readFedAuthInfoOpt(parser, callback);
+      });
+    });
+  });
+}
+
+module.exports.featureExtAckParser = featureExtAckParser;
+function featureExtAckParser(parser, colMetadata, options, callback) {
```
arthurschreiber: What do you think about breaking this up into two separate files, one for parsing `FEDAUTHINFO` and the other for `FEATUREEXTACK`?

src/prelogin-payload.js
arthurschreiber: Yeah, this code is a tiny bit convoluted. 😅 

What do you think about changing this to:

```js
    const options = [
      this.createVersionOption(),
      this.createEncryptionOption(),
      this.createInstanceOption(),
      this.createThreadIdOption(),
      this.createMarsOption()
    ];

    if (this.options.fedAuthRequested) {
      options.push(this.createFedAuthOption());
    }
```

I think this might be less confusing. 😄

src/login7-payload.js
```diff
@@ -256,6 +316,53 @@ module.exports = class Login7Payload {
     return buffer.data;
   }
 
+  CreateFedAuthExtData(buffer, fedAuthInfo) {
```
arthurschreiber: What do you think about changing this to `createFedAuthExtData ` to match the method naming convention?

src/connection.js
arthurschreiber: Yeah, we should catch this and require it to be a `boolean` if it is set. As this change is orthogonal to active directory auth, could you do this in a separate pull request?

src/prelogin-payload.js
```diff
@@ -195,6 +210,10 @@ module.exports = class PreloginPayload {
     this.marsString = marsByValue[this.mars];
   }
 
+  extractFedAuth(offset) {
+    return this.fedAuthRequired = this.data.readUInt8(offset);
+  }
+
   toString(indent) {
     indent || (indent = '');
     return indent + 'PreLogin - ' + sprintf('version:%d.%d.%d.%d %d, encryption:0x%02X(%s), instopt:0x%02X, threadId:0x%08X, mars:0x%02X(%s)', this.version.major, this.version.minor, this.version.patch, this.version.trivial, this.version.subbuild, this.encryption ? this.encryption : 0, this.encryptionString ? this.encryptionString : 0, this.instance ? this.instance : 0, this.threadId ? this.threadId : 0, this.mars ? this.mars : 0, this.marsString ? this.marsString : 0);
```
v-suhame: Can you add `fedAuthRequired` to the debug payload data here?

src/connection.js
Hadis-Fard: @v-suhame @arthurschreiber sent  PR #692 for checking the boolean config types

src/prelogin-payload.js
Hadis-Fard: @v-suhame if fedauthreq is true, it spreads the first array which contains call to fedauth, otherwise the empty array, which nothing :metal:, as @arthurschreiber said it is convoluted, and I'll make the changes

src/token/feature-ext-parser.js
Hadis-Fard: good idea, I play with it to see what i can do

Hadis-Fard:
@v-suhame @arthurschreiber Hey guys, I fixed the issues you caught, would you please review the changes I made?

- Added separate file for fedauth info parser
- flattened the nested callbacks by reading buffer at once and parsing it using buffer operator.
- Added fedAuthRequired to the debug payload data.
- Fixed the error messages
- Used push operator to add createfedAuth () to array

src/connection.js
```diff
@@ -865,7 +921,18 @@ class Connection extends EventEmitter {
     this.debug.payload(function() {
       return preloginPayload.toString('  ');
     });
-
+    if (this.fedAuthInfo.method != undefined) {
+      if (0 != preloginPayload.fedAuthRequired && 1 != preloginPayload.fedAuthRequired) {
+        this.emit('connect', ConnectionError(`Server sent an unexpected response for federated authentication value during negotiation. Value was  ${preloginPayload.fedAuthRequired}`, 'EFEDAUTH'));
+        return this.close();
```
v-suhame: Can you fix the double space in error message? `Value was  __ ${preloginPayload.fedAuthRequired}`  :stuck_out_tongue_winking_eye:

src/login7-payload.js
```diff
@@ -222,6 +272,16 @@ module.exports = class Login7Payload {
     return variableData.offset += value.length * 2;
   }
 
+  addVariableDataInt(variableData, value) {
+    value || (value = '');
+    variableData.offsetsAndLengths.writeUInt16LE(variableData.offset);
+    variableData.offsetsAndLengths.writeUInt16LE(4);
+    //placeHolder
+    this.tempHolder = variableData.data.getPos();
```
v-suhame: Can you change `tempHolder` into a more meaningful variable name? Like tempHolder of what?

Hadis-Fard: Now that i look at this again, i guess it can be better designed, it doesn't make sense that addVariableDataInt write the placeholder in this, it should return it or there should be another func..

src/connection.js
```diff
@@ -929,6 +997,19 @@ class Connection extends EventEmitter {
     process.nextTick(boundTransitionTo, this.STATE.SENT_NTLM_RESPONSE);
   }
 
+  sendFedAuthResponsePacket() {
+    const accessTokenLen = Buffer.byteLength(this.fedAuthInfo.token.accessToken, 'ucs2');
+    const data = new WritableTrackingBuffer(4 + accessTokenLen);
```
arthurschreiber: What do you think about using a regular `Buffer` instead of a `WritableTrackingBuffer` here? The length of the data is fixed, so `WritableTrackingBuffer` just adds bookkeeping overhead without providing any real advantages.

src/connection.js
```diff
@@ -929,6 +997,19 @@ class Connection extends EventEmitter {
...
+    data.writeUInt32LE(accessTokenLen + 4);
+    data.writeUInt32LE(accessTokenLen);
+    data.writeString(this.fedAuthInfo.token.accessToken, 'ucs2');
+    this.messageIo.sendMessage(TYPE.FEDAUTH_TOKEN, data.data);
+
+    const boundTransitionTo = this.transitionTo.bind(this);
+    // sent the fedAuth token message, the rest is similar to standard login 7
+    process.nextTick(boundTransitionTo, this.STATE.SENT_LOGIN7_WITH_STANDARD_LOGIN);
```
arthurschreiber: What do you think about using an arrow function here, instead of using `Function.prototype.bind`?

```js
process.nextTick(() => {
  this.transitionTo(this.STATE.SENT_LOGIN7_WITH_STANDARD_LOGIN);
});
```

src/token/fedauth-info-parser.js
```diff
@@ -0,0 +1,62 @@
+const async = require('async');
+
+var readFedAuthInfoOpt = (data, infoData, callback) => {
```
arthurschreiber: What do you think about just changing this to a regular function definition?

```js
function readFedAuthInfoOpt(data, infoData, callback) {
  ...
}
```

Arrow functions are usually only preferred when we want to have a function bound to the surrounding `this` scope, or we want a shorter way of defining a closure, neither of which applies here. 😅

src/token/fedauth-info-parser.js
```diff
@@ -0,0 +1,62 @@
...
+  const FEDAUTHINFOID = {
```
arthurschreiber: What do you think about pulling this constant to the top level scope, to ensure it does not get re-defined on each function call?

src/token/fedauth-info-parser.js
```diff
@@ -0,0 +1,62 @@
...
+      infoData.stsurlLen = fedAuthInfoDataLen;
+      infoData.stsurlOffset = fedAuthInfoDataOffset;
+      break;
+              // ignoring unknown fedauthinfo options
+    default:
+      break;
+  }
+
+  if (infoData.spnLen && infoData.stsurlLen) {
+    const spn = data.slice(infoData.spnOffset, infoData.spnOffset + infoData.spnLen);
```
arthurschreiber: Instead of slicing out a new `Buffer`, and then converting the `Buffer` to a string via `Buffer.prototype.toString`, I think we could use `Buffer.prototype.toString` directly. Also, I think by specifying the correct encoding (`ucs2`), we could get rid of the `String.prototype.replace` call to filter out the null bytes:

```js
const spn = data.toString('ucs2', infoData.spnOffset, infoData.spnOffset + infoData.spnLen);
const stsurl = data.toString('ucs2', infoData.stsurlOffset, infoData.stsurlOffset + infoData.stsurlLen);
```

src/connection.js
```diff
@@ -478,6 +516,23 @@ class Connection extends EventEmitter {
       this.emit('charsetChange', token.newValue);
     });
 
+    this.tokenStreamParser.on('fedAuthInfo', (token) => {
+      const clientId = '7f98cb04-cd1e-40df-9140-3bf7e2cea4db';
```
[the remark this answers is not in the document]
Hadis-Fard: With this id Azure Active Direcotry would know what's getting connected to it, it's registered within Azure. :)

src/connection.js
```diff
@@ -865,7 +921,18 @@ class Connection extends EventEmitter {
     this.debug.payload(function() {
       return preloginPayload.toString('  ');
     });
-
+    if (this.fedAuthInfo.method != undefined) {
+      if (0 != preloginPayload.fedAuthRequired && 1 != preloginPayload.fedAuthRequired) {
```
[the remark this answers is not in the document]
Hadis-Fard: 🤔  if we move it to preloginpayload, we won't have access to the connection class, and can't access Connection error or close. the `processPreLoginResponse` gets the feauth prelogin response and processes it similar to `encryptionString`

src/connection.js
```diff
@@ -865,7 +921,18 @@ class Connection extends EventEmitter {
...
+        this.emit('connect', ConnectionError(`Server sent an unexpected response for federated authentication value during negotiation. Value was  ${preloginPayload.fedAuthRequired}`, 'EFEDAUTH'));
+        return this.close();
+      }
+      // fedAuthRequired is used for capability negotiation when choosing between SSPI and federated authentication
+      if (preloginPayload.fedAuthRequired === 1 && this.config.domain) {
```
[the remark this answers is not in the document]
Hadis-Fard: right on :ok_hand: thanks for catching this

Hadis-Fard:
❤️  the detailed reviews! so here is the summary of update:

- removed extra spaces
- used buffer instead of writable tracking buffer
- pushed the constants to the outer scope
- used arrow function instead of bind(this) for state transitioning
- restricted fedauth method to `udefined`
- removed the error to let the server decide on authentication mechanism

Hadis-Fard:
commitlint is not happy about all previous commit messages here, opened #707

PR #664, opened 2017-12-19
Documentation for BulkLoad options.

v-suhame:
Adding docs to `gh-page` till we migrate to documentation.js


arthurschreiber:
Cool, this is great! What do you think about specifying the corresponding Transact-SQL Keyword, so users can look up additional information?

v-suhame:
Done 👍  that would be more helpful.

arthurschreiber:
@v-suhame Thank you! ❤️

PR #674, opened 2018-01-07
Stricter Data Type checks

arthurschreiber:
This pull request tightens up the type checks in the `writeParameterData` methods across all data types that implement this method.

I tried to keep backwards compatibility where possible, as I don't want this to be a breaking change. E.g. this still allows passing in strings to types like `TinyInt`, `SmallInt`, `Int`, `Money`, `Float` and so on. I'm not super happy about doing type conversion, as it can be lossy and lead to incorrect data being stored. But as changing this will be a breaking change for many users, I want to punt on that for now.

I did not try to keep compatibility with object-to-value conversion (via `.valueOf` methods), as this was never an explicitly supported feature and I doubt anyone really relied on this.

This change also tries to fix some cases where `tedious` was causing a loss of precision, especially with types like `Float` and `Real`, where passing a `number` type would always cause a conversion to a `string` and then back to a `number` via `parseFloat`. 😞

This pull request also does not try to fix the schism between data type handling of plain `Request` and `BulkLoad`. Currently, the difference between those two in the way parameters are handled can lead to different results. This is really crazy, but changing this could also potentially be backwards incompatible.

@v-suhame @chdh @atsage What do you think about these changes?

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1117,12 +1117,19 @@ const TYPE = module.exports.TYPE = {
     },
 
     writeParameterData: function(buffer, parameter) {
-      if (parameter.value != null) {
-        buffer.writeUInt16LE(parameter.length);
-        buffer.writeBuffer(parameter.value.slice(0, Math.min(parameter.length, this.maximumLength)));
-      } else {
+      const value = parameter.value;
+
+      if (value === undefined || value === null) {
         buffer.writeUInt16LE(NULL);
+        return;
       }
+
+      if (!(value instanceof Buffer)) {
+        throw new TypeError(`parameter.value must be a Buffer, undefined or null. Received type ${typeof value}`);
+      }
+
+      buffer.writeUInt16LE(parameter.length);
+      buffer.writeBuffer(value.slice(0, Math.min(parameter.length, this.maximumLength)));
     },
 
     validate: function(value) {
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -472,31 +472,38 @@ const TYPE = module.exports.TYPE = {
     },
 
     writeParameterData: function(buffer, parameter) {
-      if (parameter.value != null) {
-        const sign = parameter.value < 0 ? 0 : 1;
-        const value = Math.round(Math.abs(parameter.value * Math.pow(10, parameter.scale)));
-        if (parameter.precision <= 9) {
-          buffer.writeUInt8(5);
-          buffer.writeUInt8(sign);
-          buffer.writeUInt32LE(value);
-        } else if (parameter.precision <= 19) {
-          buffer.writeUInt8(9);
-          buffer.writeUInt8(sign);
-          buffer.writeUInt64LE(value);
-        } else if (parameter.precision <= 28) {
-          buffer.writeUInt8(13);
-          buffer.writeUInt8(sign);
-          buffer.writeUInt64LE(value);
-          buffer.writeUInt32LE(0x00000000);
-        } else {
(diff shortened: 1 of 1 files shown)


src/data-type.js
```diff
@@ -29,11 +29,18 @@ const TYPE = module.exports.TYPE = {
...
-      } else {
+      const value = parameter.value;
+
+      if (value === undefined || value === null) {
         buffer.writeUInt8(0);
+        return;
+      }
+
+      buffer.writeUInt8(1);
+      if (typeof value === 'number') {
```
chdh: Why not use the same program logic in validate(), i.e. test whether the value is already a number before using parseInt()?

src/data-type.js
```diff
@@ -174,24 +198,30 @@ const TYPE = module.exports.TYPE = {
...
+
+      if (value === null || value === undefined) {
+        buffer.writeUInt8(0);
+        return;
+      }
 
-        buffer.writeUInt8(4);
-        buffer.writeUInt16LE(days);
+      if (!(value instanceof Date)) {
+        throw new TypeError(`parameter.value must be a Date, undefined or null. Received type ${typeof value}`);
```
chdh: This test should also be made in `validate()`, because `Request.validateParameters()` calls `validate()` and generates a better error message if the exception occurs there.
If the error occurs later in the constructor of `RpcRequestPayload`, the error message would not indicate the parameter name.

src/data-type.js
```diff
@@ -414,31 +472,38 @@ const TYPE = module.exports.TYPE = {
...
+        buffer.writeUInt8(13);
+        buffer.writeUInt8(sign);
+        buffer.writeUInt64LE(decimalValue);
+        buffer.writeUInt32LE(0x00000000);
+      } else {
+        buffer.writeUInt8(17);
+        buffer.writeUInt8(sign);
+        buffer.writeUInt64LE(decimalValue);
+        buffer.writeUInt32LE(0x00000000);
+        buffer.writeUInt32LE(0x00000000);
```
chdh: There is still an unnecessary `parseFloat()` in the corresponding `validate()` function below.

src/data-type.js
```diff
@@ -137,11 +154,18 @@ const TYPE = module.exports.TYPE = {
...
-      } else {
+      const value = parameter.value;
+
+      if (value === undefined || value === null) {
         buffer.writeUInt8(0);
+        return;
+      }
+
+      buffer.writeUInt8(4);
+      if (typeof value === 'number') {
```
chdh: If it's a request parameter and the value is the number 0.000000006, the `parseInt()` call in the corresponding `validate()` function below will still convert it to 6, because in `Request.validateParameters()` the return value of validate() is used to replace the original parameter value.

chdh:
I would split the program logic for each data type into the following 3 functions:

- `convert`: Convert the value if necessary and return the converted value (or the original input value if no conversion is necessary).
- `validate`: Check whether the passed value is valid for the data type, without any conversion. Return a boolean that indicates whether the value is valid or not. The error message is better composed in the caller.
- `write`: Write the value to the buffer, without any conversion or validation.

Usage:

- In `validateParameters()`: Call `convert()` and `validate()`. (Replace the parameter value with the converted value, like it's been done so far).
- In `RpcRequestPayload:` Call `write()`.
- In `BulkLoad`: Call `validate()` and `write()`.

Add an "autoconvert" option that can be used to skip the `convert()` call in `validateParameters()`.

arthurschreiber:
Closing this in favor of #677. 😄

PR #676, opened 2018-01-10
refactor: Split data type definitions into separate files.

arthurschreiber:
`src/data-types.js` is one of the largest files we have in tedious, containing all data type definitions and related code. Extracting this into separate files per data type will make this a bit easier to handle. 😅


v-suhame:
That's a good move 😌

PR #677, opened 2018-01-10
feature: Improved Data Type validation/conversion.

arthurschreiber:
This refactors the data type logic in tedious, making it more strict and more accurate.

If the parameter specified by the user can not be converted to the data type of the parameter without a loss of precision, the request will not be executed and an error will be passed to the request's callback instead.

The same parameter value conversion logic is now also used for both regular requests as well as for bulk loads, fixing a long standing issue where storing the same data via these two methods would result either result in a failure or different data being stored in the database.

---

This is dependent upon https://github.com/tediousjs/tedious/pull/676 being merged first.


arthurschreiber:
@v-suhame @chdh @atsage @Hadis-Fard Would you mind taking a look at this?

src/data-types/timen.js
```diff
@@ -77,17 +77,22 @@ module.exports = {
...
+    if (value === undefined || value === null) {
       return null;
     }
+
+    let dateValue;
     if (value instanceof Date) {
-      return value;
+      dateValue = value;
+    } else {
+      dateValue = new Date(Date.parse(value));
```
v-suhame: Need to remove this redundant conversion to [Date](https://github.com/tediousjs/tedious/pull/677/files#diff-76524e539b02f72f0abdc709930eb753R46) too 😄

arthurschreiber: Yeah. I'm just keeping it because the old code allowed passing in strings as values for `Time`. We can remove this in a breaking change later. 👍

v-suhame: I brought that up because [datetime2n](https://github.com/tediousjs/tedious/pull/677/files#diff-1ba56de971ccce1d484f6cb0debb4f72L49) and  [datetimeoffsetn](https://github.com/tediousjs/tedious/pull/677/files#diff-5f9fdee646ff4bda759de09fc175797fL43) had conversion removed. Shouldn't the same behaviour apply for all 3 datatype?

src/data-types/money.js
```diff
@@ -23,14 +23,22 @@ module.exports = {
...
+    let numberValue;
+    if (typeof value === 'number') {
+      numberValue = value;
+    } else {
+      numberValue = parseFloat(value);
     }
-    return value;
+
+    if (!Number.isFinite(numberValue) || (typeof value === 'string' && value !== numberValue.toString()) || numberValue.toString() !== numberValue.toFixed(4)) {
+      return new TypeError(`The given value could not be converted to ${this.name}`);
```
v-suhame: How about removing or having an alternative for `numberValue.toString() !== numberValue.toFixed(4)`? 

This condition is too strict, if input has scale < 4 it will throw exception, but driver should allow scale <= 4

v-suhame: Can you also add a test for money/smallmoney type with scale < 4? All current tests are for scale 4.

src/data-types/nvarchar.js
```diff
@@ -69,16 +69,16 @@ module.exports = {
     }
   },
 
-  validate: function(value) {
-    if (value == null) {
+  validate(value, length, precision, scale) {
+    if (value === undefined || value === null) {
```
v-suhame: Is precision & scale here by mistake?

src/data-types/real.js
```diff
@@ -17,20 +17,28 @@ module.exports = {
...
+    let numberValue;
+    if (typeof value === 'number') {
+      numberValue = value;
+    } else {
+      numberValue = parseFloat(value);
     }
-    return value;
+
+    if (!Number.isFinite(numberValue) || (typeof value === 'string' && value !== numberValue.toString()) || numberValue !== Math.fround(numberValue)) {
+      return new TypeError(`The given value could not be converted to ${this.name}`);
```
v-suhame: Wouldn't using Math.fround() be a breaking change? 🤔 I'm afraid that most of the data passed would fail at this point, and hinder the usage of `real` type.

src/data-types/timen.js
arthurschreiber: Yeah, unfortunately I think that there might be users out there that use `time` and pass in timestrings. I'm not 100% sure, and I'd rather not break this right now, but sometime in the future. 😅

src/data-types/money.js
arthurschreiber: You're right, I missed handling of numbers that have less than 4 numbers after the decimal point. 😓

src/data-types/nvarchar.js
arthurschreiber: Yeah, will drop those! 👍

src/data-types/real.js
arthurschreiber: Yeah, this one's a bit harder to decide.

The loss of precision is a huge issue tho, because most users will not know that `real` is limited and won't save the exact value they're trying to send, which can lead to huge issues because the data in the database can no longer be trusted. That's why I think raising an error here is better than before and fixes a bug. What do you think? 😄

v-suhame: I agree to raising error here ✌️  I was just wondering how much of an impact it would have on existing user(i.e. will it be a breaking change), but as you pointed, users must have already run into precision loss, so it better to have this in sooner rather than later.

src/data-types/timen.js
v-suhame: 😅 I think I should elaborated a bit more.  If users are passing `time` in timestrings format, then it is likely they will do the same for `datetime2` and `datetimeoffset`, as they all have nano second precision.

Code fragment from timen.js and datetimeoffsetn.js looks like,

```node
validate(value) {
  ...
  dateValue = new Date(Date.parse(value));
  ...
}

writeParameterData: function(..){
  ..
  if (parameter.value != null) {
    const time = new Date(+parameter.value); // datetime2n.js doesn't have this conversion
    ...
}
```
We should add the conversion(pointed in the code fragment) back into [datetime2n.js](https://github.com/tediousjs/tedious/pull/677/files#diff-1ba56de971ccce1d484f6cb0debb4f72L49) to allow string passed into `datetime2`.

src/data-types/nvarchar.js
arthurschreiber: Done in f9b49e1.

src/data-types/timen.js
arthurschreiber: Ah, yes, you're totally right! ❤️ Fixed this in 0d4e197ad58566b9e50c837f61c88aa49a3ffa2a.

arthurschreiber: i think all 3 types now use `Date.parse` in the `validate` method to allow conversion from strings, and they no longer perform redundant conversion inside the `writeParameterData` method.

src/data-types/varchar.js
```diff
@@ -69,16 +69,16 @@ module.exports = {
...
+  validate(value, length, precision, scale) {
+    if (value === undefined || value === null) {
       return null;
     }
-    if (typeof value !== 'string') {
-      if (typeof value.toString !== 'function') {
-        return TypeError('Invalid string.');
-      }
-      value = value.toString();
+
```
v-suhame: A minor correction: remove precision, scale

src/request.js
```diff
@@ -175,11 +175,13 @@ module.exports = class Request extends EventEmitter {
   validateParameters() {
     for (let i = 0, len = this.parameters.length; i < len; i++) {
       const parameter = this.parameters[i];
-      const value = parameter.type.validate(parameter.value);
+      const value = parameter.type.validate(parameter.value, parameter.length, parameter.precision, parameter.scale);
```
v-suhame: Precision and scale are never actually used the validate(), let's remove them.

arthurschreiber: Yeah, that's a good point. They *should* be used for `numeric` and `decimal` (and maybe for some of the date types?). I'll take another look! 👍

src/data-types/varchar.js
arthurschreiber: Good idea! 👍

src/request.js
arthurschreiber: Done in 0968cb4.

arthurschreiber:
Heya @tediousjs/microsoft-contributors 👋

how do you feel about this changeset? Should we declare it as a breaking change and merge and release it together with other breaking changes as `v3.0`?

v-suhame:
+1 to releasing this along with other breaking changes. This PR is good to go once https://github.com/tediousjs/tedious/pull/677#pullrequestreview-89294724 is handled 😄

src/bulk-load.js
```diff
@@ -129,12 +129,15 @@ module.exports = class BulkLoad extends EventEmitter {
...
       const c = this.columns[i];
-      c.type.writeParameterData(this.rowsData, {
+
+      const parameter = {
         length: c.length,
         scale: c.scale,
         precision: c.precision,
-        value: row[arr ? i : c.objName]
-      }, this.options);
+        value: c.type.validate(row[arr ? i : c.objName], c.length, c.precision, c.scale)
```
v-suhame: `validate()` doesn't throw error but it returns error if validation fails. Can you add check to verify type of  `validate()`'s return value before passing it to `writeParameterData`?

MichaelSun90:
@arthurschreiber, I think this PR may be outdated since the recent merged pip related changes? If you can confirm that, then I think can close this PR for now.

src/data-types/money.js
IanChokS: `numberValue.toString().split('.')[1].length <= 4`

src/data-types/tvp.js
```diff
@@ -43,36 +43,48 @@ module.exports = {
...
+          if (column === undefined || column.type === undefined) {
+            return new TypeError();
+          }
+
+          return {
+            value: column.type.validate(columnValue, column.length, column.scale, column.precision),
+            length: column.length,
+            scale: column.scale,
+            precision: column.precision
+          };
```
IanChokS: Returning an object causes data-type value inputs to vary: if using TVP, data-type inputs would be this object, whereas anything else would be a single value. To avoid this two variations of data-type inputs, this TVP `.validate()` method should only return the single value, such as `return column.type.validate(columnValue, column.length, column.scale, column.precision)`

PR #680, opened 2018-01-11
documentation: Applying theme to GitHub  pages

v-suhame:
Imported theme to GitHub pages in attempt to make it more visually appealing, it will look like https://v-suhame.github.io/tedious/ and it also make the site responsive.

I have couple of questions/suggestions:
- How about adding the sample from [/examples](https://github.com/tediousjs/tedious/tree/master/examples) into the GitHub pages? That will make it easier to find, we have one line in the documentation pointing to it, but its easy to overlook it when someone tries to ramp-up. 
- ChangeLog in GitHub pages hasn't been updated since long, can we somehow archive it as all the changes are already in https://github.com/tediousjs/tedious/releases


arthurschreiber:
@v-suhame Thank you for preparing this! ❤️ 

Unfortunately, I'm not super convinced about the result. Not that the existing theme is great, but the Jekyll theme does simply not feel like an improvement to me. 😞

v-suhame:
Ah okay! Did you have any another suggestions in mind to improve the UI? 
Also, any thoughts on the other 2 points (in the comment above)?

arthurschreiber:
> ChangeLog in GitHub pages hasn't been updated since long, can we somehow archive it as all the changes are already in https://github.com/tediousjs/tedious/releases

I think we could back-fill the changelogs from the website to https://github.com/tediousjs/tedious/releases, and then just make the changelog link point to the github release page?

arthurschreiber:
> How about adding the sample from /examples into the GitHub pages? That will make it easier to find, we have one line in the documentation pointing to it, but its easy to overlook it when someone tries to ramp-up.

I think that would be a great idea, I'm just not sure how easy it would be to keep those in sync. If it requires manual intervention, the risk of these two falling out of sync will be pretty high. Maybe we could just add a link from the website to the examples directory here on GitHub, as a stop gap solution? 🤔

v-suhame:
>I think we could back-fill the changelogs from the website to https://github.com/tediousjs/tedious/releases, and then just make the changelog link point to the github release page?

GitHub release page already has all the changelogs since the beginning of time 😄  we can just link  changelog to the github release page, as you pointed.

>I think that would be a great idea, I'm just not sure how easy it would be to keep those in sync. If it requires manual intervention, the risk of these two falling out of sync will be pretty high. Maybe we could just add a link from the website to the examples directory here on GitHub, as a stop gap solution? 🤔

/examples hasn't changed since 4 years 😆  but you're right, there is possibility for them going out of sync. We can create a menu item `Examples` and link it to the examples directory in github, same soltion as changelog ✌️

arthurschreiber:
Cool, that sounds great! Do you want to prepare a pull request for these changes? If not, I'll try to get to it sometime soon! 👍

PR #681, opened 2018-01-11
fixed package deprecation warning messages

Hadis-Fard:
Chores:

- replaced prepublish with [prepare](https://docs.npmjs.com/misc/scripts#deprecation-note)

- Updated Semantic release package version to 11, as version 6 is outdated and might installs deprecated npmconf and nodeuuid packages


arthurschreiber:
Hey @Hadis-Fard 👋

Thank you for preparing these changes and opening a pull request.

Regarding the change from `prepublish` to `prepare`, I don't think we can include this change. Node v4 is still in LTS, and it originally shipped with `npm@2.14.2`, and even the latest v4 release  only includes a version of `npm@2.x`. Even Node v6.9.0, which shipped with `npm@3.10.8`, does not understand `prepare`. 😞 It looks like we'll have to wait for quite some time before we can safely migrate from `prepublish` to `prepare`.

Regarding the bump of `semantic-release`: Yay, thank you so much! 😄 If you extract this into a separate pull request or revert the `prepublish` change, I can merge this.

On a separate note: `semantic-release` depends on commits being made according to the [Angular Commit Message Conventions](https://gist.github.com/stephenparish/9941e89d80e2bc58a153). For one-off contributions I usually just squash all commits in a pr and fix the commit message for the pull request myself, but if you plan on submitting more contributions in the future, it'd be ✨ if you could try and follow the convention.

Again, thank you very much for these changes! ❤️

Issue #659, opened 2017-12-12
unhandled error seperately when connecting with timeout


realshaft:
I do have an issue with understanding this behaviour. Even if I listen in 'connected' to the error and cancel the connection, after a while (a minute or so, independent of the server timeout of 15s) the error event gets also called.

Qs: 
- If I cancel the connection in 'connected', why is there still being a error sent?
- Where does the second timeout comes from? when can I get sure that all asnc operations are done in tedious?

```
var connection = new Connection(config);

// Attempt to connect and execute queries if connection goes through
connection.on('connect', function(err) {
      if (err) {
        context.app.error(`Connection to ${xxx} not successful: ${err}`); // <--- Is executed so the error is handled
        connection.cancel();
      } else {
        context.app.info('Connected to ' + context.app.get('xxx').server);
});

// But after another timeout(?) the error event has been dispatched as well
connection.on('error', err => {
      context.app.error(`Connection to ${context.app.get('xxx').server} not successful: ${err}`);
});
```

v-suhame:
Can you share the error you are getting in `connect` and `error` callback? 

Connection is usually closed after  `error` event is emitted, and call to `cancel` is executed only if there is a `request` pending. So, calling `connection.cancel();` while checking for error in `connect` event would return false.

realshaft:
With the above code I do get this when the database is not reachable

in connection.on('connect'):
```
2017-12-13T13:35:20.753Z - error: Connection to 10.110.0.2 not successful: ConnectionError: Failed to connect to 10.110.0.2:11433 in 15000ms
2017-12-13T13:35:20.754Z - info: Connection to 10.110.0.2 closed
```
And as you can see after more than a minute I do get in connection.on('error') event as well:
```
2017-12-13T13:36:23.710Z - error: Connection to 10.110.0.2 not successful: ConnectionError: Connection lost - connect ETIMEDOUT 10.110.0.2:11433
```

arthurschreiber:
This was fixed via #682. 👍

kuseni:
I think this issue appears again in 2.6.4.

Code:
```javascript
log('started');

var conn = new Connection({
	server:   '8.8.8.8',
	userName: 'test',
	password: 'test'});

conn.on('connect', (err) => {
	log(`connect(): err="${err}"`);
	conn.close();
});

conn.on('end',    ()   => { log(`end()`); });
conn.on('error', (err) => { log(`error(): err="${err}"`); });
[...]

PR #682, opened 2018-01-13
Fix Issue #659

Hadis-Fard:
issue #659 : socket is assigned and attempted to connect after the connection has transitioned to final state by connect timeout.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -689,8 +689,8 @@ class Connection extends EventEmitter {
       if (err) {
         return this.socketError(err);
       }
-      if(this.state === this.STATE.FINAL) {
-        if(socket) {
+      if (this.state === this.STATE.FINAL) {
+        if (socket) {
           socket.destroy();
         }
       } else {
--- a/src/connection.js
+++ b/src/connection.js
@@ -693,18 +693,18 @@ class Connection extends EventEmitter {
       if (this.state === this.STATE.FINAL) {
         socket.destroy();
         return;
-      } else {
-        this.socket = socket;
-        this.socket.on('error', this.socketError);
-        this.socket.on('close', this.socketClose);
-        this.socket.on('end', this.socketEnd);
-        this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-        this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
-        this.messageIo.on('message', () => { this.dispatchEvent('message'); });
-        this.messageIo.on('secure', this.emit.bind(this, 'secure'));
-
-        this.socketConnect();
       }
+
+      this.socket = socket;
+      this.socket.on('error', this.socketError);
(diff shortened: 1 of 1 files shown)


src/connection.js
```diff
@@ -689,17 +689,22 @@ class Connection extends EventEmitter {
       if (err) {
         return this.socketError(err);
       }
+      if (this.state === this.STATE.FINAL) {
+        if (socket) {
```
arthurschreiber: I'm pretty sure `socket` is always set here.

Can we change the code to do an early return instead?

```js
if (this.state === this.STATE.FINAL) {
  socket.destroy();
  return;
}

this.socket = socket;
...
```

Also, could you add an empty newline before the `if` statement here?

arthurschreiber:
This is great, thank you very much! I left a tiny comment regarding the code style, other than that, this is looking great! 👍

Can we also add an integration test for this, somehow?

src/connection.js
```diff
@@ -690,16 +690,21 @@ class Connection extends EventEmitter {
...
-      this.socket.on('close', this.socketClose);
-      this.socket.on('end', this.socketEnd);
-      this.messageIo = new MessageIO(this.socket, this.config.options.packetSize, this.debug);
-      this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
-      this.messageIo.on('message', () => { this.dispatchEvent('message'); });
-      this.messageIo.on('secure', this.emit.bind(this, 'secure'));
-
-      this.socketConnect();
+      if (this.state === this.STATE.FINAL) {
+          socket.destroy();
```
arthurschreiber: The whitespace is still a bit off here. 😅

Hadis-Fard:
@arthurschreiber the pause-resume test has failed on Appveyor build. The test is flaky and fails when I test on a remote server, there's also a discussion about it in [518](https://github.com/tediousjs/tedious/pull/518#issuecomment-342658545), could you please fix it?

arthurschreiber:
Perfect! Thanks for following up with the style changes. 😍 I'll merge and release this ASAP!

arthurschreiber:
This was just released as part of `tedious@2.2.4` (currently still tagged as `next`).

pbarbiero:
@arthurschreiber is there any chance we can get this back-ported to the 1.x branch for those of us stuck for the time being?  We've started having servers die due to this recently, and we are not in a position to upgrade to v2 currently in some cases.

arthurschreiber:
If I remember correctly, the upgrade to v2 should be pretty painless. The only real change was dropping support for node versions earlier to 4.

If possible I'd like to avoid spending time on backporting changes, but if you absolutely can not upgrade, I can go and do it. Please let me know! 😄

pbarbiero:
Thanks Arthur,

If thats the main churn to v2, I can probably force it through our release process. I don't want to burden you with more releases in light of that.

I appreciate the response!

PR #684, opened 2018-01-16
chore: update semantic-release version

Hadis-Fard:
Version 6 installs deprecated packages, updated to fix warning messages on install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ jobs:
       before_script: skip
       script: npm run lint
     - stage: release
-      node_js: 4
+      node_js: 8
       before_install: skip
       before_script: skip
-      script: npm run semantic-release || true
+      script: npm run semantic-release
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "mitm": "^1.3.2",
     "nodeunit": "^0.11.0",
     "rimraf": "^2.6.1",
-    "semantic-release": "^6.3.2",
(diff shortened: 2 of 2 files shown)


arthurschreiber:
@Hadis-Fard Thank you so much! 😍 

Could you bump this to the latest `semantic-release` version? Right now, this would be `12.2.2`.

Also, there have been a few breaking changes between the version of `semantic-release` we're currently using and the latest version released. From going through their changelog, it looks like we'll also need to change this line https://github.com/tediousjs/tedious/blob/f5a0f95a5a522ad6a312170201b829e70c866976/package.json#L74 to
```
"semantic-release": "semantic-release"
```

We'll also be able to update our travis configuration on this line https://github.com/tediousjs/tedious/blob/f5a0f95a5a522ad6a312170201b829e70c866976/.travis.yml#L62 to
```
script: npm run semantic-release
```

Could you include these changes in this pull request, as merging this pull request without these changes will break the semantic-release configuration. 😅

arthurschreiber:
Also, I think this will require upgrading the `release` stage configuration in the travis config to run with node v8 instead of v4. 😅

Hadis-Fard:
aaa thanks for checking the release notes, done and done 😸

arthurschreiber:
Sweet! 😍 I'll try to get this merged this weekend.

arthurschreiber:
Thank you @Hadis-Fard! I squashed the changes together before merging.

PR #687, opened 2018-01-20
test: Add a test to verify no errors are thrown after connect timeout.

arthurschreiber:
This adds a test case for #682.


arthurschreiber:
@Hadis-Fard This test fails without the code change made in #682.

PR #691, opened 2018-01-23
 feature: Deprecate data type aliases.

arthurschreiber:
This deprecates data type aliases  `DateN`, `DateTime2N`, `DateTimeOffsetN`, `TimeN` and `UniqueIdentifierN` in favor of their 'proper' names.

* `DateN` -> `Date`
* `DateTime2N` -> `DateTime2`
* `DateTimeOffsetN` -> `DateTimeOffset`
* `TimeN` -> `Time`
* `UniqueIdentifierN` -> `UniqueIdentifier`

The deprecated aliases will be removed in the next major version, together with types that shouldn't have been exposed in the first place (`IntN`, `BitN`, `DateTimeN`, etc.).

### DIFF
--- a/src/data-type.js
+++ b/src/data-type.js
@@ -1,3 +1,5 @@
+const deprecate = require('depd')('tedious');
+
 const Null = require('./data-types/null');
 const TinyInt = require('./data-types/tinyint');
 const Bit = require('./data-types/bit');
@@ -14,7 +16,7 @@ const SmallMoney = require('./data-types/smallmoney');
 const BigInt = require('./data-types/bigint');
 const Image = require('./data-types/image');
 const Text = require('./data-types/text');
-const UniqueIdentifierN = require('./data-types/uniqueidentifiern');
+const UniqueIdentifier = require('./data-types/uniqueidentifier');
 const IntN = require('./data-types/intn');
 const NText = require('./data-types/ntext');
 const BitN = require('./data-types/bitn');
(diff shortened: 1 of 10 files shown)


arthurschreiber:
Some more information around this: I want to deprecate exposing TDS internals to our users. No one should have to know about the difference between "nullable" and "non-nullable" data types (e.g. `TinyInt`/`SmallInt`/`Int`/`BigInt` vs `IntN`).

They were not usable before, but I think they're exposed via column metadata and via `tedious.TYPES`. The first step here is to deprecate these "internal" data-types, and a later step will change the type in column metadata to never expose these internal types.

@v-suhame @Hadis-Fard What do you think? 😄

v-suhame:
I totally agree! I too was a bit confessed with these types when I was ramping up 😅 Having same aliases as SQL Server datatype would stop the confusion.

src/data-type.js
```diff
@@ -93,7 +95,13 @@ for (const id in TYPE) {
...
         typeByName[alias] = type;
+
+        deprecate.property(typeByName, alias, 'The `' + alias + '` data type alias is deprecated, please use `' + type.name + '` instead.');
       }
     }
   }
 }
+
+['IntN', 'BitN', 'FloatN', 'MoneyN', 'DateTimeN'].forEach(function(name) {
+  deprecate.property(typeByName, name, 'The `' + name + '` data type is internal and will be removed.');
```
v-suhame: DecimalN and NumericN should also be added for deprecation.

arthurschreiber: > DecimalN and NumericN should also be added for deprecation.

Ah, yes, I missed those! 😅 I added them both.

simoneb:
Came here after noticing deprecation warnings popping up in our code which uses package node-mssql and reading about the issue [here](https://github.com/patriksimek/node-mssql/issues/604).

I'm wondering why these warnings are logged, it seems like they're logged regardless of whether the deprecated data types are being used or not.

v-suhame:
@simoneb deprecation message are logged only if the corresponding type is used. `node-mssql` references the deprecated types internally, say if you're using `Int` type `node-mssql` references `IntN` too as pointed in https://github.com/patriksimek/node-mssql/issues/604#issuecomment-364166616.

As @arthurschreiber has explained, deprecating these aliases will make data types less confusing. It obviously mean all the packages depending on `tedious` need to get rid of those aliases, but it should be relatively straight forward to that. Sorry for the inconvenience 🙏

arthurschreiber:
Hey @simoneb (and everyone else who's coming to this from https://github.com/patriksimek/node-mssql/issues/604#issuecomment-364166616).

We especially added these deprecations so we can remove these aliases in the next major version of tedious, and users will have enough time upfront to adjust to these changes before. The alternative would've been removing these without an upfront deprecation period, and that does not seem like a viable alternative to me. 🤷‍♂️ 

Datatypes like `IntN` were never usable, and are an implementation detail internal to `tedious` and should never have been exposed in the first place (one of the reasons we're going to hide them in the future). For one reason or another, `node-mssql` is referencing these data types, which is leading to the deprecation warning being emitted.

If you do not care about deprecations, you can start node.js with `--no-deprecation` or set the `NO_DEPRECATION` environment variable to `tedious`, as described over at https://github.com/dougwilson/nodejs-depd. You should be aware that this means you are explicitly ignoring that you're using code which is definitely going to break in the future.

PR #692, opened 2018-01-24
BREAKING CHANGE: restricting boolean Config types to true and false

Hadis-Fard:
- Added type checking for boolean config option

- ~~Changed the default value of `enableCursorCloseOnCommit` from `undefined` to `false` to match the documents and the rest of boolean options~~

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -110,10 +110,7 @@ class Connection extends EventEmitter {
       }
 
       if (config.options.abortTransactionOnError != undefined) {
-        if (typeof config.options.abortTransactionOnError !== 'boolean') {
-          throw new TypeError('options.abortTransactionOnError must be a boolean (true or false).');
-        }
-
+        this.validateBooleanOptions(config.options.abortTransactionOnError, 'options.abortTransactionOnError');
         this.config.options.abortTransactionOnError = config.options.abortTransactionOnError;
       }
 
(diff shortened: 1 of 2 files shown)


src/connection.js
```diff
@@ -78,7 +78,7 @@ class Connection extends EventEmitter {
         enableAnsiWarnings: true,
         enableArithAbort: false,
         enableConcatNullYieldsNull: true,
-        enableCursorCloseOnCommit: undefined,
+        enableCursorCloseOnCommit: false,
```
v-suhame: You need to revert the changes made for `enableCursorCloseOnCommit` 😅 More details in #643 & #653

Hadis-Fard: :stuck_out_tongue_closed_eyes: Sure

arthurschreiber:
@Hadis-Fard These changes will be backwards incompatible, because we're changing how non-boolean values are treated.

Could you open a separate pull request that deprecated passing non-boolean values first, similar to #695?

Hadis-Fard:
TODO:
- expand the type errors to the other types as well as boolean types. similar to deprecations in #698
- check for null values in getinitialSql function.
- introducing the changes that are going to be deprecated in #698 

Will resume the work after 698 is merged.

src/connection.js
```diff
@@ -389,6 +366,12 @@ class Connection extends EventEmitter {
     };
   }
 
+  validateBooleanOptions(option, optionName) {
```
arthurschreiber: How do you feel about making this a "regular" function instead of a method on `Connection`? Basically, just like `deprecateNonBooleanConfigValue` on `master`.

arthurschreiber:
Closed in favor of #811.

PR #693, opened 2018-01-26
ci: lint commit messages

arthurschreiber:
I just messed up with when merging #691 - `feature:` is not a valid type that is recognized by `semantic-release`, so no new release was pushed.

I had to fix this up locally and force-push the `master` branch, which is really not a great situation to be in.

As this is the second time something like this happened, I looked around and found `commitlint`, which can be set up to run in Travis and will check that all commits contained in a pull request follow the a commit message style - in our case that is the "conventional" style.

As this runs in the `lint` stage on travis, unit and integration tests will always run before, and a botched commit message won't prevent tests from running.

`commitlint` can also be used to check commit messages *before* pushing them remotely, either via a git commit hook (see http://marionebl.github.io/commitlint/#/guides-local-setup), or by executing `commitlint` and specifying git revspecs for `--from` and `--to`.

```bash
node_modules/.bin/commitlint --from origin/master --to HEAD
```

☝️ This would check all commits made in my currently checked out branch that have not been merged into the remote `master` branch yet.


arthurschreiber:
The job log over at https://travis-ci.org/tediousjs/tedious/jobs/333638510 contains the output for the commits contained in this pull request, as an example.

Hadis-Fard:
❤️  to this, what do you think about adding "Guidelines for Creating Pull Requests" in Readme and put the conventions there as well?

arthurschreiber:
@Hadis-Fard That's a great idea! 😍

arthurschreiber:
@Hadis-Fard I opened #702 to not forget about adding contributor guidelines, and to allow this pull request to move forward. :thumbsup:

v-suhame:
@arthurschreiber Thanks for setting this up 😃

Issue #610, opened 2017-08-31
Cannot connect via encrypted connection to SQL Server 2008 SP2


eplowe:
My app is trying to connect to SQL Server 2008 SP2. The connections are configured to be encrypted. The connection always times out at the configured connectionTimeout. I've talked to the DBA, and he's seeing the following error:

"The prelogin packet used to open the connection is structurally invalid; the connection has been closed. Please contact the vendor of the client library. [CLIENT: *.*.*.*]" 

Any ideas?

slmmm:
Have you set your connection options for encrypted before attempting to connect?
**options: {
  encrypt: true
}**

eplowe:
I have. When it's false I get a fast fail telling me that the server is
configured only to accept encrypted connections.

eplowe:
I made a switch over to mssqlnodev8, it connects fine. But, this locks me to using windows boxes only.

v-suhame:
@eplowe try enabling the debug log in driver to get more details. If you need assistance, you can share the logs either in github or via email. 

```
options: {
    debug: {
      packet: true,
      data: true,
      payload: true,
      token: true,
      log: true
    }
}
```

eplowe:
connected to *.*.*.*:1433

Sent
  type:0x12(PRELOGIN), status:0x01(EOM), length:0x002F, spid:0x0000, packetId:0x
01, window:0x00
  0000  00001A00 06010020 00010200 21000103 00220004 04002600 01FF0000 00010001
 ....... ....!... ."....&. ........
  0020  01000000 000000  .......
  PreLogin - version:0.0.0.1 1, encryption:0x01(ON), instopt:0x00, threadId:0x00
000000, mars:0x00(OFF)
State change: Connecting -> SentPrelogin

Received
  type:0x04(TABULAR_RESULT), status:0x01(EOM), length:0x002B, spid:0x0000, packe
tId:0x01, window:0x00
  0000  00001A00 06010020 00010200 21000103 00220000 04002200 01FF0A00 0FA00000
 ....... ....!... ."....". ........
  0020  010000  ...
  PreLogin - version:10.0.15.160 0, encryption:0x01(ON), instopt:0x00, threadId:
0x00000000, mars:0x00(OFF)
State change: SentPrelogin -> SentTLSSSLNegotiation

Sent
  type:0x12(PRELOGIN), status:0x01(EOM), length:0x00FC, spid:0x0000, packetId:0x
01, window:0x00
[...]

PR #694, opened 2018-01-28
Default to `true` for `options.encrypt`

arthurschreiber:
This changes the default of `options.encrypt` from `false` to `true`, effectively switching on connection encryption unless explicitly turned off.

It also makes the type checking of the `options.encrypt` value stricter, only accepting `boolean` values.

Both of these changes are breaking backwards compatibility. The plan is to deprecate the current behaviour, and then later release these changes on a major version bump.

Fixes #606.


v-suhame:
On a side note, while updating the documentation, remember to update the requirements for older versions of SQL Server. 
For SQL Server <2008 SP4, users should have `options.cryptoCredentialsDetails.ciphers` in conenction config (from [issue#610](https://github.com/tediousjs/tedious/issues/610#issuecomment-339812682)) if `options.encrypt : true`.

arthurschreiber:
@v-suhame Thanks for reminding me! We probably should support this out-of-the box.

I read the documentation you linked to in #610. From what I understand, SQL Server supports ciphers based on the operating system it is running on. In the case of SQL Server 2008 running on Windows Server 2003, the list looks like this:

* `TLS_RSA_WITH_RC4_128_MD5`
* `TLS_RSA_WITH_RC4_128_SHA`
* `TLS_RSA_WITH_3DES_EDE_CBC_SHA`
* `TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA`
* `TLS_RSA_WITH_DES_CBC_SHA`
* `TLS_DHE_DSS_WITH_DES_CBC_SHA`
* `TLS_RSA_EXPORT1024_WITH_RC4_56_SHA`
* `TLS_RSA_EXPORT1024_WITH_DES_CBC_SHA`
* `TLS_DHE_DSS_EXPORT1024_WITH_DES_CBC_SHA`
* `TLS_RSA_EXPORT_WITH_RC4_40_MD5`
* `TLS_RSA_EXPORT_WITH_RC2_CBC_40_MD5`
* `TLS_RSA_WITH_NULL_MD5`
* `TLS_RSA_WITH_NULL_SHA`

None of these ciphers is accepted by newer Node versions, as their security is rather questionable.

According to https://wiki.mozilla.org/Security/Server_Side_TLS, the only cipher out of this list that might be used for backwards-compatibility without totally compromising security is `TLS_RSA_WITH_3DES_EDE_CBC_SHA`.

This specific suite can be enabled by using `DES-CBC3-SHA` in the `ciphers` option.

I think we can make tedious inform the user more thoroughly when the encryption handshake fails, and let them know that potentially the cipher suite configuration they're using is too strict? I'd rather not include that cipher by default (because of security implications), but the current behaviour is pretty opaque and unfortunate from a user experience perspective.

v-suhame:
>  I think we can make tedious inform the user more thoroughly when the encryption handshake fails, and let them know that potentially the cipher suite configuration they're using is too strict?

👍 to this. It will be more helpful than putting them on GitHub Pages.

Right now, there is no error thrown if handshake fails due to incompatible cipher suite. Error gets thrown at the later phase during prelogin and its totally unrelated to the root of failure ([error msg](https://github.com/tediousjs/tedious/issues/610#issue-254404872)). So yes, it would be super helpful if right message is thrown at right place.

v-suhame:
I'm not sure if there are other ways to verify the socket closed on server side, apart from sending data/packet to it 🤔

tvrprasad:
I think you just need to look into how socket close, end and error events are being handled. As long as these events are handled correctly in all states of the TDS state machine, server closing the socket will just fold into one of those... I think :-)

v-suhame:
Hey @tvrprasad 👋 
Socket closed on server side doesn't trigger any socket events in client side (as far as I've checked) 😅 Apparently JDBC driver & SqlClient too have this issue with encrypt ON for <=SQL Server 2008.

arthurschreiber:
> Socket closed on server side doesn't trigger any socket events in client side (as far as I've checked) 😅 Apparently JDBC driver & SqlClient too have this issue with encrypt ON for <=SQL Server 2008.

I think it does, as well. At least I had a POC that could detect the handshake failed by listening to the socket `close` event. I'm not sure how feasible that actually is, tho.

tvrprasad:
What's the feasibility issue?

v-suhame:
😕 Let me re-iterate what is happening, to see if we are on the same page :
If older version of SQL Server is used with `encrypt: true`, handshake fails and server closes the socket. `tedious` is unaware of this (we already have listener for socket `close`, it doesn't get triggered) and sends Login7, then it waits for the response, and eventually timeout throwing `ETIMEOUT`. `tedious` enters `FINAL` state at this point and closes the socket.

So, we should find a way to let the user know about handshake failure. Isn't that what we are looking at?

arthurschreiber:
> handshake fails and server closes the socket. tedious is unaware of this (we already have listener for socket close, it doesn't get triggered)

I don't think that's what is happening. See the handler here https://github.com/tediousjs/tedious/blob/76c7d06185c84913d059a2d747248afacf0ba3e4/src/connection.js#L956-L970

If the `close` event happens, `tedious` just switches to the `Final` state and closes the connection - silently. I _think_ we should instead raise an error if the connection is closed when we don't expect it to close. What do you think? 🤔

v-suhame:
The close event actually happens after `ETIMEOUT`. Here is the log,
>--log : connected to sql-2k8-sp3-1.mdomain.ad:1433
--log : Sent
--log :   type:0x12(PRELOGIN), status:0x01(EOM), length:0x002F, spid:0x0000, packetId:0x01, window:0x00
--log :   PreLogin - version:0.0.0.1 1, encryption:0x01(ON), instopt:0x00, threadId:0x00000000, mars:0x00(OFF)
--log : State change: Connecting -> SentPrelogin
--log : Received
--log :   type:0x04(TABULAR_RESULT), status:0x01(EOM), length:0x002B, spid:0x0000, packetId:0x01, window:0x00
--log :   PreLogin - version:10.0.21.124 0, encryption:0x01(ON), instopt:0x00, threadId:0x00000000, mars:0x00(OFF)
--log : State change: SentPrelogin -> SentTLSSSLNegotiation
--log : Sent
--log :   type:0x12(PRELOGIN), status:0x01(EOM), length:0x00E4, spid:0x0000, packetId:0x01, window:0x00
--log : Received
--log :   type:0x12(PRELOGIN), status:0x01(EOM), length:0x0360, spid:0x0000, packetId:0x00, window:0x00
--log : Sent
--log :   type:0x12(PRELOGIN), status:0x01(EOM), length:0x014E, spid:0x0000, packetId:0x01, window:0x00
--log : Received
--log :   type:0x12(PRELOGIN), status:0x01(EOM), length:0x0043, spid:0x0000, packetId:0x00, window:0x00
--log : TLS negotiated (AES128-SHA, TLSv1/SSLv3)
--log : Sent
--log :   type:0x10(LOGIN7), status:0x01(EOM), length:0x00F6, spid:0x0000, packetId:0x01, window:0x00
--log :   Login7 - TDS:0x74000004, PacketSize:0x00001000, ClientProgVer:0x00000000, ClientPID:0x00004EA0, ConnectionID:0x00000000
           Flags1:0xE0, Flags2:0x00, TypeFlags:0x00, Flags3:0x08, ClientTimezone:480, ClientLCID:0x00000409
           Hostname:'<>', Username:'<>', Password:'<>', AppName:'Tedious', ServerName:'sql-2k8-sp3-1.mdomain.ad', LibraryName:'Tedious'
           Language:'us_english', Database:'test', SSPI:'', AttachDbFile:'', ChangePassword:''
--log : State change: SentTLSSSLNegotiation -> SentLogin7WithStandardLogin
--log : Failed to connect to sql-2k8-sp3-1.mdomain.ad:1433 in 15000ms
Error Caught in 'connect' callback:  { ConnectionError: Failed to connect to sql-2k8-sp3-1.mdomain.ad:1433 in 15000ms
    at ConnectionError (F:/tedious/src/errors.js:10:12)
    at Connection.connectTimeout (F:/tedious/src/connection.js:866:26)
    at ontimeout (timers.js:458:11)
    at tryOnTimeout (timers.js:296:5)
    at Timer.listOnTimeout (timers.js:259:5)
  message: 'Failed to connect to sql-2k8-sp3-1.mdomain.ad:1433 in 15000ms',
  code: 'ETIMEOUT' }
--log : State change: SentLogin7WithStandardLogin -> Final
--log : connection to sql-2k8-sp3-1.mdomain.ad:1433 closed
--log : State is already Final

log is for simple
```node
connection.on('debug', function(text) {
  console.log('--log :', text);
});
connection.on('connect', function (err) {
  if (err) { console.log('Error Caught in connect callback: ', err) }
});
```
>I think we should instead raise an error if the connection is closed when we don't expect it to close. What do you think? 🤔
As you can see from the logs, `socketClose()` is called after `connect` event is fired with timeout error.

arthurschreiber:
Mh, that is weird. I'll try to dig up the test example where I've run into this.

arthurschreiber:
@Suraiya-Hameed @tvrprasad I figured out what was causing these issues!

The SSL/TLS implementation on older Windows versions is buggy and does not handle the countermeasures against the BEAST attack that were put into OpenSSL well. (OpenSSL inserts empty fragments to fix the exploit). This caused the connection process to hang during the TLS/SSL handshake forever, which then lead to the connection timeout.

Fortunately, there is an option to disable the countermeasure, which in turn fixes connecting to SQLServer instances that still use that buggy SSL/TLS implementation. 🎉

I'll open a separate PR for the fix, and then we can re-evaluate changing this option.

arthurschreiber:
I opened (and merged) https://github.com/tediousjs/tedious/pull/820, so we can move forward with this PR now.

arthurschreiber:
This was part of https://github.com/tediousjs/tedious/pull/864.

PR #704, opened 2018-02-02
 feat: IDN support

v-suhame:
To access SQL Server with International Domain Name, enable `options.serverNameAsACE`, this converts IDN serverName to ASCII Compatible Encoding (ACE).

TODO:
- [ ] add tests
- [ ] documentation


v-suhame:
I was considering moving the punny code conversion a level up (into connection.js), rather than having them at two files, connector.js and sender.js. But went with the flow of existing design and added them in both (like how `net.isIP` check and other conditions are in both files).

arthurschreiber:
@v-suhame 👋 This is great! ❤️

What do you think about dropping the config option altogether, and just always convert domain names to ACE? 🤔 It won't make a difference for regular domain names, and exotic names will work out of the box, improving the user experience.

src/connector.js
```diff
@@ -1,5 +1,6 @@
 const net = require('net');
 const dns = require('dns');
+const punycode = require('punycode');
```
arthurschreiber: The `punycode` module that ships with Node.js has been deprecated, and it is recommended to use https://github.com/bestiejs/punycode.js instead. See the comment over at https://nodejs.org/api/punycode.html#punycode_punycode.

src/connector.js
```diff
@@ -39,7 +40,8 @@ class Connector {
   }
 
   executeForHostname(cb) {
-    dns.lookup(this.options.host, { all: true }, (err, addresses) => {
+    const serverName = this.options.serverNameAsACE ? (punycode.toASCII(this.options.host)).trim() : this.options.host;
```
arthurschreiber: Are you sure the `trim` call is required here?

v-suhame: Yes, it is needed. We have couple of IDN servers internally, using the value of `punycode.toASCII()` in `dns.lookup` worked for few IDN servers and not for all, punycode conversion turns out to have trailing spaces 😅 
I will test its need again after switching to bestiejs/punycode.js 👍

src/connector.js
v-suhame: 🙇‍♀️ Thanks for pointing that out!

src/connector.js
arthurschreiber: > turns out that punycode conversion has trailing spaces 😅

Wow, that seems like a pretty serious bug? :O Do you have an example of such a domain name?

v-suhame: My apologies 😢Just went back and did a double-triple check, problem was in `config.server`, that had a trailing spaces, so of course the punycode had it to.
Thanks @arthurschreiber 😌

arthurschreiber:
I think we could integration test this somehow via hosts entries on appveyor? 🤔 
https://www.appveyor.com/docs/build-configuration/#hosts-entries

Not sure how well that will work with encryption (because the host name will not match the certificate), but it might be worth a try. 😄

v-suhame:
Opened a new PR for this #705

PR #705, opened 2018-02-02
feat: IDN support

v-suhame:
Use ASCII Compatible Encoding (ACE) value of domain name by default during connection.

I had to close #704 due to messed up commit name 😅 (not according to commitlint standard).

### DIFF
--- a/test/unit/connector-test.js
+++ b/test/unit/connector-test.js
@@ -1,5 +1,7 @@
 const Mitm = require('mitm');
 const sinon = require('sinon');
+const dns = require('dns');
+const punycode = require('punycode');
 
 const ParallelConnectionStrategy = require('../../src/connector')
   .ParallelConnectionStrategy;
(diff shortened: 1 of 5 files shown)


v-suhame:
>I think we could integration test this somehow via hosts entries on appveyor? 🤔
https://www.appveyor.com/docs/build-configuration/#hosts-entries
>Not sure how well that will work with encryption (because the host name will not match the certificate), but it might be worth a try. 😄

@arthurschreiber I was thinking of adding just unit tests, will take a look at hosts entries

arthurschreiber:
> @arthurschreiber I was thinking of adding just unit tests, will take a look at hosts entries

Integration tests might be total overkill, and unit tests might be more than enough to cover this. Whatever you feel makes the most sense! 👍

test/unit/connector-test.js
```diff
@@ -452,3 +454,42 @@ exports['ParallelConnectionStrategy'] = {
     });
   }
 };
+
+exports['Test unicode SQL Server name'] = {
+  setUp: function (done) {
+    this.sinon = sinon.sandbox.create();
```
Hadis-Fard: I don't think using Sinon :bowtie: here is necessary, as we are checking if dns.lookup is called with the server name, test could be simpler :sweat_smile: something like:

```js
  'testIDN': function(test) {
    const server = '本地主机.ad';
    const connector = new Connector({ host: server, port: 12345 }, false);
    connector.executeForHostname((err) => { test.ifError(err);});
    test.done();
  }
```

arthurschreiber:
Cool, this is great! 👍 Feel free to merge this! 🎉

v-suhame:
Thanks for reviewing 🙏 @Hadis-Fard @arthurschreiber

PR #707, opened 2018-02-08
feat: ActiveDirectoryPassword and sqlpassword authentication support

Hadis-Fard:
squash merged of #661, please see details and previous discussion there.


Hadis-Fard:
noticed the latency in generating token in adal-node lib, follow up on AzureAD/azure-activedirectory-library-for-nodejs#192

src/connection.js
```diff
@@ -1619,10 +1707,13 @@ Connection.prototype.STATE = {
...
-              this.transitionTo(this.STATE.SENT_LOGIN7_WITH_NTLM);
-            } else {
-              this.transitionTo(this.STATE.SENT_LOGIN7_WITH_STANDARD_LOGIN);
+            if (this.fedAuthInfo.requiredPreLoginResponse) {
+              return this.transitionTo(this.STATE.SENT_LOGIN7_WITH_FEDAUTH);
+            } else if (this.config.domain) {
+              return this.transitionTo(this.STATE.SENT_LOGIN7_WITH_NTLM);
+            }
+            else {
+              return this.transitionTo(this.STATE.SENT_LOGIN7_WITH_STANDARD_LOGIN);
```
v-suhame: Can you align the else condition to match with the rest?
```nodejs
} else {
```

src/token/fedauth-info-parser.js
```diff
@@ -0,0 +1,61 @@
+const async = require('async');
+
+function readFedAuthInfoOpt(data, infoData, FEDAUTHINFOID, callback) {
```
arthurschreiber: It'd be great if this function could be refactored a bit for better readability. It can be a regular non-async method, and I think it should return a `{ stsurl: '...', spn: '...' }` object.

Having it take and modify the `infoData` object in place is a bit confusing. 🤔

arthurschreiber: Also, `FEDAUTHINFOID` could be a top-level constant, instead of being passed in as well.

src/token/feature-ext-parser.js
```diff
@@ -0,0 +1,27 @@
+
+module.exports = function featureExtAckParser(parser, colMetadata, options, callback) {
```
arthurschreiber: I don't think this parser is quite right. 🤔 

According to the MS-TDS spec, the `FEATUREEXTACK` token can return multiple different features, while this implementation only ever returns the latest feature that was read.

Also, it only returns the feature data length, not the actual data itself.

What do you think about changing the token to return an object with the following structure:

```
{
  name: 'FEATUREEXTACK',
  event: 'featureExtAck',
  features: [
    {
      id: number,
      data: Buffer,
    },
    ....
  ]
}
```

Hadis-Fard:
Thanks @arthurschreiber for reviewing this PR . It contains the AAD foundation and support of two auth methods - sqlpassword and AD password, and their unit tests.  there are two other auth method -token based auth and AD integrated auth - to be implemented, but those can be added incrementally in separate PRs after the foundation is merged, which is the same approach we took in implementing AAD in other client drivers.
I will go through your review comments 🙇‍♀️ , about the ci we can go with the idea discussed in #624, @v-suhame could you help in giving the context or ideas you have 🙏 ?

v-suhame:
I have set environment variables in Travis (`AZURE_PASS`, `AZURE_SERVER`, `AZURE_USERNAME`) that has all the credentials needed for CI testing against Azure DB. We can add it as a stage after `lint` so, we end up testing with Azure only if every other test passes.
```yaml
    - stage: azure
      node_js: 8
      before_install:
        - mkdir ~/.tedious
        - >
          echo '{
            "config": {
              "server": "'${AZURE_SERVER}'",
              "userName": "'${AZURE_USERNAME}'",
              "password": "'${AZURE_PASS}'",
              "options": {
                "port": 1433,
                "database": "test",
                "encrypt": "true"
              }
            }
          }' > ~/.tedious/test-connection.json
      before_script: skip
      script: npm run test-integration
```
How does that sound? Also, can we just test with the latest version of node? 🤔

v-suhame:
We should also fix the test that fails intermittently (test/integration/pause-resume-test.js), it fails every time with remote server 😅 
@arthurschreiber Does @chdh's suggestion in https://github.com/tediousjs/tedious/pull/518#issuecomment-342657658 sounds good or did you have anything else in mind?

arthurschreiber:
@v-suhame That looks great! Can we add this change to the travis configuration in this pull request here?

v-suhame:
I forgot AAD had 4 auth mechanisms 😅 The initial credentials I had set was for `SQL Server Authentication`, I have added 2 more variables for `Active Directory - Password` authentication (`AZURE_AD_USER` and `AZURE_AD_PASS`). For now I can add 2 stages, one to test sqlpassword and other for AD password, we can always update it when we find a better alternative.

Hadis-Fard:
@v-suhame would be great if you could add azure CI in another PR to test master with regular user and no "authentication" keyword? I'm seeing failures on master when i run the tests against azure, would be better to have azure CI on master to avoid surprises

v-suhame:
Here is the result for azure ci test that I ran against my fork of tedious master branch, https://travis-ci.org/v-suhame/tedious/jobs/388396034, you can use this as reference for now. All tests are passing except for the `pause-resume-test` and `connection-test`. I'll have a look at the `connection-test` failure. We need to either increase the timer for `pause-resume-test` or come up with other alternative. 
And yes, we can add one more stage for just azure testing without involving AAD.

src/connection.js
```diff
@@ -481,7 +496,25 @@ class Connection extends EventEmitter {
...
+        if (!(config.options.authentication.toUpperCase() === this.fedAuthInfo.ValidFedAuthEnum.SqlPassword ||
+        config.options.authentication.toUpperCase() === this.fedAuthInfo.ValidFedAuthEnum.ActiveDirectoryPassword)) {
+          throw new Error('An invalid authentication method is specified');
+        }
+        if (this.config.options.tdsVersion < '7_4') {
+          throw new Error(`Azure Active Directory authentication is not supported in the TDS version ${this.config.options.tdsVersion}`);
+        }
+        this.config.options.encrypt = true;
+        this.fedAuthInfo.method = config.options.authentication;
+        if (!config.options.trustServerCertificate) {
```
v-suhame: `trustServerCertificate` by default is set to `true` for regular cases [here](https://github.com/tediousjs/tedious/pull/707/files#diff-875da84119066ec244742e95d1209f0cR483), but it is set to `false` for AAD. Can we stick with the existing default behaviour here?

src/connection.js
```diff
@@ -1696,6 +1808,47 @@ Connection.prototype.STATE = {
...
+      data: function(data) {
+        if (this.fedAuthInfo.responsePending) {
+          // We got data from the server while we're waiting for adal authentication context
+          // call to complete on the client. We cannot process server data
+          // until this call completes as the state can change on completion of
+          // the call. Queue it for later.
+          const boundDispatchEvent = this.dispatchEvent.bind(this);
+          return setImmediate(boundDispatchEvent, 'data', data);
+        } else {
+          return this.sendDataToTokenStreamParser(data);
```
v-suhame: I think we don't need this `if` condition in `data` event. As adal library will be called after `fedAuthInfo` event gets triggered (which happens after `SENT_LOGIN7_WITH_FEDAUTH.data` event), it makes sense to have this condition just in `message` event.

src/connection.js
```diff
@@ -1696,6 +1808,47 @@ Connection.prototype.STATE = {
       }
     }
   },
+  SENT_LOGIN7_WITH_FEDAUTH: {
+    name: 'SentLogin7Withfedauth',
```
v-suhame: Can you split this into 2 states, as per `3.2.1 Abstract Data Model` and name them to be inline with  `Sent LOGIN7 Record with Complete Authentication Token State`(3.2.5.3) and `Sent LOGIN7 Record with Federated Authentication Information Request State`(3.2.5.5)?

src/token/feature-ext-parser.js
```diff
@@ -0,0 +1,27 @@
+
```
v-suhame: Do you mind renaming this file to feature-ext-ack-parser.js?

src/connection.js
```diff
@@ -599,6 +633,45 @@ class Connection extends EventEmitter {
...
+          } else {
+            this.fedAuthInfo.responsePending = false;
+            this.fedAuthInfo.token = tokenResponse;
+          }
+        });
+      }
+    });
+
+    this.tokenStreamParser.on('featureExtAck', (token) => {
+      let value = undefined;
```
v-suhame: Right now there is no check to verify if the server has sent `featureExtAck`. Connection state switches from `SENT_LOGIN7_WITH_FEDAUTH` to `SENT_LOGIN7_WITH_STANDARD_LOGIN` after sending the `FEDAUTH_TOKEN` packet, so the implementation uses the general check in [`processLogin7NTLMResponse`](https://github.com/tediousjs/tedious/pull/707/files#diff-875da84119066ec244742e95d1209f0cL1181) to switch to `LOGGED_IN` state irrespective of Server's lack of response for `FEDAUTH_TOKEN`. Can you take care of this? It can go in `Sent LOGIN7 Record with Complete Authentication Token State` (TDS spec 3.2.5.3)

src/login7-payload.js
```diff
@@ -90,25 +91,46 @@ const NTLMFlags = {
...
  */
-module.exports = class Login7Payload {
+module.exports.Login7Payload = class Login7Payload {
   constructor(loginData) {
     this.loginData = loginData;
-
     const lengthLength = 4;
+
+
+    this.fedAuthPreloginRequired = (this.loginData.fedAuthInfo && this.loginData.fedAuthInfo.requiredPreLoginResponse);
```
v-suhame: Can you rename this variable to `fedAuthLogin7Required` or something similar? 😅

src/connection.js
```diff
@@ -599,6 +633,45 @@ class Connection extends EventEmitter {
...
+      if ((value = token.featureAckOpts.get(FEDAUTH_OPTIONS.FEATURE_ID)) !== undefined) {
+        if (!this.fedAuthInfo.fedAuthInfoRequested) {
+          throw new Error('Did not request federated authentication, but received the acknowledgment');
+        }
+        switch (this.fedAuthInfo.fedAuthLibrary) {
+          case FEDAUTH_OPTIONS.LIBRARY_ADAL:
+            if (0 !== value.length) {
+              throw new Error(`Federated authentication acknowledgment for ${this.fedAuthInfo.method} authentication method includes extra data`);
```
v-suhame: Can you refer to the feature as `Active Directory Authentication` in all the error message? Referring it as `federated authentication` and `Active Directory Authentication` interchangeably will cause confusion 😄

src/connection.js
```diff
@@ -599,6 +633,45 @@ class Connection extends EventEmitter {
...
+            }
+            break;
+          default:
+            throw new Error('Attempting to use unknown federated authentication library');
```
v-suhame: How about making all these Errors a part of `ConnectionError` and emitting them in `connect` event?

src/login7-payload.js
```diff
@@ -166,23 +191,32 @@ module.exports = class Login7Payload {
...
-    this.addVariableDataString(variableData, this.loginData.domain ? '' : this.loginData.userName);
-    this.addVariableDataBuffer(variableData, this.loginData.domain ? Buffer.alloc(0) : this.createPasswordBuffer());
+    this.addVariableDataString(variableData, (this.loginData.domain || this.fedAuthPreloginRequired) ? '' : this.loginData.userName);
+    this.addVariableDataBuffer(variableData, (this.loginData.domain || this.fedAuthPreloginRequired) ? Buffer.alloc(0) : this.createPasswordBuffer());
     this.addVariableDataString(variableData, this.loginData.appName);
     this.addVariableDataString(variableData, this.loginData.serverName);
-    this.addVariableDataString(variableData, '');
+    let extensionPos = undefined;
+    if (this.fedAuthPreloginRequired) {
+      // hold the positon at which the FeaturExt offset is going to be written. We write 0 in the placeholder till the offset is available.
```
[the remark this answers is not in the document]
Hadis-Fard: 🤦‍♀️

v-suhame:
@Hadis-Fard Thanks for applying the review comments 👍 LGTM

arthurschreiber:
@Hadis-Fard @v-suhame Heya 👋

I don't think this is in a mergeable state yet. 😬My biggest concern right now is that the azure integration tests seem to be completely failing. You can see the output here:

* https://travis-ci.org/tediousjs/tedious/jobs/401504525
* https://travis-ci.org/tediousjs/tedious/jobs/401504524

Any idea what's going on there? 🤔

fenngineering:
Will this PR handle Windows Authentication to Standard (non Azure) SQL Servers?

arthurschreiber:
@fenngineering No, but we're working on that as well! 😄

fenngineering:
@arthurschreiber how far are you on getting a onsite Windows Authentication PR raised? Is there anything I could do to help move this a long?

v-suhame:
Hi @fenngineering, thanks for the offer to help 😃 We have a PR for Windows Authentication  https://github.com/tediousjs/tedious/pull/497, we will work on it after https://github.com/tediousjs/tedious/pull/624. 
>will the onsite Windows Authentication PR be linux compatable?

No, Windows Authentication as the name suggests is just for Windows, we have plans to support Kerberos Integrated Authentication, that should work on Windows and Linux.

src/connection.js
```diff
@@ -1059,6 +1159,22 @@ class Connection extends EventEmitter {
     });
   }
 
+  sendFedAuthResponsePacket() {
+    const accessTokenLen = Buffer.byteLength(this.fedAuthInfo.token.accessToken, 'ucs2');
+    const data = new Buffer(8 + accessTokenLen);
+    let offset = 0;
+    data.writeUInt32LE(accessTokenLen + 4, offset);
```
arthurschreiber: `Buffer#writeUInt32LE` returns the new offset after writing to the buffer, so you could simplify this like:

```js
let offset = 0;
offset = data.writeUInt32LE(accessTokenLen + 4, offset);
offset = data.writeUInt32LE(accessTokenLen, offset);
data.write(this.fedAuthInfo.token.accessToken, offset, 'ucs2');
```

lenisha:
Any news on fixing tests and merging this PR please?

Hadis-Fard:
@arthurschreiber I changed the code so that the user doesn't set the authentication type to `azure-active-direcotry`, and setting the `activedirectorypassword` according to MSFT spec would have a feature parity with other drivers as well. Also fedauth codes in prelogin-payload were partially dropped, added them back. Finally added the config validations fixing the failing tests. I believe the PR is ready to merge, would you please review? 🙏

src/connection.js
```diff
@@ -1080,6 +1214,22 @@ class Connection extends EventEmitter {
     });
   }
 
+  sendFedAuthResponsePacket() {
+    const accessTokenLen = Buffer.byteLength(this.fedAuthInfo.token.accessToken, 'ucs2');
+    const data = new Buffer(8 + accessTokenLen);
```
Suraiya-Hameed: Can you replace Buffer() with Buffer.alloc()?

src/login7-payload.js
```diff
@@ -237,7 +256,12 @@ module.exports = class Login7Payload {
     offset = fixedData.writeUInt16LE(dataOffset, offset);
 
     // (cchUnused / cbExtension): 2-byte
-    offset = fixedData.writeUInt16LE(0, offset);
+    const extensions = this.buildFeatureExt();
+    offset = fixedData.writeUInt16LE(4, offset);
+    const extensionOffset = new Buffer(4);
+    extensionOffset.writeUInt32LE(dataOffset += 4, 0);
```
Suraiya-Hameed: also, update Buffer() to Buffer.alloc()/Buffer.from() for all 3 occurrences in this file

src/connection.js
```diff
@@ -85,6 +87,21 @@ class Connection extends EventEmitter {
       throw new TypeError('Invalid server: ' + config.server);
     }
 
+    this.fedAuthInfo = {
+      ValidFedAuthEnum: {
+        SqlPassword: 'SQLPASSWORD',
```
Suraiya-Hameed: `SqlPassword` authentication seems to fail, can you remove it?

src/connection.js
```diff
@@ -605,6 +665,57 @@ class Connection extends EventEmitter {
...
+          default:
+            this.loginError = ConnectionError('Attempting to use unknown Active Directory authentication library');
+            this.loggedIn = false;
+        }
+      }
+      else {
+        this.loginError = ConnectionError('Received acknowledgement for unknown feature');
+        this.loggedIn = false;
+      }
+
```
Suraiya-Hameed: If `fedAuthAck = undefined` and `this.fedAuthInfo.fedAuthInfoRequested = true`, then there are 2 conditions that are met, one with error message 'Did not receive Active Directory authentication acknowledgement' and other with 'Received acknowledgement for unknown feature'.

Can you restructure all these error messages such that one will not overwrite the other?

dixitsuneel:
Seems like this has been approved now. Do we know when can this be merged and made available to general public?

Hadis-Fard:
@dixitsuneel there has been an update on config options on master branch which directly affects AAD, we're going to merge the master, adopt to the new design of configs, test and hopefully 🤞 merge the PR

src/login7-payload.js
arthurschreiber: Done. 👍

arthurschreiber:
@Hadis-Fard This should be almost ready to merge now. I had to update some tests to better handle missing permissions and will extract these changes out of this PR, but I _think_ this should be ready otherwise. Would you mind taking another look over the refactorings I made? 🙇

Hadis-Fard:
@arthurschreiber  in addition to 'azure-active-directory' user should set the method of authentication which could be`ActiveDirectoryIntegrated`, `ActiveDirectoryPassword`  https://docs.microsoft.com/en-us/sql/connect/jdbc/connecting-using-azure-active-directory-authentication?view=sql-server-2017 and also `MSI token`, each with different underlying implementation, for example, adal-library is only valid for activedirectorypassword, and token based auth with use another lib, [see 2.2.6.4 Login7](https://msdn.microsoft.com/en-us/library/dd304019.aspx#Appendix_A_Target_23) FedAuth row. 

so in the config in addition to  `azure-active-password` we should have another field to set the method as ones of the above, 

```js
new Connection({
  "server": "localhost",
  "authentication": {
    "type": "default",
    "method" : "ACTIVEDIRECOTRYPASSWORD"
    "options": {
      "userName": "sa",
      "password": "foobar"
    }
  }
});
```

and have the `FedAuthEnum` back in code, check the inclusion and check tds version

```js
 if (this.config.authentication.type === 'azure-active-directory') {        if(!Object.values(this.fedAuthInfo.ValidFedAuthEnum).includes(config.authentication.toUpperCase())) {          
        throw new Error(`${config.authentication} is not a valid authentication method`);
       }
 if (this.config.options.tdsVersion < '7_4') {
         throw new Error(`Azure Active Directory authentication is not supported in the TDS version ${this.config.options.tdsVersion}`);
        }

```

arthurschreiber:
> noticed the latency in generating token in adal-node lib, follow up on [AzureAD/azure-activedirectory-library-for-nodejs#192](https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/issues/192)

The latency in getting a token was due to the way execution was delayed via recursive `setImmediate` calls. This was basically filling up the event loop and delaying the token response handling. I made a few refactorings that got rid of this waiting, and this sped up the whole token request process.

src/connection.js
```diff
@@ -744,7 +743,7 @@ class Connection extends EventEmitter {
         switch (this.fedAuthInfo.fedAuthLibrary) {
           case FEDAUTH_OPTIONS.LIBRARY_ADAL:
             if (0 !== fedAuthAck.length) {
-              this.loginError = ConnectionError(`Active Directory authentication acknowledgment for ${this.fedAuthInfo.method} authentication method includes extra data`);
```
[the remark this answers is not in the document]
arthurschreiber: I'd rather have individual auth types, like `azure-active-directory-integrated`. I was not really able to find a lot of documentation on the other two methods, but for those we could also have separate authentication types? 🤔

src/connection.js
```diff
@@ -739,17 +738,14 @@ class Connection extends EventEmitter {
         if (!this.fedAuthInfo.fedAuthInfoRequested) {
           this.loginError = ConnectionError('Did not request Active Directory authentication, but received the acknowledgment');
           this.loggedIn = false;
-        }
-        switch (this.fedAuthInfo.fedAuthLibrary) {
-          case FEDAUTH_OPTIONS.LIBRARY_ADAL:
```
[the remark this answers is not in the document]
arthurschreiber: Here as well, I'd rather have this as separate authentication types, instead of mixing this all together into one type with a large set of options.

Hadis-Fard:
👍 to have separate type for each kind of AAD auth. However all other drivers are using the same keyword for [AAD methods ](https://docs.microsoft.com/en-us/azure/sql-database/active-directory-interactive-connect-azure-sql-db#sqlauthenticationmethod-enum). I suggest to rename `azure-active-directory-password` to `ActiveDirectoryPassword` and so on.
❤️  fixing the latency issue  ❤️

arthurschreiber:
> * `"options"` in json file after `"authentication"` entry won't get set.

@Hadis-Fard there's two `options` values now, one at the top level configuration and one specific to the `authentication` that is being used. As can be seen in the Travis CI build, this works as expected. 👍

Hadis-Fard:
@arthurschreiber setting encrypt in the options object after authentication would raise `Server requires encryption, set \'encrypt\' config option to true.'` error.

```js
var connectionADnew = {
    server: "ADServer.windows.net",
    authentication: {
      type: "azure-active-directory-password",
      options: {
        userName: "ADuser",
        password: "pwd",
    },
    options: {
      encrypt: true,
    }
  }
};
```

arthurschreiber:
@Hadis-Fard I only get the error you mentioned if I don't set `encrypt` at all (or set it to `false`). This is expected until https://github.com/tediousjs/tedious/pull/694 is merged (which I hope to do sometime next week).

Hadis-Fard:
@arthurschreiber 😕 we don't need to worry about this error that much since as you said it'll gonna get fixed by merging #694

PR #708, opened 2018-02-18
refactor: switch to arrow functions

arthurschreiber:


arthurschreiber:
@Hadis-Fard Good idea!

PR #710, opened 2018-02-22
fix: support properly TIME(x) when x<5

kalinkrustev:


arthurschreiber:
@kalinkrustev Oh wow, that is a pretty neat fix! Do you think you could also add some test cases for this? If not, that's okay too! ❤️

kalinkrustev:
I did manual tests and it resolved my issues. I can check about test cases tomorrow.

arthurschreiber:
> I can check about test cases tomorrow.

Cool, thanks! Once you pushed a test case, I'll merge this and it'll trigger a new release. ❤️

kalinkrustev:
Added some integration tests, that ensure time is stored with the intended precision.

Issue #594, opened 2017-06-24
Performance Issue using Blobs (Image)  Starting with versions >= 1.13.0


egandro:
Hello,

I can confirm that "something" happened between the 1.12.x and 1.13.0 Tag that has a massive impact on performance when querying a column with an image colum.

Getting a 10mb image takes 1 sec with 1.12 and 10 sec with 1.13.

I checking this.

arthurschreiber:
I've been working on a new parser over at https://github.com/tediousjs/node-tds-token-stream. The parser is not based on callbacks nor generators, making it extremely fast, with very low memory consumption, which are the two reasons that are causing the abysmal performance you're experiencing.

Unfortunately, I've not had a lot of time to work on it, so it's in a very early stage. Any help would be highly appreciated!

egandro:
You are currently using som Yacc/Lex like grammar for the Parser?

arthurschreiber:
@egandro No, we don't. 😄 The parser we have is for parsing binary data, and last time I checked Yacc/Lex is not really good at that.

I'm aware that performance is an issue when parsing large sets of data, but there is no "quick" fix here, unfortunately. 😞

I'll close this issue as fixing this is on our roadmap, but it will take some while till we get there.

PR #716, opened 2018-03-12
feat: replace recursive callbacks with async await

v-suhame:
This PR replaces portion of parser to use async/await instead of nested callbacks. This is an initial draft to gather feedback. It works for int/smallint/bigint/tinyint and varchar types, we can extend it based on initial review. CI tests are expected to fail as I have commented out support for other data-types.

This initial draft works for `SENT_CLIENT_REQUEST` state, so I had to add few extra functions to switch between callbacks and async/await while entering and exiting this state.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1734,6 +1734,13 @@ Connection.prototype.STATE = {
   },
   SENT_CLIENT_REQUEST: {
     name: 'SentClientRequest',
+    enter: function() {
+      // TODO: remove 'enter' function after all response from SQL Server are parsed using async/await
+      this.messageIo.setAsyncAwaitFlow(true);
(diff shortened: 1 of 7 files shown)


v-suhame:
This doesn't do any good to the performance 😦 
Referring to https://github.com/tediousjs/tedious/issues/594#issuecomment-312436307 and #512, I'll look into https://github.com/tediousjs/node-tds-token-stream, and see if I can contribute. 
@arthurschreiber got any pointer on where to start? Is this something that's still in the plan?

arthurschreiber:
@v-suhame Yeah, switching to a callback-less parser is still planned, but I had no time to continue working on it. Any help would be greatly appreciated! ❤️ 🙇

Issue #725, opened 2018-04-03
nvarchar(max) Issue with Proc Parameter


RealNateB:
When attempting to return nvarchar(max) from a stored proc, it limits to 4000. 

Fixed issue in `nvarchar.js`, `resolveLength()` as shown below. I don't know your code base well enough to know if this breaks anything but for my purposes and my checking, it resolved the issue. I assume this would also apply to varchar(max). 

I tried changing .maximumLength to "max" but there are some parts of the code expecting a number there.

```js
  resolveLength: function resolveLength(parameter) {
    if (parameter.length != null) {
      return parameter.length;
    } else if (parameter.value != null) {
      if (Buffer.isBuffer(parameter.value)) {
        return parameter.value.length || 1;
      } else {
        return parameter.value.toString().length || 1;
      }
    } else {
        return "max";
      // return this.maximumLength;
    }
  },
```

v-suhame:
@RealNateB Thanks for reporting the issue, nice catch! This issue will affects all the variable-length data types (nvarchar/varchar/varbainary) returned by `OUTPUT` parameter. We will take deeper look at it :)

Issue #679, opened 2018-01-10
BulkLoad - Invalid column type from bcp client 


mhingston:
Tedious: 2.2.3
Node 9.3.0
Linux x64
SQL Server 2016 (SP1) (KB3182545) - 13.0.4001.0 (X64)

BulkLoad is failing when trying to insert into a column with type **nvarchar(max)**.

SQL:
```SQL
CREATE TABLE [dbo].[Test](
    [Body] [nvarchar](max) NOT NULL
)
```

JS:
```javascript
const {Connection, TYPES} = require('tedious');
const config = {
    userName: '...',
    password: '...',
    server: '...',
    options: {
        database: '...'
    }
};

const connection = new Connection(config);

connection.on('connect', error => {
    if(error) {
        console.log(error.message);
    }

    const table = 'Test';
    const bulkLoad = connection.newBulkLoad(table, (error, rowCount) => {
        if(error) {
            console.log(error.message); // Invalid column type from bcp client for colid 1.
        }

        console.log(rowCount); // 0
    });

    bulkLoad.addColumn('Body', TYPES.NVarChar, {nullable: false});
    bulkLoad.addRow({Body: '123'});
    connection.execBulkLoad(bulkLoad);
})
```

Error Message:

> Invalid column type from bcp client for colid 1.

arthurschreiber:
Hey there! 👋🏻

It looks like you’re passing an object `{ Body: '...' }` instead of a an array of values. Try changing it to `[ '...' ]` instead. 😊

mhingston:
Thanks @arthurschreiber, I was using [this reference](http://tediousjs.github.io/tedious/bulk-load.html) which uses an object. Do the docs need updating?

arthurschreiber:
Oh, maybe I’m wrong? I’m on my mobile right now, I’ll check again later or tomorrow. 😖

mhingston:
![screenshot_2018-01-10_20-08-34](https://user-images.githubusercontent.com/9764437/34793121-35c57874-f642-11e7-9f20-dc78e6f19648.png)
This is strange, but I'm still getting ''Invalid column type from bcp client for colid 1'', have tried using:

    bulkLoad.addRow({Body: '123'});
    bulkLoad.addRow(['123']);
    bulkLoad.addRow('123');

v-suhame:
The Error seems to be due to the `max` type used 😓 Running your code without max column (like Nvarchar (4000)) works just fine. I'll dig more into it.

[...]

PR #726, opened 2018-04-04
docs: improve documentation 

Suraiya-Hameed:
Updating docs to use length as `Infinity` for max types. #725 & #679
+
documenting the `order` event in `Request`


arthurschreiber:
Thanks! ❤️

PR #727, opened 2018-04-05
docs: pre-checks to run before submitting pull prequest

v-suhame:
Adding a pull request template that lists the per-checks to run before submitting a PR.


arthurschreiber:
Thank you for this, this is great! ❤️

Issue #678, opened 2018-01-10
[RFC] Better Data Types


arthurschreiber:
inspired by issues like https://github.com/tediousjs/tedious/issues/608, https://github.com/tediousjs/tedious/issues/675, https://github.com/tediousjs/tedious/issues/487, https://github.com/tediousjs/tedious/issues/474, https://github.com/tediousjs/tedious/issues/163, https://github.com/tediousjs/tedious/issues/339 and comments made by @chdh over at https://github.com/tediousjs/tedious/pull/674, I want to propose the following changes to the data type system to make it more robust and extensible.

These *will* be breaking changes, so I'm trying to collect some feedback first. 😅 

### What's currently wrong with data types?

Many (but not all) of the data types supported in SQL Server do not map 1:1 to JavaScript types. This causes a lot of friction in the form of precision loss or conversion failures. E.g. `numeric`/`decimal` (depending on their `precision` and `scale`), `bigint`, and many of the date and time formats, cannot be converted to/from JavaScript without a loss of precision or information.

There are JavaScript libraries (`big-number`, `bigdecimal`, `moment`, `js-joda`), that implement custom versions of compatible types in JavaScript itself, but I don't want to add these libraries as dependencies of `tedious` and "force" our users to use them, especially because there are often many competing libraries for the same use-case, and it's not always clear which one is the "best".

### How do we fix this?

I want data types to be pluggable with custom encoder/decoder functions. A encoder function would be responsible for converting a JavaScript object into it's TDS binary representation, and the decoder function would be responsible for converting the binary representation back into a JavaScript object.

`tedious` would ship with a default set of encoder and decoder functions. Data Types which can be mapped to JavaScript types directly would be fully supported out-of the box, while all other data types will come with default implementations that will encode from and decode to plain JavaScript strings. This would ensure that `tedious` can work with all data-types out of the box, *without* loss of information.

Separate plug-in libraries can then provide more specific conversion functions.

### Fantasy API

Here's a very rought example of how this could look like:

```js
const tedious = require('tedious');
const BigInt = tedious.TYPES.BigInt;
const BigInteger = require("big-integer");

const MAX_BIGINT = new BigInteger('9223372036854775807')
const MIN_BIGINT = new BigInteger('-9223372036854775808')

tedious.TYPES.BigInt.defineEncoder(function(value: any) : Buffer {
[...]

PR #732, opened 2018-04-18
Option to parse decimal and numeric fields as strings

darko1979:
I added a boolean option **returnDecimalAndNumericAsString** that enables parsing decimal and numeric fields as strings instead of float.
This was needed for our project because of precision loss using float.
Usage example:

```
var Connection = require('tedious').Connection;
var config = {
  server: 'localhost',
  userName: 'test',
  password: 'test',
  options: {
    instanceName: 'SQLEXPRESS',
    returnDecimalAndNumericAsString: true
  }
};
var connection = new Connection(config);
...
```


v-suhame:
@darko1979 Sorry that you're running into precision loss issue. It is a known one, you can read more about the plan for better data type support at #678. It will be optimal to have a fix for all data types and also for both read and write operation.

FYI, you might want to extend the fix to consider negative values too, if you're using this in your project.

Thanks!

darko1979:
@v-suhame Thanks for noticing I missed adding sign for negative numbers. I have seen and read the plan for better data type support #678, it sound good, but until it's realized this is our only solution for now.

v-suhame:
Marking this as a deprecated property makes sense 👍 I'll review it soon. Can you also add some test cases?

src/value-parser.js
```diff
@@ -276,27 +276,40 @@ function valueParse(parser, metaData, options, callback) {
...
-                default:
-                  return parser.emit('error', new Error(sprintf('Unsupported numeric size %d', dataLength - 1)));
+              if (options.returnDecimalAndNumericAsString) {
+                return parser.readBuffer(dataLength - 1, (buffer) => {
+                  let value = convertLEBytesToString(buffer);
+                  if (metaData.scale > 0) {
+                    value = value.padStart(metaData.scale + 1, '0');
+                    const idx = value.length - metaData.scale;
+                    value = value.slice(0, idx) + '.' + value.slice(idx);
+                    if (sign == -1) value = '-' + value;
```
v-suhame: Can you move this check for `sign` outside the `if` condition?

src/value-parser.js
```diff
@@ -276,27 +276,40 @@ function valueParse(parser, metaData, options, callback) {
...
-                case 16:
-                  readValue = parser.readUNumeric128LE;
-                  break;
-                default:
-                  return parser.emit('error', new Error(sprintf('Unsupported numeric size %d', dataLength - 1)));
+              if (options.returnDecimalAndNumericAsString) {
+                return parser.readBuffer(dataLength - 1, (buffer) => {
+                  let value = convertLEBytesToString(buffer);
+                  if (metaData.scale > 0) {
+                    value = value.padStart(metaData.scale + 1, '0');
```
v-suhame: `padStart` will work only if the value is zero, as `convertLEBytesToString` gives padded value for everything else. Can you add a check for calling `padStart`?

src/value-parser.js
darko1979: My mistake, in our project I added sign check correctly outside if condition, did not notice I pasted that line here in a wrong place.

darko1979:
I'm not sure what kind of test should I write?  For our project I created a temp table and did some inserts and reads for various values and checked if read value is correct.

KAMAELUA (approved):

PR #738, opened 2018-05-10
Fix `tls.createSecurePair` deprecation warning

arthurschreiber:
This is based on the great work done by @joux3 over in #689. 🙇 

I forked `duplexpair` into a `native-duplexpair`, a version that uses the `Duplex` stream exposed by Node.js' built-in `stream` module, instead of the `readable-stream` module. This solves an issue where the `duplexpair` usage on Node.JS versions before the 8.x line would cause a crash.


arthurschreiber:
A few additional notes - I'm not happy about forking `duplexpair`, but for as long as we support Node.js 6.x, this is what I believe to be the best solution.

Once we drop support for Node.js 6.x, we can switch over to the original `duplexpair` module. 👍

LiamDotPro:
Four years, Is it finally time for the depreciation warning to be depreciated, hold me boys.. It's happening 💯

ovikholt:
How is this coming along?

LiamDotPro:
@arthurschreiber Are there any updates to share on this? When can we expect this to be merged?

src/message-io.js
```diff
@@ -83,47 +86,33 @@ module.exports = class MessageIO extends EventEmitter {
...
+    const securePair = this.securePair = {
+      cleartext: tls.connect({
+        socket: duplexpair.socket1,
+        secureContext: credentials,
+        rejectUnauthorized: !trustServerCertificate
+      }),
+      encrypted: duplexpair.socket2
+    };
 
-        // Verify that server's identity matches it's certificate's names
```
arthurschreiber: By switching to `tls.connect`, we no longer need to perform server verification ourselves, and can instead make use of the `rejectUnauthorized` option. 🎉

src/message-io.js
```diff
@@ -83,47 +86,33 @@ module.exports = class MessageIO extends EventEmitter {
...
-    // Forward the error to the socket so the connection gets properly cleaned up.
-    this.securePair.cleartext.on('error', (err) => {
-      this.socket.destroy(err);
-    });
-
-    // On Node >= 0.12, the encrypted stream automatically starts spewing out
-    // data once we attach a `data` listener. But on Node <= 0.10.x, this is not
-    // the case. We need to kick the cleartext stream once to get the
-    // encrypted end of the secure pair to emit the TLS handshake data.
-    this.securePair.cleartext.write('');
```
arthurschreiber: By switching to `tls.connect`, we no longer need to "kick" the secure socket to start performing the handshake - it's all handled automatically. 👍 (Also, the comment here was severely out-of-date. 🤷‍♂️)

LiamDotPro:
@arthurschreiber Legendary.

PR #739, opened 2018-05-10
BREAKING CHANGE: Drop support for Node.js v4.x

arthurschreiber:
Node.js v4.x is no longer maintained by the Node.js project, and I think we should also drop support for it.


Hadis-Fard:
Now that we are not supporting v4, what are your thoughts about removing  `new Buffer` in favor of `buffer.from` and `buffer.alloc`  according to https://github.com/nodejs/node/wiki/Breaking-changes-between-v5-and-v6 ?

Hadis-Fard:
Updated the buffers in #742

kibertoad:
@Suraiya-Hameed So what still needs to happen before we can merge this?

kibertoad:
It does sound like a lot of hassle without obvious benefits - and if it's preferable to avoid releasing multiple versions, PRs could be merged to a separate branch before going to master.

kibertoad:
I can prepare up-to-date version of https://github.com/tediousjs/tedious/pull/742 if there is a decision to merge this and carry on with the obsolence plan.

PR #742, opened 2018-05-11
refactor: use buffer from and alloc instead of buffer constructor

Hadis-Fard:
Used the `buffer.alloc` and `buffer.from` instead of `new buffer`
https://github.com/nodejs/node/blob/master/lib/buffer.js#L164-L171


kibertoad:
@arthurschreiber As promised: https://github.com/tediousjs/tedious/pull/797
I would really appreciate if you could squash it on merge instead of manual squash as I've already had to redo entire PR once after rebase gone wrong.

arthurschreiber:
This was superseded by #797

Issue #733, opened 2018-04-18
Logging undefined value in ConnectionError message


mribichich:
When there's a `ConnectionError` by a `Timeout` the message is being logged with the port `undefined`

`ConnectionError: Failed to connect to SRVPROD03:undefined in 15000ms`

It seems it comes from `connection.js` [line 867](https://github.com/tediousjs/tedious/blob/b038b447b06d844a3b954efb2a30a46937b8c564/src/connection.js#L867)

Here's the stack:

```
{ ConnectionError: Failed to connect to SRVPROD03:undefined in 15000ms
    at Connection.tedious.once.err (C:\Users\Administrator\Desktop\zkaccess\node_modules\mssql\lib\tedious.js:216:17)
    at Object.onceWrapper (events.js:272:13)
    at Connection.emit (events.js:180:13)
    at Connection.connectTimeout (C:\Users\Administrator\Desktop\zkaccess\node_modules\tedious\lib\connection.js:795:12)
    at ontimeout (timers.js:466:11)
    at tryOnTimeout (timers.js:304:5)
    at Timer.listOnTimeout (timers.js:267:5)
  code: 'ETIMEOUT',
  originalError:
   { ConnectionError: Failed to connect to SRVPROD03:undefined in 15000ms
    at ConnectionError (C:\Users\Administrator\Desktop\zkaccess\node_modules\tedious\lib\errors.js:12:12)
    at Connection.connectTimeout (C:\Users\Administrator\Desktop\zkaccess\node_modules\tedious\lib\connection.js:795:28)
    at ontimeout (timers.js:466:11)
    at tryOnTimeout (timers.js:304:5)
    at Timer.listOnTimeout (timers.js:267:5)
     message: 'Failed to connect to SRVPROD03:undefined in 15000ms',
     code: 'ETIMEOUT' },
  name: 'ConnectionError' }
```

v-suhame:
Hey @mribichich, thanks for reporting the issue. Port number is set to `undefined` if `instanceName` is used, that's causing the issue . Feel free to submit a pull request with fix + test 😄 
Using a template literals for the `message` variable (in the line you pointed) should do it.

PR #744, opened 2018-05-17
fix: update timeout error with instance name

v-suhame:
Addresses #733


v-suhame:
True, maybe we can push it along with other breaking change (node 4 drop).

arthurschreiber:
This is looking great. I'd have marked this as a `fix`, because `refactor`s generally shouldn't result in a change in behavior. Anyway, feel free to merge! 👍

PR #758, opened 2018-06-20
validate bulk load data

mottykohn:
This will add flexibility. e.g. Dates can be strings...
At a speed cost...

**Before submitting a PR :**
6. Run commitlint (`node_modules/.bin/commitlint --from origin/master --to HEAD`). Refer [commit conventions](http://marionebl.github.io/commitlint/#/concepts-commit-conventions) and [commit rules](http://marionebl.github.io/commitlint/#/reference-rules).

### DIFF
--- a/src/bulk-load.js
+++ b/src/bulk-load.js
@@ -133,7 +133,7 @@ module.exports = class BulkLoad extends EventEmitter {
         length: c.length,
         scale: c.scale,
         precision: c.precision,
-        value: row[arr ? i : c.objName]
+        value: c.type.validate(row[arr ? i : c.objName])
       }, this.options);
     }
   }


v-suhame:
@mottykohn Thanks for submitting the PR.

I get error if I try to send date as string, both with and without your fix 😅 
```node
  bulkLoad.addColumn('_date', TYPES.Date, { nullable: true });
  bulkLoad.addRow({ _date: '1990-05-05' });
```

And its the same error in both the cases,
```
src\data-types\date.js:28
        var dstDiff = -(parameter.value.getTimezoneOffset() - YEAR_ONE.getTimezoneOffset()) * 60 * 1000;
                                        ^
TypeError: parameter.value.getTimezoneOffset is not a function
```

It is because string value becomes number at the below condition, and `writeParameterData()` tries `getTime()` or `getTimezoneOffset()` on that number.

https://github.com/tediousjs/tedious/blob/c22c894d887797fc0062a07b68ffaad9b3a06e72/src/data-types/date.js#L33-L39


Best way to send string to any SQL Server datatype, would be to set the TYPE as VarChar or NVarChar and let Server handle the conversion.

```node
  bulkLoad.addColumn('_date', TYPES.VarChar, { nullable: true });
```

v-suhame:
Whatever check+conversion `validate()` does is flaky, that is causes error even with your fix, sorry about that 😓

mottykohn:
Didn't see that because I'm on option.useUTC=true. Should I fix by 
```
    if (!(value instanceof Date)) {
      value = Date.parse(value);
      if (!options.useUTC && !isNaN(value))
        value = new Date(value + new Date().getTimezoneOffset() * 60 * 1000)
    }
```
 I am using this library dynamically where I don't know column types in advance. But I see that I'm running into the old timezone problem.

v-suhame:
Unfortunately we will run into this issue for the other temporal data-types (date, datetime, datetime2, time) too. Wouldn't setting the type as VarChar  or NVarChar work for your case?

mottykohn:
I'm using mssql.recordset.toTable() function - I guess I could try to loop over columns and switch all date* columns to Varchars, Just not sure how easy that is to reassign a column type in that library.

Issue #755, opened 2018-06-13
DateTime precision error.


heroboy:
code:
```js
var d1 = new Date(2018, 6 - 1, 11);
var d2 = new Date(d1.getTime() + 24 * 3600 * 1000 - 1); //change to -2 ,no error.
TestDate();
function TestDate()
{
	db.acquire((err, conn) =>
	{
		if (err) return;

		var request = new tedious.Request("select @v1;", function (err)
		{
			if (err)
			{
				console.log(err);
				return;
			}
			conn.release();
		});
		request.on('row', cols =>
			console.log(cols.map(x => x.value))
		);
		request.on('returnValue', (name, value) =>
		{
			console.log(name, value);
		})

		request.addParameter('v1', tedious.TYPES.DateTime, d2)
		request.addOutputParameter('v4', tedious.TYPES.VarChar, null);
		conn.execSql(request);
	})
}
```

error message:
'The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. 
Parameter 3 ("@v1"): The supplied value is not a valid instance of data type datetime. Check the source 
data for invalid values. An example of an invalid value is data of numeric type with scale greater than 
precision.',

heroboy:
I find in `DateTime.js`,the write value is `{ days: 43261, threeHundredthsOfSecond: 25920000 }`
`2590000 = 24 * 3600 * 300`, is one day. Maybe it can not great than this.

heroboy:
You should use `useUTC: false,` to test this code.

Issue #524, opened 2017-03-03
Upgrade SQL version for integration tests


tvrprasad:
Had to disable two tests (dateTimePrecision_9_day_flip and outputDatePrecision_9_day_flip) in PR https://github.com/tediousjs/tedious/pull/521/files due to what seems like a corner case bug in the version of SQL Server running on AppVeyor.

@arthurschreiber How do we upgrade the SQL Server version for the integration tests?

arthurschreiber:
This is part of the AppVeyor setup - installing new SQL Server versions on every build is really not feasible. But we could ask them nicely if they could just upgrade their images to contain the latest SQL Server patches.

tvrprasad:
Would you be sending them mail? Let me know if there is something I can help with.

Hadis-Fard:
related to #762

PR #762, opened 2018-06-27
fix: datetime max millisecond issue

Hadis-Fard:
Datetime round millisecond in nearest .xx3, .xx7, .xx0, when the ms is rounded to 000 in case of 23:59:59:999, day should be incremented. the issue manifest itself in useUTC true or false

Fixes: #755

### DIFF
--- a/src/data-types/datetime.ts
+++ b/src/data-types/datetime.ts
@@ -39,6 +39,12 @@ const DateTime: DataType = {
       threeHundredthsOfSecond = milliseconds / (3 + (1 / 3));
       threeHundredthsOfSecond = Math.round(threeHundredthsOfSecond);
 
+      // 25920000 equals one day
+      if (threeHundredthsOfSecond === 25920000) {
+        days += 1;
+        threeHundredthsOfSecond = 0;
+      }
+
       buffer.writeUInt8(8);
       buffer.writeInt32LE(days);
 
--- a/test/integration/parameterised-statements-test.js
+++ b/test/integration/parameterised-statements-test.js
@@ -608,9 +608,16 @@ describe('Parameterised Statements Test', function() {
     );
   });
 
-  // This test fails on the version of SQL Server in AppVeyor.
-  // exports.dateTimePrecision_9_day_flip = (test) ->
-  //  execSql(test, TYPES.DateTime, new Date('January 1, 1998 23:59:59.999'), null, null, new Date('January 2, 1998 00:00:00.000'))
+  it('should test date time precision_9 hr flip', function(done) {
+    execSql(
+      done,
+      TYPES.DateTime,
+      new Date('January 1, 1998 23:59:59.999'),
+      null,
+      null,
+      new Date('January 2, 1998 00:00:00.000')
+    );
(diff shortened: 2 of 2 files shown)


v-suhame:
Looks good. Can you add a test too? Under \test\integration\datatypes-in-resultset, maybe?

src/data-types/datetime.js
```diff
@@ -37,6 +37,11 @@ module.exports = {
 
       threeHundredthsOfSecond = milliseconds / (3 + (1 / 3));
       threeHundredthsOfSecond = Math.round(threeHundredthsOfSecond);
+      // 25920000 equals one day
+      if (threeHundredthsOfSecond === 25920000) {
```
arthurschreiber: Is this the maximum value `threeHundredthsOfSecond` can have?

arthurschreiber:
Is this fix "correct"? According to https://github.com/tediousjs/tedious/issues/524, this failure might be due to a bug in SQL Server itself?

src/data-types/datetime.js
Hadis-Fard: according to TDS
>One 4-byte signed integer that represents the number of days since January 1, 1900. Negative numbers are allowed to represent dates since January 1, 1753.
One 4-byte unsigned integer that represents the number of one three-hundredths of a second (300 counts per second) elapsed since 12 AM that day

the buffer consists of days/threehundredth of a second, 25920000 equals to a full day and if we hit it, we should send the day separately

Hadis-Fard:
@arthurschreiber it makes sense since the number of days and time are sent *separately*, so if the time is full 24 hrs then it should change the day. this fixed the two tests mentioned in #524

MichaelSun90:
Hi @arthurschreiber, I have looked into this issue and I think this is a valid and important fix. However, I think we can still add some extra comments to make it more clear. 

As the comment on line 40, we could add some extra explanation to clear things up:
// Since we are rounding milliseconds to 300ths of a second (the precision of a DateTime), if we happen
// to round up to a full day's worth of 300ths of a second (25920000 ), we need to carry that over to the // day part

In addition, I think we can add some comments above line 38 to just explain the expression: milliseconds / (3 + (1 / 3)). I just want to confirm with you first that I understand this correctly. Is this expression equivalent to "milliseconds /1000 * 300"?  I understand this as convert millisecond number back to seconds by divided it by 1000 and count the "one three-hundredths of a second" by dividing it with 300. If this is correct, is it better for us to add some comment here to explain this?

Is it possible for you to help add the comments to this PR, or I can create a new PR and you can close this one?

daviduzumeri:
Any movement on this? This just bit my ass passing in an end-of-day value. Basically any time you try to do an outer limit of 23:59:59.999 in a where clause it'll explode on you.

arthurschreiber:
Thank you very much for preparing this @Hadis-Fard, sorry this has taken so long to get merged! 🙇

franck-mahieu:
Hello, this issue was really fixed ?
I use the version 9.2.3 and have the same approximate milliseconds issue.

PR #763, opened 2018-07-08
fix: improve handling of TLS errors

arthurschreiber:
This fixes an issue in the TLS negotiation code of `tedious` that has been around for quite some time. 😞I also think this is one of the issues that has been plagueing people that wanted to use `tedious` in combination with encrypted connections (as is required e.g. when connecting to Azure).

Errors happening in the TLS layer were basically not handled at all. Usually, unhandled `error` events cause an uncaught exception in node.js (and bring down the whole process), but it looks like that this is not happening on `SecurePair`/`TLSSocket` instances. 🤷‍♂️ In our cases, these unhandled errors caused the connection to just go into some weird blocking state and become unusable. This most likely also caused issues with connection pool libraries. 🤔 

Anyway, by listening to `error` events and forwarding them to the underlying network socket instance, errors in the TLS layer get bubbled up and handled correctly - the connection gets closed with an error.

A very simple test case that shows this behaviour is to connect to an SQL Server instance, but block any cipher from being used.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -931,7 +931,7 @@ class Connection extends EventEmitter {
   }
 
   socketError(error) {
-    if (this.state === this.STATE.CONNECTING) {
+    if (this.state === this.STATE.CONNECTING || this.state === this.STATE.SENT_TLSSSLNEGOTIATION) {
       const message = `Failed to connect to ${this.config.server}:${this.config.options.port} - ${error.message}`;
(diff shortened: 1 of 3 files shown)


arthurschreiber:
@v-suhame Thanks for the review!

This has been released as `tedious@2.6.3` and tagged as `latest`.

PR #768, opened 2018-07-23
refactor: switch to a `Message` stream based architecture

arthurschreiber:
For as long as `tedious` has existed, it's internals have been very unsupportive of stream based processing. The two main disadvantages of this are higher memory usage (because data is buffered) and lower performance (data is only sent after everything was buffered).

Previous changes in `tedious`'s internals moved us a bit closer to having internals be based on Node.js stream primitives, but we're still not there yet. 😅

This change here represents another big step in this direction. It introduces a new message based stream architecture and consists of three pieces:

* `Message`: A simple `PassThrough` stream that just forwards whatever is written into it, without modification.
* `IncomingMessageStream`: A `Transform` stream that takes in a raw TDS packet stream, creates `Message` instances and extracts and forwards the packet data into these messages.
* `OutgoingMessageStream`: A `Duplex` stream that takes in `Message` instances, reads their data and wraps it into raw TDS packets.

This basically makes both `IncomingMessageStream` and `OutgoingMessageStream` a "stream of streams". 😬All these streams correctly implement and respect backpressure (e.g. when the underlying socket gets saturated).

The changes are currently restricted to the `MessageIO` layer, so we don't reap any of the stream benefits just yet, but it lets us gradually move in the right direction.

This is partially based on the work #528.

### DIFF
--- a/src/incoming-message-stream.js
+++ b/src/incoming-message-stream.js
@@ -1,0 +1,86 @@
+// @flow
+
+const BufferList = require('bl');
+const { Transform } = require('readable-stream');
+
+const Message = require('./message');
+const { Packet, HEADER_LENGTH } = require('./packet');
+
+import type Debug from './debug';
+
+/**
+  IncomingMessageStream
+  Transform received TDS data into individual IncomingMessage streams.
+*/
+class IncomingMessageStream extends Transform {
+  debug: Debug;
+  bl: BufferList;
+  currentMessage: Message | typeof undefined;
+
+  constructor(debug: Debug) {
+    super({ readableObjectMode: true });
+
+    this.debug = debug;
+
+    this.currentMessage = undefined;
+    this.bl = new BufferList();
+  }
+
+  processBufferedData(callback: () => void) {
+    // The packet header is always 8 bytes of length.
+    while (this.bl.length >= HEADER_LENGTH) {
+      // Get the full packet length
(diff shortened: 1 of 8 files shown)


src/message-io.js
```diff
@@ -135,56 +96,18 @@ module.exports = class MessageIO extends EventEmitter {
...
-      this.securePair.cleartext.write(packet.buffer);
-    } else {
-      this.socket.write(packet.buffer);
-    }
-  }
-
-  logPacket(direction, packet) {
-    this.debug.packet(direction, packet);
-    this.debug.data(packet);
+    const message = new Message({ type: packetType, resetConnection: resetConnection });
```
chdh: Where will these message objects be destroyed and event handlers released?

src/incoming-message-stream.js
```diff
@@ -0,0 +1,86 @@
...
+        this.bl.consume(length);
+
+        // TODO: Get rid of creating `Packet` instances here.
+        const packet = new Packet(data);
+        this.debug.packet('Received', packet);
+        this.debug.data(packet);
+
+        let message = this.currentMessage;
+        if (message === undefined) {
+          message = new Message({ type: packet.type(), resetConnection: false });
```
chdh: Where will these message objects be destroyed and event handlers released?

PR #769, opened 2018-07-23
fix: `end` events on the socket outside the `FINAL` state are errors

arthurschreiber:
If we receive an `end` event outside the `FINAL` state, the remote side hung up on us, and this should be treated like an error.

/cc @landondavidson

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -952,7 +952,11 @@ class Connection extends EventEmitter {
 
   socketEnd() {
     this.debug.log('socket ended');
-    this.transitionTo(this.STATE.FINAL);
+    if (this.state !== this.STATE.FINAL) {
+      const error = new Error('socket hang up');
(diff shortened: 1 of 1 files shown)


willmorgan:
@arthurschreiber Is there any news on when this might make it into a 2.6.3 release?

arthurschreiber:
@willmorgan is there a reason you can’t use the newly released v2.7.1 instead? There’s been no breaking changes between v2.6.x and v2.7.1.

willmorgan:
@arthurschreiber That'd be even better 😉

willmorgan:
@arthurschreiber @Suraiya-Hameed This is a good fix in theory but it hasn't landed for the consumers of the node-mssql library which is under the same organisation. Can someone turn the cranks and get someone to look into this PR, please?

https://github.com/tediousjs/node-mssql/pull/718

PR #772, opened 2018-07-27
ci: enable Node.js 10 builds

arthurschreiber:
This enables builds for Node.js 10.x on Appveyor and Travis.


scripts/appveyor/config.ps1
```diff
@@ -25,7 +25,7 @@ $config = @{
       database = "master";
       requestTimeout = 25000;
       cryptoCredentialsDetails = @{
-        ciphers = "RC4-MD5"
```
arthurschreiber: The `RC4-MD5` cipher is no longer available in default Node.js 10.x builds, as it's totally insecure. As I'm not really sure which other ciphers are available and compatible with SQL Server on Appveyor, I just took the "old backward compatibility" cipher list as recommended by https://wiki.mozilla.org/Security/Server_Side_TLS#Old_backward_compatibility.

v-suhame:
For appveyor, I've added SQL Server 2017 under `services` and got rid of the powershell script. I don't know why we had `cryptoCredentialsDetails` option configured, but tests are running fine without it (at least in SQL Server 2017)  

Sample run:
https://ci.appveyor.com/project/v-suhame/tedious/build/14/job/do5rrk8071rt3owp

PR #773, opened 2018-07-27
docs: replace Gitter with Slack

arthurschreiber:
I'm not using Gitter, and I don't think any other contributor really does either.


v-suhame:
Can we delete or archive the gitter lobby, so existing users don't post questions there by any chance?

PR #776, opened 2018-07-30
feat: add `Request.setTimeout()` and `BulkLoad.setTimeout()`

chdh:
Implements Request.setTimeout() and BulkLoad.setTimeout(), which can be used to override the timeout specified when the connection was created.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -37,7 +37,7 @@ module.exports = class Request extends EventEmitter {
   handle: ?any; // TODO: Figure out the type here.
   error: ?Error;
   connection: ?any; // TODO: This should be `Connection`, not `any`.
-  timeout: ?number | typeof undefined;
+  timeout: number | typeof undefined;
 
   callback: (?Error) => void;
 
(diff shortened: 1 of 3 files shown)


src/request.js
```diff
@@ -37,6 +37,7 @@ module.exports = class Request extends EventEmitter {
   handle: ?any; // TODO: Figure out the type here.
   error: ?Error;
   connection: ?any; // TODO: This should be `Connection`, not `any`.
+  timeout: ?number;
```
arthurschreiber: I think the correct type to use here would be:

```js
timeout: number | typeof undefined;
```

`?number` is `number | null | typeof undefined;`, but we shouldn't allow `null` here, but let's not add another incorrect type here.

chdh:
@arthurschreiber 
Thanks for the review. I have completed the changes and adapted the API documentation (#777). Can I merge this PR myself, or will you do that?

arthurschreiber:
Thank you for preparing this! ❤️ 

> Can I merge this PR myself, or will you do that?

I usually try to take on last review pass before merging. In this case, all the changes were looking great, but to ensure that the `fix` commits you made don't show up in the automated changelog generated over at https://github.com/tediousjs/tedious/releases, I squashed all the changes together.

chdh:
@arthurschreiber 
Thanks. (I expected the commits to get squashed.)

PR #790, opened 2018-09-14
chore: adding issue template

Suraiya-Hameed:


package.json
```diff
@@ -1,17 +1,6 @@
...
-    "Bret Copeland <bret@atlantisflight.org> (https://github.com/bretcope)",
-    "Bryan Ross <bryan@rossipedia.com> (https://github.com/rossipedia)",
-    "Ciaran Jessup <ciaranj@gmail.com>",
-    "Cort Fritz <cfritz@caa.com>",
-    "lastonesky",
-    "Patrik Simek <patrik@patriksimek.cz>",
-    "Phil Dodderidge <pdodde@poyntz.com>",
-    "Zach Aller"
-  ],
+  "contributors": "https://github.com/tediousjs/tedious/graphs/contributors",
```
arthurschreiber: I don't `contributors` accepts a single string to be set. 🤔

Instead of changing this here to a URL, I think we could automatically generate a list of authors from the git commit log before we publish the package via `semantic-release`.

Suraiya-Hameed: You are right 😅 Do you suggest generating a CONTRIBUTORS.md with list of authors, something with

```bash
$ git log --format='%aN' | sort -u > CONTRIBUTORS.md
```

Hadis-Fard:
@arthurschreiber could you merge this PR? would be good to have template for issues

PR #795, opened 2018-09-23
chore: upgrade to `big-number@1.0.0`

kibertoad:
Update big-number to fix some implementation bugs; public API doesn't seem to have changed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "dependencies": {
     "babel-runtime": "^6.26.0",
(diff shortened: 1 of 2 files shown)


arthurschreiber:
@kibertoad would you mind squashing all these three commits together into a single commit, using the first commits message to make `commitlint` happy?

kibertoad:
Will do, but why can't it be squashed automatically by GitHub during merge instead?..

PR #796, opened 2018-09-23
chore: upgrade to `big-number@1.0.0`

kibertoad:


kibertoad:
@arthurschreiber Are you sure about having draconian commit structure rules? They are only marginally useful when there is a huge team to manage, and for a project with limited amount of contributors like Tedious it is likely to bring more harm than good, because last thing that people who volunteer to contribute their free time want to worry about is whether some arbitrary formal policies are adhered do - they usually have enough of that at work :)

arthurschreiber:
I understand the commit naming rules might seem “draconian” at first, but they’re part of the release automation we have in place for `tedious`, called `semantic-release`. When a pull request is merged, commit messages are automatically analyzed and a new release might be cut, depending on the types of commits that were merged. This has proven to be a much better approach then manually pushing releases. 

I understand that this might mean a higher barrier on contributors. Usually I don’t ask for rebasing/squashing of commits on drive-by Pull Requests, and instead do this myself once all other issues in the PR were resolved, but I’maveling right now and don’t have a development machine with me. 😅

Thank you for this contribution and for squashing the commits together! 🙇🏻‍♂️

PR #797, opened 2018-09-24
refactor: use buffer from and alloc instead of buffer constructor (updated)

kibertoad:


src/ntlm-payload.js
```diff
@@ -150,20 +150,20 @@ class NTLMResponsePayload {
...
     const hash = this.ntHash(password);
-    const identity = new Buffer(user.toUpperCase() + domain.toUpperCase(), 'ucs2');
+    const identity = Buffer.from(user.toUpperCase() + domain.toUpperCase(), 'ucs2');
     return this.hmacMD5(identity, hash);
   }
 
   ntHash(text: string) {
-    const hash = new Buffer(21).fill(0);
+    const hash = Buffer.alloc(21).fill(0);
     hash.fill(0);
```
kibertoad: Is there any good reason to do fill twice, or this should be safe to be rewritten as `Buffer.alloc(21, 0)` with `hash.fill(0);` dropped?

kibertoad: OK, pushed the change

kibertoad:
@arthurschreiber All systems green! Is there anything left?

arthurschreiber:
@Hadis-Fard @kibertoad  Thanks to both of you for working on this! 🙇🏻‍♂️

Issue #195, opened 2014-09-25
Custom instanceLookup timeout/retries


asilvas:
https://github.com/pekim/tedious/blob/96130ca96a971c50d983ed94307e2ea1f684c81f/src/connection.coffee#L440

does not expose the timeout & retries arguments for:

https://github.com/pekim/tedious/blob/fc19fba187b205671decfd4f056d482b5357d126/src/instance-lookup.coffee#L11

This can frequently be an issue in inactive development environments, which happens to be a huge pain for our integration tests.

PR #802, opened 2018-09-28
fix: remove retries for instance-lookup

### LINKED ISSUE #672: Unavailable SQL_SERVER_BROWSER_PORT gets wrong error message
The correct and informative error messages in "instance-lookup.js" are not getting delivered, because the catch-all error message 'Failed to connect to' from "connection.js" dominates.  

It is clear that the dominating error message is not intended for my situation.  The message from "connection.js" makes an assumption that the "port" is known, when in fact the whole point of function "InstanceLookup()" is to get the port.  The message shows 'undefined' as the port.

Context: I am going through a VPN, and I presume that the administrators have not opened up access to SQL_SERVER_BROWSER_PORT.  It took me many hours to come to this conclusion; correct messages from "instance-lookup.js" would have helped.

Suraiya-Hameed:
Removed option.retries from instance-lookup.js, this option does not exist and it is never passed from connection.js.
Also, fixes #672

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -939,7 +939,7 @@ class Connection extends EventEmitter {
 
   socketError(error) {
     if (this.state === this.STATE.CONNECTING || this.state === this.STATE.SENT_TLSSSLNEGOTIATION) {
-      const message = `Failed to connect to ${this.config.server}:${this.config.options.port} - ${error.message}`;
+      const message = `Failed to connect to ${this.config.server}: ${this.config.options.port ? `:${this.config.options.port}` : `\\${this.config.options.instanceName}`} - ${error.message}`;
       this.debug.log(message);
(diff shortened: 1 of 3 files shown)


Suraiya-Hameed:
Same here. From issue #195, probably the were intending to remove retries but missed it!

IanChokS:
@arthurschreiber eslint test is failing because of previous commit messages not following commitlint. Not sure how to fix. But overall this PR has been updated 😄

arthurschreiber:
@IanChokS I double checked this. This removes the `retries` option, and _also_ completely removes the retries. This might be problematic, as the server instance lookup is done via UDP, which is not reliable and can have packets silently dropped.

I think maybe we should remove the retries option, but also exposing a way to abort the lookup? Then we could just keep on retrying until we hit the connection timeout. @IanChokS @MichaelSun90 what do you think?

PR #803, opened 2018-09-30
upgrade outdated dependencies

arthurschreiber:
This upgrades us to:

* `semantic-release@15.9.16`
* `nodeunit@0.11.3`
* `babel-eslint@10.0.1`
* `flow-bin@0.81.0`
* `sinon@6.3.4`
* `readable-stream@3.0.3`
* `commitlint@7.1.2`
* `bl@2.0.1`
* `async@2.6.1`

I also checked in `package-json.lock`, as that is recommended since `npm@5`. This file is not published in `tedious` releases and has no impact there - it's just relevant when working on `tedious` itself.


.travis.yml
```diff
@@ -17,10 +17,6 @@ stages:
 services:
   - docker
 
-cache:
```
kibertoad: why not?

arthurschreiber: This is mentioned in the commit message of https://github.com/tediousjs/tedious/pull/803/commits/e5e9387cf0c69975f68574107eeeda15c52a6b48: When a `package-lock.json` exists in the repo, Travis performs always a "clean installation" of dependencies that deletes the `node_modules` folder always, thus completely negating the caching of the folder.

kibertoad: grrr, Travis being stupid. One more reason why CircleCI is so much better :D

PR #804, opened 2018-09-30
chore: replace readable-stream with native streams

kibertoad:
Since Node 4 support was dropped, native streams could be used instead of readable-stream library.


kibertoad:
@arthurschreiber I apologize for commit history being a mess again, I deleted and recreated my entire fork repo, but apparently some history remained in my local repo. If it's not feasible to just squash all of it, please let me know, I'll recreate the PR anew.

arthurschreiber:
@kibertoad Thanks again for another PR! ❤️

I'm wondering a bit about the reason for dropping our use of `readable-stream`. The built-in `stream` module in Node.js underwent a bunch of breaking changes in Node.js v9.x and v10.x, and by upgrading `readable-stream` to v3.x (as suggested in https://github.com/tediousjs/tedious/pull/803) we can ensure that streams work the same way in `tedious` across all supported versions of Node.js.

What do you think? 🙇

kibertoad:
I would say that there is no definite good answer here; it is generally a question of whether to prioritise predictability or consistency. Generally as a user I expect Streams to behave the way I see them documented in the documentation for the Node.js version that I'm using; and when I notice that they don't, I have to go inside the library that I use and then go "oh, they are using custom library for that. How does it work?.."
However, it definitely prevents things from breaking when someone updates Node.js. Personally I prefer the "test coverage + adding new Node versions into CI ASAP" approach for addressing it, but it obviously requires some additional effort.
A bit of anecdotal reference: "mysql2" replaced their readable-streams implementation with native Streams in May, and I'm not aware of any issues that were raised due to that.

arthurschreiber:
After thinking about this a bit more, I think we won't switch over to `readable-stream` - at least not yet. We don't really expose Stream instances directly to consumers (at least not yet), and I want to make use of the new [`stream.pipeline`](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) functionality (which was added in Node v10.x) soon, and that is also part of `readable-stream` v3.x.

Still, thank you a lot for preparing this pull request! 🙇 ❤️

kibertoad:
@arthurschreiber Yup, that makes perfect sense if streams are not exposed to the end user. Shall I prepare PR for the upgrade to readable-stream or is that something that you plan to do yourself already?

arthurschreiber:
@kibertoad It's already part of https://github.com/tediousjs/tedious/pull/803. 😄

kibertoad:
awesome!

Issue #782, opened 2018-08-16
Unhandled ETIMEDOUT error on connection timeout


kuseni:
Node.js 8.11.4
tedious@2.6.4
Windows 10 Pro 1803 17134.191

Code:
```javascript
log('started');

var conn = new Connection({
	server:   '8.8.8.8',
	userName: 'test',
	password: 'test' });

conn.on('connect', function(err) {
	log(`connect(): err="${err}"`);
	if(err)
		return;
	conn.close();
});

conn.on('end',   ()    => { log('end()'); });
conn.on('debug', (msg) => { log(`debug(): ${msg}`); });
//conn.on('error', (err) => { log(`error(): err="${err}"`); });
```

Log:
```
2018-08-16 23:41:15.941: started
2018-08-16 23:41:30.954: debug(): Failed to connect to 8.8.8.8:1433 in 15000ms
2018-08-16 23:41:30.959: connect(): err="ConnectionError: Failed to connect to 8.8.8.8:1433 in 15000ms"
2018-08-16 23:41:30.976: debug(): State change: Connecting -> Final
2018-08-16 23:41:30.980: end()
2018-08-16 23:41:36.956: debug(): Connection lost - connect ETIMEDOUT 8.8.8.8:1433

events.js:183
      throw er; // Unhandled 'error' event
      ^
ConnectionError: Connection lost - connect ETIMEDOUT 8.8.8.8:1433
    at ConnectionError (***\node_modules\tedious\lib\errors.js:12:12)
    at Connection.socketError (***\node_modules\tedious\lib\connection.js:1020:28)
    at ***\node_modules\tedious\lib\connection.js:861:25
    at Socket.onError (***\node_modules\tedious\lib\connector.js:49:9)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

I expect not receiving unhandled error and closing app immediatly after connect event (not 6 seconds later).

Suraiya-Hameed:
#659 should have fixed this 🤔 looks like something is amiss as your are already in the latest driver, @Hadis-Fard should be able to have a look at it once she is back from vacation.

Hadis-Fard:
[...]

PR #806, opened 2018-10-10
fix unhandled error after timeout when connecting with ip

Hadis-Fard:
passing the socket to the connection and checking the final state in the socket error fixed the issue #782

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -942,7 +942,7 @@ class Connection extends EventEmitter {
       const message = `Failed to connect to ${this.config.server}:${this.config.options.port} - ${error.message}`;
       this.debug.log(message);
       this.emit('connect', ConnectionError(message, 'ESOCKET'));
-    } else if (this.state === this.STATE.FINAL){
+    } else if (this.state === this.STATE.FINAL) {
       this.socket.destroy();
     } else {
       const message = `Connection lost - ${error.message}`;
--- a/src/connection.js
+++ b/src/connection.js
@@ -814,12 +814,12 @@ class Connection extends EventEmitter {
 
     new Connector(connectOpts, multiSubnetFailover).execute((err, socket) => {
       if (err) {
+        this.socket = socket;
         return this.socketError(err);
       }
 
       if (this.state === this.STATE.FINAL) {
-        socket.destroy();
-        return;
+        return socket.destroy();
       }
 
       this.socket = socket;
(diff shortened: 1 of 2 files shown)


kuseni:
@Hadis-Fard , yes, socket is created in initialiseConnection at 21:56:46 (in my last comment in #782),  but connection timeout emmits connect event (with err != null) at 21:57:01. Socket is created already at 21:57:01, but it's local for Connector yet and is not visible for Connection (so connectTimeout can't destroy it). Socket's error event is emmited at 21:57:07, so Node can't exit at 21:57:01 (immediately after connection timeout) because there is a non-destroyed socket and waits 6 seconds more to exit. I think it's more correctly to destroy socket on connection timeout and don't wait 6 seconds more (I'm not good in tedious architecture and Connection states, so I can't propose a solution - publishing socket to Connection immediately on creation OR providing Connector.cancel() method OR something else - but I am convicted that Node should exit immediately after connection timeout, not 6 seconds later)

kuseni:
I think it's possible to destroy socket in connecting state:
```javascript
const util = require('util');
const net = require('net');

function log(...args) {
	var dt = new Date();
	dt.setMinutes(dt.getMinutes() - dt.getTimezoneOffset());
	process.stdout.write(`${dt.toISOString().slice(0, -1).replace('T', ' ')}: ${util.format.apply(null, args).replace(/\n/g, '\n  ')}\n`);
}

log('started');

const socket = net.connect({ host: '8.8.8.8', port: 1433});
socket.on('error', (err) => log(`${err}`));
setTimeout(() => {
	log(`timeout, socket.connecting = ${socket.connecting}`);
	socket.destroy();
}, 15000);
process.on('exit', () => log('exit'));
```

```
2018-10-17 13:09:57.607: started
2018-10-17 13:10:12.615: timeout, socket.connecting = true
2018-10-17 13:10:12.619: exit
```

arthurschreiber:
I think a better fix would be to push down the timeout handling into the `Connector` class, and remove it from `Connection`.

Suraiya-Hameed:
🤔 we would still need part of timeout handling in `Connection`, as `Connector` class only handles socket and we need connection timeout to consider entire login process.

arthurschreiber:
I'm closing this, as the underlying issue was properly fixed via https://github.com/tediousjs/tedious/pull/1210.

PR #811, opened 2018-10-22
Configuration rework

arthurschreiber:
BREAKING CHANGE: This drops support for previousl deprecated 
configuration values and types, and makes the checks a lot stricter than 
before.

Many configuration options no longer accept `null` values and instead 
rely on values being present or `undefined` instead.

In other cases, configuration options that previously e.g. accepted 
string values and automatically converted them to numbers now only 
accept actual numbers.

Some options explicitly allow `null` values to be passed in and take 
that as a hint to use whatever is configured on the SQL Server instead.

---

Another change included here is a rework of how authentication options are specified. Instead of having to specify authentication options via the top level `userName`, `password` and `domain` options, a new `authentication` sub object is added, which can hold the `type` of authentication to use (either `default` or `ntlm` right now). Additionally, each authentication type also takes a set of `options`.

```js
new Connection({
  "server": "localhost",
  "userName": "sa",
  "password": "foobar"
});

// becomes

new Connection({
  "server": "localhost",
  "authentication": {
    "type": "default",
    "options": {
      "userName": "sa",
      "password": "foobar"
    }
  }
});
```

When `config.authentication` is not specified, `tedious` will still recognize the old config options, but will emit a deprecation warning.

This new configuration scheme for authentication will allow us to easily support other authentication types and their options, without having to add more and more top-level config options.

---

These changes are a pre-requisite to #707.

The changes in this pull request are pretty similar to https://github.com/tediousjs/tedious/pull/692, but updated to the latest changes that were made in the `master` branch. Thank you @Hadis-Fard for your work!

One thing I want to point out here is that I did not refactor these checks into re-usable functions, as keeping these checks inline will ensure that `flow` can actually understand their meaning, even though we don't really use `flow` to check the `connection.js` file yet. 🤷‍♂️

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -43,36 +43,6 @@ const DEFAULT_TDS_VERSION = '7_4';
 const DEFAULT_LANGUAGE = 'us_english';
 const DEFAULT_DATEFORMAT = 'mdy';
 
-function deprecateNonBooleanConfigValue(optionName, value) {
-  if (typeof value !== 'boolean') {
-    deprecate(`Passing non-boolean values for ${optionName} is deprecated and will be removed. Please specify \`true\` or \`false\` instead.`);
-  }
-}
-
-function deprecateNullConfigValue(optionName, value) {
-  if (value === null) {
-    deprecate(`Passing \`null\` for ${optionName} is deprecated and will be removed. Please pass an explicit value or \`undefined\` instead.`);
-  }
-}
-
-function deprecateNullFallbackToDefaultConfigValue(optionName, value) {
-  if (value === null) {
-    deprecate(`Passing \`null\` for ${optionName} will not fallback to a default value in future tedious versions. Please set a value explicitly if you require a different value from the one configured for your target SQL Server.`);
-  }
-}
-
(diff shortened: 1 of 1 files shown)


src/connection.js
```diff
@@ -82,23 +52,20 @@ class Connection extends EventEmitter {
...
-      deprecateNonStringConfigValue('domain', config.domain);
+    if (config.domain !== undefined && typeof config.domain !== 'string') {
+      throw new TypeError('The "config.domain" property must be of type string.');
     }
-    deprecateNullConfigValue('domain', config.domain);
 
-    if (config.userName != undefined) {
-      deprecateNonStringConfigValue('userName', config.userName);
+    if (config.userName !== undefined && typeof config.userName !== 'string') {
+      throw new TypeError('The "config.userName" property must be of type string.');
```
Suraiya-Hameed: Shouldn't all these conditions be this `||` instead of `&&`?

arthurschreiber: I think this check is right. 🤔 `config.userName`, `config.password` and `config.domain` are all optional and thus can be `undefined`, or  need to be a string.

PR #820, opened 2018-11-02
Stop OpenSSL from inserting empty fragments

arthurschreiber:
SQLServer versions running on older Windows versions that have a buggy 
SSL/TLS implementation don't play well with the BEAST countermeasure 
implemented by OpenSSL, and cause the connection process to hang during 
the SSL/TLS handshake.

Due to the nature of the BEAST attack (it's specific to web browsers), 
we can safely disable the counter measures to fix the connection 
process, without jeopardizing security. 🎉

### DIFF
--- a/src/message-io.js
+++ b/src/message-io.js
@@ -7,8 +7,10 @@ import type { Duplex } from 'stream';
 import type { TLSSocket } from 'tls';
 import type { Socket } from 'net';
 
+// $FlowFixMe
+const constants = require('constants');
+
 const tls = require('tls');
-const crypto = require('crypto');
 const DuplexPair = require('native-duplexpair');
 const { EventEmitter} = require('events');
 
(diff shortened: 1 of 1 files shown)


src/message-io.js
```diff
@@ -7,8 +7,10 @@ import type { Duplex } from 'stream';
 import type { TLSSocket } from 'tls';
 import type { Socket } from 'net';
 
+// $FlowFixMe
+const constants = require('constants');
```
arthurschreiber: For some reason, Flow does not know about this module. 🤷‍♂️

src/message-io.js
```diff
@@ -62,7 +64,20 @@ module.exports = class MessageIO extends EventEmitter {
   }
 
   startTls(credentialsDetails: Object, hostname: string, trustServerCertificate: boolean) {
-    const credentials = tls.createSecureContext ? tls.createSecureContext(credentialsDetails) : crypto.createCredentials(credentialsDetails);
```
arthurschreiber: `tls.createSecureContext` is available in all Node.js versions that we support, so I removed this conditional here.

Issue #814, opened 2018-10-27
`parseInt` should not be used to convert float to int


heroboy:
Because `parseInt` convert the parameter to string first.
In source code like `int.js`:
https://github.com/tediousjs/tedious/blob/7504a2525b7a1aec1b14f07ec2416aa5d12e7612/src/data-types/int.js#L26-L38

If the input value is a number, it will make some undesirable result. For example:
```
parseInt(4.7 * 1e22, 10); // Very large number becomes 4
parseInt(0.00000000000434, 10); // Very small number becomes 4
```
More example can reference to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

Hadis-Fard:
@heroboy Could you send a repo using Tedious where there's an issue with int datatype?

heroboy:
@Hadis-Fard 
check this example:
```javascript
function TestInt()
{
	db.acquire((err,conn)=>{
		if (err) return;
		var request = new tedious.Request("select @v;",function(){
			conn.release();
		});
		request.on('row',cols=>{
			//The output is [1]. But should be [0]
			console.log(cols.map(x => x.value))
		});
		request.addParameter('v', tedious.TYPES.Int, 0.0000000000000000000001);
		conn.execSql(request);
	});
}
```

PR #834, opened 2018-11-26
convert values using `Number` instead of `parseInt`

Hadis-Fard:
Using `parseInt` in parsing string to numbers would have different result for the same value, using `Math.round` would fix the issue (#814)

``` js
function sendReq(){
    const request = new Request("select @v1, @v2;", (err) =>{
        console.log(err);
    });
    request.on('row', cols =>{
        console.log(cols.map(x => x.value));
    });
    request.addParameter('v1', TYPES.TinyInt, 0.000000000000001);  // results in 1
    request.addParameter('v2', TYPES.TinyInt, "0.00000000000001"); // results in 0
    conn.execSql(request);
}
```
this PR fixes the commitlint issues in #823

### DIFF
--- a/src/data-types/int.ts
+++ b/src/data-types/int.ts
@@ -15,13 +15,14 @@ const Int: DataType = {
     buffer.writeUInt8(4);
   },
 
-  writeParameterData: function(buffer, parameter, options, cb) {
+  writeParameterData: function(buffer, parameter, _options, cb) {
     if (parameter.value != null) {
       buffer.writeUInt8(4);
-      buffer.writeInt32LE(parseInt(parameter.value));
+      buffer.writeInt32LE(Number(parameter.value));
     } else {
       buffer.writeUInt8(0);
     }
+
     cb();
   },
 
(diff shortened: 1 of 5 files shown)


arthurschreiber:
@Hadis-Fard Would you mind adding test cases for this?

src/data-types/int.ts
```diff
@@ -15,28 +15,35 @@ const Int: DataType = {
...
       return new TypeError('Invalid number.');
     }
+
     if (value < -2147483648 || value > 2147483647) {
-      return new TypeError('Value must be between -2147483648 and 2147483647.');
+      return new TypeError('Value must be between -2147483648 and 2147483647, inclusive.');
     }
-    return value;
+
+    return value | 0;
```
arthurschreiber: For `smallint`, `tinyint` and `int`, we use `| 0` here to force the value to be converted internally to a 32 bit integer in the V8 engine. I haven't benchmarked this, but it _might_ improve performance by giving V8 better optimization hints. 🤷‍♂

arthurschreiber: We can't use the same thing for `bigint`, because `| 0` only works with the first 32 bits of a value and thus would lead to incorrect results.

Issue #475, opened 2016-11-04
tedious TVP, is a becoming blocking operation for large data


RajatBanerjee:
I am trying to do a call a stored procedure with a TVP as a parameter . This TVP is built from a CSV, with ~ 30000 rows . When the request begins, the thread gets blocked and any request into the application is not handled .

I have tried calling a sp whihc internally sends the response back after a 5 mins delay, but in that case the operayionn is Async . This leads  me to believe that the Sp call itself is not a blocking call, but the build up upto the sp call \ processing the data inside tedious is blocking

```js
 let tvp = {
        columns: [
            { name: 'Domain', type: TYPES.VarChar, length: 100, nullable: true },
            { name: 'LanguageISO', type: TYPES.VarChar, length: 10, nullable: true },
            { name: 'Channel', type: TYPES.VarChar, length: 10, nullable: true },
            { name: 'DateStart', type: TYPES.VarChar },
            { name: 'DateEnd', type: TYPES.VarChar },
        ],
        rows: [
        ]
    };

//some for loop
   tvp.rows.push(newRow)

     const connection = new Connection(connectionStrings[environment]);

    // /**on establishing connection execute teh request */
    connection.on('connect', function (err) {
        if (err) {
            console.error("Exception==>", err);
            callback(err);
        } else {
            logger.info("Connection established with sql server", stopwatch.ms);
           
            const request = new Request("someStoredProcedure", function (error, rowCount) {
                connection.close();
                if (!error) {
                    logger.info("Procedure call complete", stopwatch.ms);
                     callback(null, output);
                    }
                    /**excel is not valid */
                    else {
                        callback('File validation Failed', output);
                   }
                } else {
                    logger.error("DB Error => ", error, stopwatch.ms);
                    callback(error.message, output);
                }
            });
            request.addParameter('Parameter1', TYPES.TVP, tvp);
            request.on('row', function (columns) {
                var row = [];
                columns.forEach(function (column) {
                    row.push(column.value);
                });
                output.push(row);
            });
            logger.info("Begin procedure call", stopwatch.ms);
            connection.callProcedure(request);
        }
    });
```

arthurschreiber:
Yes, you're right, writing large TVP is currently a blocking operation.

[...]

PR #844, opened 2019-01-11
Non blocking tvp

pgraham-acn:
Addresses #475 by adding setImmediate callbacks to allow work waiting on the event loop to process after each parameter and each row of a TVP parameter is written to the internal buffer.


pgraham-acn:
Forgot to rebase to master. Was working against 2.7.1 since that is the Tedious version currently being used in our repos. Will open a new PR against master.

Issue #545, opened 2017-03-12
Calling cancel in SENT_ATTENTION returns false and writes debug error


tvrprasad:
This should not be an error and we should return true. User has no way to know that the connection is in SENT_ATTENTION state and even otherwise it should not be error.

ElfenLiedGH:
@tvrprasad i saw it when "requestTimeout" was end.

tvrprasad:
@ElfenLiedGH I don't quite follow. Can you please elaborate? Thanks.

chdh:
>>@tvrprasad i saw it when "requestTimeout" was end.

> @ElfenLiedGH I don't quite follow. Can you please elaborate? Thanks

It's exactly the program logic sequence that I described in a [comment](https://github.com/tediousjs/tedious/pull/518#discussion_r105517228) for #518:

- The application calls connection.execSql() to start a query.

- The query takes longer than expected. A request timeout occurs. The driver sends ATTENTION to the server and enters the SENT_ATTENTION state. The application has no way of knowing this.

- While the driver is in the SENT_ATTENTION state and waiting for the attention response from the server, the application calls Connection.cancel(). The current implementation of the cancel() method considers this an error, writes the debug message "Requests can only be canceled in the SentClientRequest state, not the SentAttention state" and returns false.

I still think that `cancel()` and similar API methods should just ignore such conditions and keep quiet. Moreover the [API doc](http://tediousjs.github.io/tedious/api-connection.html#function_cancel) does not mention anything about the return value (`true` or `false`) from `cancel()`.

tvrprasad:
@chdh I agreed with you on that specific comment, which is why I opened this issue :-)

Hadis-Fard:
If application calls `connection.cancel()` in row event where all rows have been received before server gets ATTENTION message, it outputs all columns and throws ` RequestError: Canceled.` error, I think it should not throw an error and just ignore.

chdh:
> I think it should not throw an error and just ignore.

I disagree. Tedious does not really throw an error, it just passes a `RequestError` object to the callback function. This is in accordance with the TDS protocol specification [3.2.5.8 Sent Attention State](https://msdn.microsoft.com/en-us/library/dd340677.aspx):
"...  indicate the completion of the query to the upper layer together with the cause of the Attention ...".

Hadis-Fard:
[...]

Issue #534, opened 2017-03-09
Add 'cancel' function on Request class


tvrprasad:
Currently 'cancel' is on Connection. This works fine since we only have a single Request per connection right now. It belongs on Request class logically though as it's the Request that's being cancelled.

We should mark the cancel function on Connection class deprecated so we can remove it in a later release.

chdh:
Some additional info to justify this:

-  In the upper level driver node-mssql, the [cancel()](https://github.com/patriksimek/node-mssql#cancel) method is a member of the Request class.
-  In the .NET API, the [Cancel()](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommand.cancel.aspx) method ist a member of the SqlCommand class.

arthurschreiber:
Yeah, this definately feels like it should be on the Request and not the Connection, long term.

We can deprecate this somewhere in the 2.x releases and drop it in 3.0.

arthurschreiber:
This will be added via https://github.com/tediousjs/tedious/pull/846. 😬

PR #846, opened 2019-01-14
Rework request cancellation

arthurschreiber:
This is an attempt to rework how request cancellation works, to make it more generic (it now works with regular requests and both streaming as well as non-streaming bulk loads) and more conformant to the MS-TDS spec.

This also adds a new `.cancel` method on the `Request` and `BulkLoad` classes, and deprecates the `cancel` method on the `Connection` class itself.

The changes in 8c9ce1d warrant a major version bump, as they change the behaviour of requests that are canceled after they were sent to the server. Instead of continuing to forward various events from the token stream parser to the canceled request, all this information will be "discarded", as described in [`3.2.5.8 Sent Attention State`](https://msdn.microsoft.com/en-us/library/dd340677.aspx) of the MS-TDS specification:

> If the response is structurally valid and it does not acknowledge the Attention as described in section 2.2.1.7, then the TDS client MUST discard any data contained in the response and remain in the "Sent Attention" state.
> 
> If the response is structurally valid and it acknowledges the Attention as described in section 2.2.1.7, then the TDS client MUST discard any data contained in the response, indicate the completion of the query to the upper layer together with the cause of the Attention (either an upper-layer cancellation as described in section 3.2.4 or query timeout as described in section 3.2.2), and enter the "Logged In" state.

Fixes: https://github.com/tediousjs/tedious/issues/545
Fixes: https://github.com/tediousjs/tedious/issues/534
Fixes: https://github.com/tediousjs/tedious/pull/566

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1029,7 +1029,7 @@ class Connection extends EventEmitter {
 
   requestTimeout() {
     this.requestTimer = undefined;
-    this.cancel();
+    this.request.cancel();
     const timeout = (this.request.timeout !== undefined) ? this.request.timeout : this.config.options.requestTimeout;
     const message = 'Timeout: Request failed to complete in ' + timeout + 'ms';
     this.request.error = RequestError(message, 'ETIMEOUT');
@@ -1486,8 +1486,7 @@ class Connection extends EventEmitter {
     });
 
     bulkLoad.once('cancel', () => {
-      request.canceled = true;
-      request.emit('cancel');
+      request.cancel();
     });
 
     this.execSqlBatch(request);
@@ -1736,8 +1735,7 @@ class Connection extends EventEmitter {
       return false;
     }
 
-    this.request.canceled = true;
-    this.request.emit('cancel');
+    this.request.cancel();
     return true;
   }
 
--- a/src/connection.js
+++ b/src/connection.js
@@ -2111,7 +2111,10 @@ Connection.prototype.STATE = {
       this.attentionReceived = false;
     },
     events: {
-      socketError: function() {
+      socketError: function(err) {
+        const sqlRequest = this.request;
+        this.request = undefined;
+        sqlRequest.callback(err);
         this.transitionTo(this.STATE.FINAL);
       },
       data: function(data) {
--- a/src/connection.js
+++ b/src/connection.js
@@ -690,13 +690,15 @@ class Connection extends EventEmitter {
       this.emit('errorMessage', token);
       if (this.loggedIn) {
         if (this.request) {
-          this.request.error = RequestError(token.message, 'EREQUEST');
-          this.request.error.number = token.number;
-          this.request.error.state = token.state;
-          this.request.error['class'] = token['class'];
-          this.request.error.serverName = token.serverName;
-          this.request.error.procName = token.procName;
-          this.request.error.lineNumber = token.lineNumber;
+          if (!this.request.canceled) {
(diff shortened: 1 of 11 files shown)


src/connection.js
```diff
@@ -1638,8 +1667,23 @@ class Connection extends EventEmitter {
       this.request.rows = [];
       this.request.rst = [];
 
+      let message;
+
+      this.request.once('cancel', () => {
```
chdh: Somehow there should be a check whether the cancel event stems from the currently active request. When the user calls calcel on an old request or BulkLoad, it should be ignored.

src/connection.js
```diff
@@ -1672,13 +1718,8 @@ class Connection extends EventEmitter {
       const message = 'Requests can only be canceled in the ' + this.STATE.SENT_CLIENT_REQUEST.name + ' state, not the ' + this.state.name + ' state';
```
chdh: When cancel() is called after a timeout, a debug message is still displayed, see #545.

src/connection.js
```diff
@@ -1638,8 +1667,23 @@ class Connection extends EventEmitter {
...
+      let message;
+
+      this.request.once('cancel', () => {
+        // There's two ways to handle request cancelation:
+        if (message.writable) {
+          // - if the message is still writable, we'll set the ignore bit
+          message.ignore = true;
+        } else {
+          // - but if the message has been ended (and thus has been fully sent off),
+          //   we need to send an `ATTENTION` message to the server
```
chdh: Can the connection already be in the SENT_ATTENTION state, because `requesttimeout()` has executed?
Or in any other state?

chdh:
I have made some comments, but the fundamental problem is that the program logic in `connection.js` is much too complicated to really comprehend it. There is such a jungle of events, callbacks, flags, states, etc.

The TDS protocol state machine should be encapsulated in a separate module and isolated from the rest.

src/connection.js
arthurschreiber: I want to deprecate `.cancel` on `Connection`, and thus did not touch this function more than necessary. We could change this function to:

```js
if (!this.request) {
  return false;
}

if (this.request.canceled) {
  return false;
}

this.request.cancel();
return true;
```

And still deprecate (and later remove) this function? 🤔

src/connection.js
arthurschreiber: You're right. 🙇 The request timeout functionality was still handling the sending of the attention message itself. I changed it to simply call `this.request.cancel` instead. This way, only the first call to `.cancel` will have any effect, and all later calls will not do anything.

src/connection.js
arthurschreiber: You're right! I'll rework this a bit to ensure we only ever handle `cancel` events from the current request. I'll also add some tests for this.

arthurschreiber:
> I have made some comments, but the fundamental problem is that the program logic in `connection.js` is much too complicated to really comprehend it. There is such a jungle of events, callbacks, flags, states, etc.
> 
> The TDS protocol state machine should be encapsulated in a separate module and isolated from the rest. There should be clear documented APIs for every module. Private and public variables should be distinguished, etc.

Yeah, you're totally right. 😞 The way the state machine is currently translated to code is... not quite optimal. I'm trying to move the code in a better direction, through a small set of changes that I have in mind. The first set of changes is over https://github.com/tediousjs/tedious/pull/847. It's nothing revolutionary, but I feel like doing small, iterative changes is going to yield results as soon as possible and will help us move in the right direction quicker.

PR #855, opened 2019-01-31
Allow returning GUIDs with lower case characters

MichaelSun90:


arthurschreiber:
Before this can be merged, this will require an additional configuration option that can toggle between the old and the new behaviour.

@MichaelSun90 Can you add that? Otherwise this will have to wait for me to find some time and implement that.

arthurschreiber:
@MichaelSun90 I reworked your changes to allow whether GUIDs should be returned in lower or uppercase via a new connection option called `lowerCaseGuids`. It defaults to `false` for backwards compatibilty. We can deprecate `false` values in the future and then completely remove the option again and always return lower case GUIDs.

Issue #856, opened 2019-01-31
NTLM authentication regression with 5.0.0 & 5.0.1


BrianLawton:
@arthurschreiber The 5.0.0 and 5.0.1 releases seem to have a regression with "ntlm" authentication whereby it's failing with error below. The 4.2.0 release continues to work. As a side note, "default" authentication works fine in 5.0.x builds but unfortunately for us doesn't meet our requirements so we're staying with 4.2.0 for now.

**Environment:**
- Windows 10 Pro & Server 2016
- NPM 6.7.0       -- we see it on machines with earlier versions too
- Node 10.15.1  -- we see it on machines with earlier versions too

**Error Messages:**
```
events.js:174
      throw er; // Unhandled 'error' event
      ^
ConnectionError: Connection lost - read ECONNRESET
    at ConnectionError (C:\tedious poc\node_modules\tedious\lib\errors.js:13:12)
    at Connection.socketError (C:\tedious poc\node_modules\tedious\lib\connection.js:1151:26)
    at Socket.socket.on.error (C:\tedious poc\node_modules\tedious\lib\connection.js:998:14)
    at Socket.emit (events.js:194:15)
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)
Emitted 'error' event at:
    at Connection.socketError (C:\tedious poc\node_modules\tedious\lib\connection.js:1151:12)
    at Socket.socket.on.error (C:\tedious poc\node_modules\tedious\lib\connection.js:998:14)
    [... lines matching original stack trace ...]
    at process._tickCallback (internal/process/next_tick.js:63:19)

```
**Repo package.json:**
```
{
  "name": "tedious-poc",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "tedious": "^5.0.1"
  }
}
```
**Repo JavaScript:**
```
var Connection = require('tedious').Connection;

var connection = new Connection({
    server: 'myWindowsServer',
    options: {
        encrypt: true,
        instanceName: 'myInstance',
        database: 'myDatabase'
    },
    // authentication: {
    //      type: "default",
    //      options:{
    //         userName: 'mySqlLogin',
    //         password: 'mySqlPassword',
[...]

PR #857, opened 2019-02-01
Fix NTLM authentication

arthurschreiber:
This fixes an issue with NTLM authentication that was introduced with a recent refactoring, as reported in https://github.com/tediousjs/tedious/issues/856.

It also extends the AppVeyor configuration to enable NTLM authentication tests, so regressions will be caught much earlier in the future. 🙇 

This fixes https://github.com/tediousjs/tedious/issues/856

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ before_test:
 
   - mkdir C:\Users\appveyor\.tedious
   - ps: >-
-      Set-Content -Value '{
+      Set-Content -Value $('{
         "config": {
           "server": "localhost",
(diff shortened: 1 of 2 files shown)


BrianLawton:
Hi @arthurschreiber, thanks for jumping on this!! You gotta love a one-line fix. :) I manually applied the change locally to 5.0.1 and it's now working. I'll watch for the official release and do another test then.

arthurschreiber:
@BrianLawton Thanks for verifying! ❤️That's good enough for me, a release should drop shortly.

PR #871, opened 2019-02-22
feat: adds extra override configuration

Naes0:
I had trouble accessing and old sql server so I added these extra configuration options to edit the TDS version, Reconnect time etc.

It only overrides the default values if they have been set otherwise it will just use the default values, so its backwards compatible.


arthurschreiber:
Hey @Naes0!

All of these options can already be passed in and changed from their default values via the `config` object that is passed to the `Connection` constructor. I'm closing this as this already works as one would expect.

Please let me know if you disagree or if I might have missed something! 🙇

PR #874, opened 2019-03-08
fix: only validate properties of the parameter object not its prototype

dhensby:
See https://github.com/tediousjs/node-mssql/issues/261#issuecomment-341202793

This fixes an issue where if some code adds to the `Object.prototype` the `validateParameters` function will iterate over them and cause a validation failure.

This may not be the right approach, happy to take feedback.

### DIFF
--- a/src/request.js
+++ b/src/request.js
@@ -102,6 +102,9 @@ class Request extends EventEmitter {
   makeParamsParameter(parameters: Parameter[]) {
     let paramsParameter = '';
     for (let i = 0, len = parameters.length; i < len; i++) {
+      if (!parameters.hasOwnProperty(i)) {
+        continue;
+      }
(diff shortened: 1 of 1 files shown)


arthurschreiber:
@dhensby I'm not sure this actually fixes the problem mentioned in the `node-mssql` issue. Could you add a test case that shows what is currently broken?

arthurschreiber:
I'm closing this one out. I don't think this fixes the actual issue. `parameters` is an `Array` here, and we're iterating over the array's values, I really don't see how properties of one of it's prototypes can have any effect here. 🤔 

If you think I closed this one in error, feel free to reopen! 👍

dhensby:
You are correct - sorry, chasing my tail a bit over here!

Issue #877, opened 2019-03-14
Request.pause() does no longer stop memory overflow (regression)


chdh:
Up to Tedious 2.6.4 everything works fine, but starting with Tedious 2.7.0 the memory keeps getting filled with data from the SQL server even after `Request.pause()` has been called.

This problem has been reported by @gsamal in a comment to #518, and in tediousjs/node-mssql#832.

I have written a test program that confirms the problem: [testPauseResume.zip](https://github.com/tediousjs/tedious/files/2968441/testPauseResume.zip)

arthurschreiber:
This was fixed via #878.

arthurschreiber:
The fix for this is part of the `6.0.1` release. 🎉

PR #878, opened 2019-03-14
fix: flow control in message-io.js

chdh:
Fix for #877

@arthurschreiber could you have a look at this? In Tedious 2.7.0 you changed the data flow architecture to a stream of streams. The inner stream must also be paused.

### DIFF
--- a/src/incoming-message-stream.js
+++ b/src/incoming-message-stream.js
@@ -26,6 +26,22 @@ class IncomingMessageStream extends Transform {
     this.bl = new BufferList();
   }
 
+  pause() {
+    super.pause();
+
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Thank you @chdh for preparing this, and sorry for having broken this with the 2.7.0 release. Do you have a suggestion how we could add a test to ensure we don't regress on this again on the future?

arthurschreiber:
@chdh FYI, I tweaked the implementation a tiny bit.

chdh:
@arthur Thanks for your fix. This is the better solution. And thanks for merging so quickly.

PR #887, opened 2019-04-08
fix: connection failure using Managed Service Identity (MSI) token due to the wrong connection type 

MichaelSun90:
Using Managed Service Identity (MSI) token forming connection will cause an error. It is caused by a logic flaw within the connection.js. This PR is fixing this specific error.


MichaelSun90:
Submitted another one with the same fix. Made a mistake with this one to include multiple unwanted commits.

Issue #876, opened 2019-03-14
update supported Nodejs versions to active LTS releases


jimmont:
Please update the supported versions of Nodejs to active LTS release lines. In particular tedious version 6.

As of May 2019 [active LTS are](https://github.com/nodejs/Release): **Nodejs version 8 and version 10**.

According to [the survey](https://github.com/tediousjs/tedious/issues/690) this could impact as much as 57% of users as they're reported as using a version older than Nodejs v8.

Changes to the project can follow if and when it makes sense to do so (ie passively).

Reasoning: it doesn't make sense to target unsupported runtimes and older versions of this product can be used as-necessary. Targeting active LTS allows using specific features that improve code quality and testing against specific targets reduces testing cost.

arthurschreiber:
As specified in https://github.com/tediousjs/tedious/blob/master/package.json#L36-L38, tedious currently supports Nodejs versions 6 and up. 6 is the current LTS version until the end of April 2019, and we'll probably keep supporting it until then at least.

Thank you for the heads up, but we'll be staying compatible with 6.x at least to the end of April. There's no rush to kick out support right the moment Nodejs 6.x becomes unsupported. 😄

jimmont:
The suggestion is to support LTS releases. The suggestion is not specific to any version, other than those which are active LTS releases. If adopted it would no longer be necessary to publish specific support other than "active LTS releases" and then update the package content to reflect the active LTS release whenever it is updated. Supporting version 6 of Nodejs would be included until the end of April as you've suggested. It seems realistic and reasonable to limit support to what is explicitly tracked as actively supported in the runtime and not outside of that, hence the suggestion. This appears to be lacking from this project, hence the suggestion. It takes some effort to move to different runtimes and making customers aware of transitions is helpful to them. So your calling out version 6 is actually in conflict with this because it fails to consider that this is within a matter of weeks where a given runtime will no longer be formally supported.

arthurschreiber:
I just opened https://github.com/tediousjs/tedious/pull/905 which will drop support for Node.js 6.x.

PR #896, opened 2019-05-08
feat: add integrated authentication support for Azure SQL databases

MichaelSun90:
This implements two new authentication methods specifically targeted at Azure SQL databases, namely `azure-active-directory-msi-vm` and `azure-active-directory-msi-app-service`.

Fixes: https://github.com/tediousjs/tedious/issues/918


src/connection.js
```diff
@@ -67,22 +71,22 @@ class Connection extends EventEmitter {
...
       const options = config.authentication.options === undefined ? {} : config.authentication.options;
 
-      if (typeof type !== 'string') {
+      if (typeof authType !== 'string') {
         throw new TypeError('The "config.authentication.type" property must be of type string.');
       }
 
-      if (type !== 'default' && type !== 'ntlm' && type !== 'azure-active-directory-password' && type !== 'azure-active-directory-access-token') {
-        throw new TypeError('The "type" property must one of "default", "ntlm", "azure-active-directory-password" or "azure-active-directory-access-token".');
+      if (authType !== 'default' && authType !== 'ntlm' && authType !== 'azure-active-directory-password' && authType !== 'azure-active-directory-access-token' && authType !== 'azure-active-directory-MSI') {
```
arthurschreiber: How do you feel about changing the authentication type to `azure-active-directory-msi` (all lower case) to match the naming of the other authentication types more closely?

src/connection.js
```diff
@@ -1274,6 +1306,95 @@ class Connection extends EventEmitter {
...
+    var tokenPath = '';
+    var tokenPort = '';
+    var tokenHeaders = { 'Metadata': true };
+    // Api version supported on Azure VM
+    var apiVersion = '2018-02-01';
+
+    // If MSI_ENDPOINT and MSI_SECRET are defined, used these instead of the hard-coded host
+    if (undefined !== msiEndpoint && undefined !== msiSecret) {
+      // Parse the environment variable as a url and set corresponding request components
+      var appUrl = url.parse(msiEndpoint, true);
```
arthurschreiber: What should happen if the URL in the `MSI_ENDPOINT` is misformed? 🤔

src/connection.js
```diff
@@ -1274,6 +1306,95 @@ class Connection extends EventEmitter {
...
+      tokenHost = appUrl.hostname;
+      tokenPath = appUrl.pathname;
+      tokenPort = appUrl.port;
+      tokenHeaders = {
+        'Secret': msiSecret
+      };
+      // Api version supported on Azure App
+      apiVersion = '2017-09-01';
+      clinetIdKeyName = '&clientid=';
```
arthurschreiber: I think this might be a typo? You probably meant to call this `clientIdKeyName` 😬

src/connection.js
```diff
@@ -1274,6 +1306,95 @@ class Connection extends EventEmitter {
...
+      port: tokenPort,
+      method: 'GET',
+      headers: tokenHeaders
+    };
+    return options;
+  }
+
+  getMSIToken() {
+    // Get the token from the azure source
+    http.get(this.getRequestOptions(), (tokenReponse) => {
```
arthurschreiber: What happens if the http request fails (not an endpoint error, but a transport error)? 🤔

You probably need to handle errors by adding an `error` handler to the `ClientRequest` object that is returned by the `http.get` call. See the documentation for `http.get` for an example.

src/connection.js
```diff
@@ -1274,6 +1306,95 @@ class Connection extends EventEmitter {
...
+
+      // Keep getting the packages
+      tokenReponse.on('data', (chunk) => {
+        data += chunk;
+      });
+
+      // Done getting the packages
+      tokenReponse.on('end', () => {
+        // If the response data contains an error or empty
+        if (-1 !== data.indexOf('error') || '' === data) {
```
arthurschreiber: In cases where the MSI endpoint returns an error, isn't there a proper JSON response returned? Checking the response for an the word `error` does not seem to be a very stable approach (just imagine the same characters randomly turning up in the token or similar).

src/connection.js
```diff
@@ -1274,6 +1306,95 @@ class Connection extends EventEmitter {
...
+          this.loginError = ConnectionError('Security token could not be authenticated or authorized.', 'EFEDAUTH');
+          this.emit('connect', this.loginError);
+          this.transitionTo(this.STATE.FINAL);
+          return;
+        }
+
+        this.sendMSIFedAuthResponsePacket(JSON.parse(data));
```
arthurschreiber: What happens if the response is malformed and JSON parsing fails? 🤔

arthurschreiber: And what happens if it does not contain a `access_token` key?

src/connection.js
MichaelSun90: I could have a check here for that, but these two ENV should come with the Function APP. The process for these two to show up should be automatic. At least from my current understanding for the Azure.

src/connection.js
MichaelSun90: Agreed, should think deeper on this.  By the way, can we expose the error message in detail or we can just use a general error message to hide the original error message for security reasons?

src/connection.js
MichaelSun90: I have added a try-catch block for the parse function.  Just wondering, is it ok that we expose a detail error message here? I saw some other method indeed just return a pre-set error message like "Security token could not be authenticated or authorized. "

MichaelSun90: Check the documentation for url.parse, it supposes throw two kinds of error: URI and Type. I added a type check for Type error, so I can represent a corresponding error message. For other errors, I will just return what error message the error provided. 
In terms of the URI, I tried a few different cases, and parse function never throw this type of error.  I also checked the parsing code, and I did not find a place that throws this error. Maybe, it will be thrown in a newer version?

Michael5011st:
Hello, any updates on this feature. Thank you for your time and have a nice day

src/connection.js
```diff
@@ -98,86 +96,78 @@ class Connection extends EventEmitter {
         if (options.password !== undefined && typeof options.password !== 'string') {
           throw new TypeError('The "config.authentication.options.password" property must be of type string.');
         }
-      }
 
-      authentication = { type: authType };
-      switch (authentication.type) {
```
MichaelSun90: Hi @arthurschreiber, I got this part form Hadis. Before she left, she did working on the MSI authentication for a while, and created this switch statement instead of set the similar object for every authentication types using multiple levels of if and else.  In my option, I think when we only have three authentication types, it makes sense that we only use if and else to handle different cases. However, now there are more types, and some of them do share some common config.authentication.options setups, I think the switch statement may actually help. Saved us some readily used if statements, and config initialization code.

funkydev:
Why don't you use [`ms-rest-azure`](https://www.npmjs.com/package/ms-rest-azure) library for authenticating with MSI and other types of the authentication?

The library is developing by the Azure team, so you will have a guarantee that all changes from azure have been included, and it has an effortless usage, which I have described [here](https://github.com/tediousjs/tedious/issues/918#issuecomment-508221198).

In my opinion, the logic for getting access token should be written using that library. Tedious should be responsible for validating passed auth configuration and passing/refreshing tokens to the connection.

arthurschreiber:
@funkydev I think you're right. We probably should also break up the MSI auth into two separate authentication types, `azure-active-directory-msi-vm` and `azure-active-directory-msi-app-service`, because their supported options are totally different.

funkydev:
@arthurschreiber You're right as well. I suppose someone should rewrite `azure-active-directory-password` authentication using `ms-rest-azure` to clean up the project. Then it will be much easier to implement the next types.

I'm able to help, but I don't fully understand how does it work.

src/login7-payload.js
```diff
@@ -60,7 +60,8 @@ const FEDAUTH_OPTIONS = {
   FEDAUTH_YES_ECHO: 0x01,
   FEDAUTH_NO_ECHO: 0x00,
   ADAL_WORKFLOW_USER_PASS: 0x01,
-  ADAL_WORKFLOW_INTEGRATED: 0x02
+  ADAL_WORKFLOW_INTEGRATED: 0x02,
+  ADAL_WORKFLOW_MSI: 0x03
```
arthurschreiber: @MichaelSun90 Is this constant correct? I added this because the Java driver uses this value when MSI authentication is used, but I don't see this value described in the latest version of the MS-TDS specification.

I _suspect_ this is simply implemented incorrectly in the Java driver, the correct value would be `0x02`, but that it works anyway because the SQL Server is treating this as a bitfield.

Could you maybe verify this with the SQLServer team? 🙇

src/connection.js
arthurschreiber: Hey @MichaelSun90 👋 

I understand where you where going with this, but there were some issues with the changes for this (like some validations going missing) and the diff for the PR was becoming hard to review because these changes are a bit unrelated to the new functionality.

I think we can and should tackle this again in the future, but not in this PR. 🙇

arthurschreiber:
@MichaelSun90 Thank you very much for the work so far! I updated this PR with some changes, and I'd be really happy if you could take a look and let me know what you think! 🙇 

@funkydev Thank you for the suggestion to switch to `ms-rest-azure`, that helped clean up some things, and should make it easier to add more authentication types in the future (if there's demand for that)! 🙇

src/connection.js
MichaelSun90: Hi @arthurschreiber, Thanks for making all the changes recently. :thumbsup:  It makes sense to me to separate this into a future PR since it indeed not closely related to the MSI authentication feature itself.

src/login7-payload.js
MichaelSun90: Hi @arthurschreiber, will look into this, and get back to asap you if I find anything. Thanks. :bow:

MichaelSun90: Hi @arthurschreiber,
I do find that JDBC has a value for MSI workflow and it is 0x03. 

    static final byte ADALWORKFLOW_ACTIVEDIRECTORYPASSWORD = 0x01;
    static final byte ADALWORKFLOW_ACTIVEDIRECTORYINTEGRATED = 0x02;
    static final byte ADALWORKFLOW_ACTIVEDIRECTORYMSI = 0x03;

.\src\main\java\com\microsoft\sqlserver\jdbc\SQLServerConnection.java

arthurschreiber: @MichaelSun90 Yes, that's where I got the value from. 😄 

But this value is not documented in the MS-TDS documentation, and I'm wondering whether the correct value is actually `0x02`, and using `0x03` just works by chance?

funkydev: @MichaelSun90 Where did you found that? 

I have found documentation of SQLClient in .Net [here](https://referencesource.microsoft.com/#system.data/system/data/sqlclient/TdsEnums.cs,230), and it contains only three workflows:

```c#
  public const byte ADALWORKFLOW_ACTIVEDIRECTORYPASSWORD = 0x01;
  public const byte ADALWORKFLOW_ACTIVEDIRECTORYINTEGRATED = 0x02;
  public const byte ADALWORKFLOW_ACTIVEDIRECTORYINTERACTIVE = 0x03;
```

I'm trying to find the info about any documentation for ADAL workflows, but I can't see anything

src/login7-payload.js
```diff
@@ -392,7 +393,17 @@ class Login7Payload {
...
+
+          switch (fedAuth.workflow) {
+            case 'integrated':
+              buffer.writeUInt8(FEDAUTH_OPTIONS.ADAL_WORKFLOW_INTEGRATED, 6);
+              break;
+
+            case 'default':
+              buffer.writeUInt8(FEDAUTH_OPTIONS.ADAL_WORKFLOW_USER_PASS, 6);
+              break;
+          }
```
funkydev: It should throw the `Unrecognized Authentication type for fedauth ADAL request` by default, like in [SQLClient](https://referencesource.microsoft.com/#System.Data/fx/src/data/System/Data/SqlClient/TdsParser.cs,7029) in C#

```C#
switch (fedAuthFeatureData.authentication) {
    case SqlAuthenticationMethod.ActiveDirectoryPassword:
        workflow = TdsEnums.ADALWORKFLOW_ACTIVEDIRECTORYPASSWORD;
        break;
    case SqlAuthenticationMethod.ActiveDirectoryIntegrated:
        workflow = TdsEnums.ADALWORKFLOW_ACTIVEDIRECTORYINTEGRATED;
        break;
    case SqlAuthenticationMethod.ActiveDirectoryInteractive:
        workflow = TdsEnums.ADALWORKFLOW_ACTIVEDIRECTORYINTERACTIVE;
        break;
    default:
        Debug.Assert(false, "Unrecognized Authentication type for fedauth ADAL request");
        break;
}
```

src/login7-payload.js
MichaelSun90: Hi @arthurschreiber and @funkydev, just confirmed this with David, there should not be an ADAL_WORKFLOW_MSI existed at all because that MSI authentication is just a token authentication. 
 
According to David: "With MSI, we've already obtained the security token from the MSI endpoint. There is no fedauth workflow. The reason for fedauth is so that the server can respond back with an authentication URL for a federated authentication provider. Meaning CompanyA uses Azure Active Directory, but they maintain their own AD servers. When AAD authentication is used, the server responds with a URL that points to the company's federated AD server, not the Azure server."

Also, I checked the Tedious code, We do separate this into different cases: 

`buildFeatureExt() {
    const buffers = [];

    const fedAuth = this.fedAuth;
    if (fedAuth) {
      switch (fedAuth.type) {
        case 'ADAL':
          const buffer = Buffer.alloc(7);
          buffer.writeUInt8(FEDAUTH_OPTIONS.FEATURE_ID, 0);
          buffer.writeUInt32LE(2, 1);
          buffer.writeUInt8((FEDAUTH_OPTIONS.LIBRARY_ADAL << 1) | (fedAuth.echo ? FEDAUTH_OPTIONS.FEDAUTH_YES_ECHO : FEDAUTH_OPTIONS.FEDAUTH_NO_ECHO), 5);
          buffer.writeUInt8(fedAuth.workflow === 'integrated' ? 0x02 : FEDAUTH_OPTIONS.ADAL_WORKFLOW_USER_PASS, 6);
          buffers.push(buffer);
          break;

        case 'SECURITYTOKEN':
          const token = Buffer.from(fedAuth.fedAuthToken, 'ucs2');
          const buf = Buffer.alloc(10);

          let offset = 0;
          offset = buf.writeUInt8(FEDAUTH_OPTIONS.FEATURE_ID, offset);
          offset = buf.writeUInt32LE(token.length + 4 + 1, offset);
          offset = buf.writeUInt8((FEDAUTH_OPTIONS.LIBRARY_SECURITYTOKEN << 1) | (fedAuth.echo ? FEDAUTH_OPTIONS.FEDAUTH_YES_ECHO : FEDAUTH_OPTIONS.FEDAUTH_NO_ECHO), offset);
          buf.writeInt32LE(token.length, offset);

          buffers.push(buf);
          buffers.push(token);

          break;
      }
    }

    buffers.push(Buffer.from([FEATURE_EXT_TERMINATOR]));

    return Buffer.concat(buffers);
  }
`

For MSI, we use securitytoken as the fedAuth.type.

funkydev: @MichaelSun90 it sounds like I supposed and I mentioned in issue #918. Please remember that access tokens have an expiration time and Tedious should renew the token if it is needed. 

Access tokens in AAD also have `audience` field inside, and it should be set as `https://database.microsoft.com/` to work correctly. The payload of the token has been described here: https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens (see `aud` description).

The methods from the lib what you've used ([`ms-rest-nodeauth`](https://github.com/Azure/ms-rest-nodeauth)) contain a property `options.tokenAudience`. That's the place where we should pass `https://database.microsoft.com/` value.

src/login7-payload.js
arthurschreiber: @funkydev The correct type is already enforced via the configuration. 😄

src/login7-payload.js
MichaelSun90: @arthurschreiber I think we can remove the line ADAL_WORKFLOW_MSI: 0x03, since that we do not really have a workflow for MSI. If you are busy, I can make a commit for this if we agree on this. 
@funkydev  I think the plan is that we do not handle the token refresh process on Tedious side. The current behavior is every time when there is a connection created, the get token logic will be called, and the behavior on the azure side should keep returning the current token until the current token got expired.

lenisha:
Tested with very simple node app on Azure App Service: https://github.com/lenisha/node-msi/blob/master/index.js 
But got errors:
```
ot request / to test add data to database via MSI AccessToken authentication.connect to database in request /add with config: {"server":"node-srv.database.windows.net","options":{"database":"","encrypt":true,"port":1433,"rowCollectionOnRequestCompletion":true,"useColumnNames":true},"authentication":{"type":"azure-active-directory-msi-app-service"}}
request database connection debugtext: connected to node-srv.database.windows.net:1433request database connection debugtext: State change: Connecting -> SentPrelogin
request database connection debugtext: State change: SentPrelogin -> SentTLSSSLNegotiation
request database connection debugtext: TLS negotiated (ECDHE-RSA-AES256-GCM-SHA384, TLSv1/SSLv3)request database connection debugtext: State change: SentTLSSSLNegotiation -> SentLogin7Withfedauth
request database connection debugtext: State change: SentLogin7Withfedauth -> ReRoutingrequest database connection debugtext: Packet size changed from 4096 to 4096request database connection debugtext: connection to node-srv.database.windows.net:1433 closed
request database connection debugtext: Rerouting to c7b6ed7d23ce.tr450.eastus2-a.worker.database.windows.net:11016request database connection debugtext: State change: ReRouting -> Connectingrequest database connection debugtext: connected to node-srv.database.windows.net:1433request database connection debugtext: State change: Connecting -> SentPrelogin
request database connection debugtext: State change: SentPrelogin -> SentTLSSSLNegotiation
request database connection debugtext: TLS negotiated (ECDHE-RSA-AES256-GCM-SHA384, TLSv1/SSLv3)request database connection debugtext: State change: SentTLSSSLNegotiation -> SentLogin7Withfedauth
request database connection debugtext: State change: SentLogin7Withfedauth -> SentLogin7WithStandardLogin
>>> Got connection azure sql server via tedious in request, next go to execute sql query statementConnected
request database connection debugtext: Requests can only be made in the LoggedIn state, not the SentLogin7WithStandardLogin stateError performing select:
{ RequestError: Requests can only be made in the LoggedIn state, not the SentLogin7WithStandardLogin stateat RequestError (D:\home\site\wwwroot\node_modules\tedious\lib\errors.js:32:12)at Connection.makeRequest (D:\home\site\wwwroot\node_modules\tedious\lib\connection.js:1728:24)at Connection.execSql (D:\home\site\wwwroot\node_modules\tedious\lib\connection.js:1507:10)at executeStatement (D:\home\site\wwwroot\index.js:90:14)at Connection.<anonymous> (D:\home\site\wwwroot\index.js:54:9)at emitOne (events.js:116:13)at Connection.emit (events.js:211:7)at Connection.message (D:\home\site\wwwroot\node_modules\tedious\lib\connection.js:2045:18)at Connection.dispatchEvent (D:\home\site\wwwroot\node_modules\tedious\lib\connection.js:1211:36)at MessageIO.messageIo.on (D:\home\site\wwwroot\node_modules\tedious\lib\connection.js:1084:14)message: 'Requests can only be made in the LoggedIn state, not the SentLogin7WithStandardLogin state',code: 'EINVALIDSTATE' }
```

redhatva:
What is the status on this? I am getting a "Login failed for user 'NT AUTHORITY\\ANONYMOUS LOGON"

const config = {
		"authentication": {
			"type": "azure-active-directory-msi-app-service"
		},
    "server": dbServer,
    "options": {
        "database": dbDatabase,
        "encrypt": true,
    }
}

funkydev:
@redhatva Make sure that you adequately configured SQL Server, especially that you have granted SQL Database access to the managed identity (or app service). You can read more about that [here](https://docs.microsoft.com/pl-pl/azure/app-service/app-service-web-tutorial-connect-msi#grant-permissions-to-managed-identity).

Anyway, I **do not recommend to use integrated authentication onto production environments** because the Node.js **application dramatically slows down**.

ckittel:
> do not recommend to use integrated authentication onto production environments because the Node.js application dramatically slows down.

@funkydev is there an issue to track this observation and watch for any mitigations?  Does this happen in all uses or only select situations? Related to caching (or not) of the token?

funkydev:
@ckittel I don't know if there is any issue for that. Some time ago, I just added the Service Principal Secret Authentication (#990) for the library (just another AAD authentication method). 

@arthurschreiber noticed that the `@azure/ms-rest-nodeauth` has abysmal performance. You can read all the discussion in  PR #990. In my opinion, it is related to authenticating and re-opening connections, but it is just an opinion.

Authors of `@azure/ms-rest-nodeauth`  library tell that they are dropping support for it and recommend to use `@azure/identity` if it is possible. You can read more about that [here](https://github.com/Azure/ms-rest-nodeauth/pull/75#issuecomment-519198541)

At the moment `tedious` library supports Node v.6.x, so there is no chance to update it because the minimum required Node is version 8.x. Anyway, there is an issue #876 for update supported version of Node to current LTS releases and the pull request #905, which drops support for Node 6.x. After merging it, someone can rewrite AAD authentication methods to use `@azure/identity` library.

PR #905, opened 2019-05-26
Drop support for Node.js 6.x and 8.x

arthurschreiber:
This drops support for Node.js 6.x and 8.x. It removes these versions from builds on AppVeyor and Travis CI and updates our babel configuration to target Node.js 10.x and newer.

This is a breaking change and will be released with major version bump.

Fixes: https://github.com/tediousjs/tedious/issues/876


arthurschreiber:
https://github.com/tediousjs/tedious/pull/1202 removed all current deprecation messages, so I updated this PR and will probably merge it next.

Issue #889, opened 2019-04-10
TVP TYPE.date not working


marianolop22:
Hi, i'm trying to send a TVP with one column type.date value.
Code example

this is my function
```js
function set(req, res) {
    var params = req.body;
    var table = {
      columns: [
        {name: 'id_service', type: TYPES.Int },
        {name: 'id_office', type: TYPES.NVarChar },
        {name: 'disabled_days', type: TYPES.NVarChar },
        {name: 'creation_date', type: TYPES.Date},
        {name: 'creation_user', type: TYPES.NVarChar}
      ],
      rows: [
      ]
    };

    for ( var i = 0; i < params.office.length; i++ ) {
        var office = [ 
            Number (params.office[i].idService ),
            params.office[i].idOffice,
            params.office[i].disabledDays,
            datesService.getTodayToSql(), //today's date in format yyyy/mm/dd '2019/04/10'
            'user'
        ];
        table.rows.push (office);
    }

    var param = {
        fecha : {
            type: TYPES.Date, value:datesService.getTodayToSql() //today's date in format yyyy/mm/dd '2019/04/10'
        }
    }

    conn.executeSPTVP ( "set_office_closed", table, param )
    .then (response =>{
        res.status(200).json({ doc:response });
    })
    .catch (error => {
        res.status(400).json({ message: 'error', result:error });
    });
}

exports.executeSPTVP = function(storedProcedure, table, params) {
    return new Promise((resolve, reject) => {
        var jsonArray = [];
        connService.acquire(function(err, connection) {
            if (err) {
                console.error(err);
                return;
            }
            let request = new Request(storedProcedure, function(err, rowCount ) {
                if (err) {
                    reject(err);
                } else {
                    resolve(jsonArray);
                }
                connection.release();
            });

            request.on('row', function(columns) {
                var obj = {};
                for (var property in columns) {
                    obj[property] = columns[property].value;
                }
                jsonArray.push(obj);
            });

            request.addParameter('tvp', TYPES.TVP, table); //add the table as parameter
            if (params) {
                for (var param in params) {
                    request.addParameter(param, params[param].type, params[param].value); //add fecha as parameter
                } 
            }
            connection.callProcedure(request);
        });
    });
}
```

when I print creation_date in sql store procedure, it returns `0001-01-01T00:00:00.000Z //bad`
when I print fecha in sql store procedure, it returns `2019-04-10T00:00:00.000Z //ok`
these are my dependencies

```json
  "dependencies": {
[...]

PR #907, opened 2019-06-03
fix: validate date input format before further process

MichaelSun90:
Fix the issue mentioned in #889. 
If the input Date format is 'YYYY/MM/DD', the current logic directly call math floor on this value will return a wrong date as output. 
Added some validation logic before pass through the input value. Also, added some error check for invalid date error. 

Added test cases for these handled scenarios


test/unit/data-type.js
```diff
@@ -85,3 +85,56 @@ exports.nanoSecondRounding = function(test) {
...
   test.done();
 };
+
+exports.dateInputWithSlashSeparator = function(test) {
+  var type = TYPES.typeByName.Date;
+  for (const [value, expectedBuffer] of [
+    ['2015/06/18 23:59:59', Buffer.from('03163a0b', 'hex')],
+  ]) {
+    var buffer = new WritableTrackingBuffer(16);
+    type.writeParameterData(buffer, { value: value }, { useUTC: false });
```
arthurschreiber: Due to the changes introduced via https://github.com/tediousjs/tedious/pull/845, calling `writeParameterData` now requires passing a callback function as the last parameter. I think it'd be enough to pass an empty callback function (like `() => {}`) to fix the test failures here.

MichaelSun90: On more question, why the other test cases called this function does not show this 
symptom？

IanChokS:
Verified this PR works with the up-to-date master branch (v6.3.0) -> I merged this branch with the latest master branch of tedious and ran the data-type.js test and it all passes. 
@arthurschreiber  I think this PR is good to merge.

IanChokS:
[Lines 389 - 392 ](https://github.com/MichaelSun90/tedious/blob/ac897edc720c6e57ff5c47157968f7a64ae50f32/src/bulk-load.ts#L389-L392) in _src/bulk-load.ts_ needs to be updated. 'types' does not have 'validate' method.

src/data-types/date.ts
```diff
@@ -23,14 +23,18 @@ const Date: DataType = {
     const value = parameter.value as any; // Temporary solution. Remove 'any' later.
 
     if (value != null) {
+      const time = new globalDate(value);
+      if ('Invalid Date' === time.toString()) {
+        throw new TypeError('Invalid date.');
+      }
```
arthurschreiber: Can this even happen? The `.validate` method returns a TypeError if an invalid date was passed. 🤔

src/bulk-load.ts
```diff
@@ -92,6 +92,10 @@ class RowTransform extends Transform {
 
     for (let i = 0; i < this.columns.length; i++) {
       const c = this.columns[i];
+      const error = c.type.validate(row[i]);
+      if (error instanceof TypeError) {
+        throw error;
+      }
```
arthurschreiber: This is a breaking change, because validation for bulk loads has never happened before. I agree that we should make this change, but in a separate PR so we can have the version number bumped.

src/data-types/date.ts
MichaelSun90: Is the 'validate' method will parse the input value first, then try to validate the date then?

src/bulk-load.ts
MichaelSun90: I will make a separate PR for this

test/unit/data-type.js
```diff
@@ -164,6 +164,19 @@ describe('Date', function() {
         assert.deepEqual(buffer, expectedBuffer);
       }
     });
+    it('dateInputWithSlashSeparator', () => {
+      const parameter = { value: '2015/06/18 23:59:59' };
+      const buffer = Buffer.concat([...TYPES.Date.generateParameterData(parameter, { useUTC: false })]);
+      assert.deepEqual(buffer, Buffer.from('03163a0b', 'hex'));
+    });
+    it('dateInputWithSlashSeparator', () => {
```
MichaelSun90: I will fix this, made a mistake in the test name

src/data-types/date.ts
MichaelSun90: I removed the unnecessary checks here and in dateTime as well.

src/bulk-load.ts
MichaelSun90: A new PR is made # 1065. I have reverted changes from this PR.

arthurschreiber:
I think with the changes in https://github.com/tediousjs/tedious/pull/1065 and `validateBulkLoadParameters` set to `true` this is no longer necessary? Please feel free to reopen if you think I'm mistaken. 🙇‍♂️

PR #924, opened 2019-07-16
missing bracket

UziTech:
missing bracket in getting started docs


arthurschreiber:
Thank you for these fixes! 🙇

PR #927, opened 2019-07-23
refactor: stand-alone examples

IanChokS:
Refactored some old examples files (in which doing so fixed some old bugs), and added a couple of new example files.


IanChokS:
Hi @arthurschreiber,

I tried looking into why the travis-ci test is failing but I am still unsure. Do you have any ideas?
Thanks

arthurschreiber:
@IanChokS Could you update all examples to follow the coding style used for the rest of the codebase? I.e. run `npx eslint examples --fix`. 🙇

IanChokS:
Done!

examples/TableValuedParameters.js
```diff
@@ -0,0 +1,151 @@
...
+
+  const request = new Request(sql, (err, rowCount) => {
+    if (err) {
+      throw err;
+    } else {
+      console.log(`'${table}' created!`);
+    }
+  });
+
+  request.on('requestCompleted', () => {
```
arthurschreiber: I think it'd be better not to advertise the `requestCompleted` callback too much, but instead use the callback passed to `Request` instead. 🤔

IanChokS: I moved these 'requesCompleted' section into the Request call backs :)

IanChokS:
@arthurschreiber I'm not exactly sure why codecov is failing the way it is. Do you have any ideas?

IanChokS:
This is continued in #991 for a cleaner commit history

PR #939, opened 2019-08-12
wip: feat: refactor AAD authentication method

funkydev:
It removes `@azure/ms-rest-nodeauth` and implements `@azure/identity` library.


funkydev:
@arthurschreiber would you like to review the current PR?

arthurschreiber:
@funkydev Thanks for opening this PR! 🙇 

As you can read up on https://github.com/Azure/ms-rest-nodeauth/pull/75#issuecomment-519198541, `@azure/ms-rest-nodeauth` is essentially deprecated and won't be developed further and is supposed to be replaced by `@azure/identity`.

I think we should migrate the existing authentication schemes over to `@azure/identity` first, before adding new ones. I tried doing this, and it turns out that `@azure/identity` does not always support passing in the same options as `@azure/ms-rest-nodeauth` did before, so we might have to either remove options or get the upstream library to support the options.

What do you think? 🙇

funkydev:
@arthurschreiber that is definitely good decision. We should consider if we really want to pass that credentials strings or just pass the reference to the `@azure/identity` Credential object (or their chains).

I'll do some research in this week and provide more details.

funkydev:
@arthurschreiber I have removed `@azure/ms-rest-nodeauth` library from `tedious` library, and I implemented authentication using `@azure/identity` credentials.

At the moment configuration for new connection looks like:
```javascript
  const {Connection} = require('tedious');

  let credentials = new ClientSecretCredential(tenantId, clientId, clientSecret);
  let connection = new Connection({
    server: DB_HOST,
    options: {
      encrypt: true,
      database: DB_NAME
    },
    authentication: {
      type: 'azure-identity-credentials',
      options: {credentials}
    }
  });
```

Now I'm going to add previously supported authentication types.

funkydev:
Currently, `@azure/identity` doesn't support interactive authorization flow. I have opened issue [#4774](https://github.com/Azure/azure-sdk-for-js/issues/4774) to implement that method.

funkydev:
`@azure/identity` doesn't support version 6 of node.js as you can read [here](https://www.npmjs.com/package/@azure/identity#prerequisites). 

PR #905 has to be merged first.

MichaelSun90:
Hi @arthurschreiber, I think you have made the similar changes in the other MSI authentication PR? If you confirmed that, then I think we can close this one for now.

MichaelSun90:
@arthurschreiber, I remember our discussion on @azure/identity. Should we close this one, and open a new one after we drop nodejs 6 support and start working on the modification for using azure/identity?

funkydev:
I closed this issue due to a discussion under Issue https://github.com/tediousjs/tedious/issues/1144.

It would be better to implement a more generic solution without injecting another dependency, which doesn't fully cover the current behavior.

PR #882, opened 2019-03-28
just a quick typo fix

tightightight:
**Before submitting a PR :**
1. Ensure your fork is created from `master` branch of [the repository](https://github.com/tediousjs/tedious).
2. Run `npm install` in the root folder.
3. After bug fix/code change, ensure all the existing tests and new tests (if any) pass (`npm run-script test-all`). During development, to run individual test use `node_modules/nodeunit test/<test_file.js> -t <test_name>`.
4. Build the driver (`npm run build`).
5. Run eslint and flow typechecker (`npm run lint`).
6. Run commitlint (`node_modules/.bin/commitlint --from origin/master --to HEAD`). Refer [commit conventions](http://marionebl.github.io/commitlint/#/concepts-commit-conventions) and [commit rules](http://marionebl.github.io/commitlint/#/reference-rules).

**Thank you for Contributing!**

PR #940, opened 2019-08-12
chore: license typo

IanChokS:
From @tightightight PR #882


arthurschreiber:
That should be a `chore` (does not trigger a release), not a `fix` (triggers a release). 🙇

IanChokS:
I will merge @tightightight PR #882 instead

PR #947, opened 2019-08-22
Emit cancel event AFTER setting the error

irgijs:
Without this PR, the only way for the cancel event listener to access the timeout error would be

```
request.on('cancel', (...args) => {
	setTimeout(() => {
		console.log('error', request.error);
	}, 1000);
});
```

With this pull request is applied, the undesirable timeout is no longer needed:
```
request.on('cancel', (...args) => {
	console.log('error', request.error);
});
```

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -1114,10 +1114,10 @@ class Connection extends EventEmitter {
   requestTimeout() {
     this.requestTimer = undefined;
     const request = this.request;
-    request.cancel();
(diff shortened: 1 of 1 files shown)


IanChokS:
Hi @irgijs,

Thanks for the PR! Could you include a test case for this change, as well as [changing your commit messages](https://help.github.com/en/articles/changing-a-commit-message) to follow the [commit lint rules](https://github.com/conventional-changelog/commitlint) (`git --amend`)?

Issue #948, opened 2019-08-23
Unreachable code: cancelling request before buffering finished


irgijs:
We noticed that a part of the code is not reachable when canceling a request before the buffering finished:

connection.js ('cancel'  eventListener)
```
if (!this.isRequestActive(request)) {
          // Cancel was called on a request that is no longer active on this connection
          return;
        } // There's three ways to handle request cancelation:


        if (this.state === this.STATE.BUILDING_CLIENT_REQUEST) {
          // The request was cancelled before buffering finished
          const sqlRequest = this.request;
          this.request = undefined;
          sqlRequest.callback(RequestError('Canceled.', 'ECANCEL'));
          this.transitionTo(this.STATE.LOGGED_IN);
        } 
```
The `isRequestActive`  function contains this:
```
 isRequestActive(request) {
    return request === this.request && this.state === this.STATE.SENT_CLIENT_REQUEST;
  }
```

Thus, if the state is BUILDING_CLIENT_REQUEST, the isRequestActive guard is never passed and the abovementioned code is never executed. Therefore, a request in the BUILDING_CLIENT_REQUEST cannot be properly cancelled.

PR #952, opened 2019-08-26
948 cancelling before buffering

IanChokS:
A fix for issue[ #948](https://github.com/tediousjs/tedious/issues/948)


src/connection.js
```diff
@@ -1693,19 +1693,15 @@ class Connection extends EventEmitter {
       let message;
 
       this.request.once('cancel', () => {
-        if (!this.isRequestActive(request)) {
```
arthurschreiber: Wouldn't it be better to have `isRequestActive` also return true if the state is `this.STATE.BUILDING_CLIENT_REQUEST`? 🤔

IanChokS: I was thinking that too, but I didn't see a Building Client Request state in the TDS docs so I wasn't too sure. Do you think I should add that as part of the `isRequestActive` method?

IanChokS:
This branch has a weird commit history. See #984 for the updated one

PR #957, opened 2019-08-29
Convert more pure JS files to TypeScript

arthurschreiber:
This converts another file containing internal functionality to TypeScript.


IanChokS:
How concerned should we be about the Node.js: 6 environment for the CI builds?

arthurschreiber:
> How concerned should we be about the Node.js: 6 environment for the CI builds?

In general, I think we should be concerned about build failures. 😬I checked the build logs and it looks like that's a flaky test failure. I restarted the failing job and that should hopefully "fix" this.

PR #966, opened 2019-09-05
Disable Azure Active Directory Password build job on Travis CI

arthurschreiber:
This disables the Azure Active Directory Password build job on Travis CI until we can fix the configuration and bring it back.


arthurschreiber:
@IanChokS thank you so much for solving that one! 🙌🏻 ❤️

PR #967, opened 2019-09-06
Clean up some value parsing code

arthurschreiber:
This slightly cleans up the value parsing code's use of `readTextPointerNull` and `readDataLength`.

`readTextPointerNull` is now only called for `Image`, `Text` and `NText` data type.

`readDataLength` is now only called for variable length data types.


IanChokS:
Travis-ci/push has passed after rerunning the Azure tests.

PR #972, opened 2019-09-19
Convert more files over to TypeScript

arthurschreiber:
This converts many of the remaining code files over to TypeScript. There's been a few non-type related changes to make the type information a bit easier to add, but overall the changes should be mostly straightforward. 😅


src/connection.d.ts
```diff
@@ -6,6 +6,7 @@ export type ConnectionOptions = {
   tdsVersion: string,
   camelCaseColumns: boolean,
   useColumnNames: boolean,
+  useUTC: boolean,
```
arthurschreiber: 👍

src/data-types/binary.ts
```diff
@@ -26,7 +26,7 @@ const Binary: { maximumLength: number } & DataType = {
     return 'binary(' + length + ')';
   },
 
-  resolveLength: function(parameter) {
+  resolveLength: function(parameter): number {
```
arthurschreiber: I don't think the return type needs to be repeated here. 🤔 

```suggestion
  resolveLength: function(parameter) {
```

src/data-types/bit.ts
```diff
@@ -14,7 +15,7 @@ module.exports = {
     buffer.writeUInt8(1);
   },
 
-  writeParameterData: function(buffer, parameter, options, cb) {
+  writeParameterData: function(buffer, parameter: ParameterData<null | unknown>, options, cb) {
```
arthurschreiber: Let's stick to any here, until the remaining code was streamlined and it's clearer what the parameter data type can be.

```suggestion
  writeParameterData: function(buffer, parameter: ParameterData<any>, options, cb) {
```

src/data-types/bit.ts
```diff
@@ -24,7 +25,7 @@ module.exports = {
     cb();
   },
 
-  validate: function(value) {
+  validate: function(value): null | boolean {
```
arthurschreiber: 👍

src/data-types/char.ts
```diff
@@ -1,20 +1,25 @@
...
+const Char: { maximumLength: number } & DataType = {
   id: 0xAF,
   type: 'BIGCHAR',
   name: 'Char',
   hasCollation: true,
   dataLengthLength: 2,
   maximumLength: 8000,
 
-  declaration: function(parameter) {
+  declaration: function(parameter: Parameter) {
```
arthurschreiber: I don't think you need to specify the type of `parameter` again.

```suggestion
  declaration: function(parameter) {
```

src/data-types/char.ts
```diff
@@ -1,20 +1,25 @@
...
+    //const value = parameter.value as null | string | { toString(): string };
+    const value = parameter.value as any; //Temporary solution. Remove 'any' later.
```
arthurschreiber: 👍

src/data-types/char.ts
```diff
@@ -27,7 +32,9 @@ module.exports = {
     }
   },
 
-  resolveLength: function(parameter) {
+  //ParameterData<any> is temporary solution. TODO: need to understand what type ParameterData<...> can be. 
+  resolveLength: function(parameter: ParameterData<any>): number {
```
arthurschreiber: I think you can leave off the return type:

```suggestion
  resolveLength: function(parameter: ParameterData<any>) {
```

src/data-types/char.ts
```diff
@@ -41,22 +48,24 @@ module.exports = {
     }
   },
 
-  writeTypeInfo: function(buffer, parameter) {
+  writeTypeInfo: function(buffer, parameter: ParameterData<Buffer | null>) {
```
arthurschreiber: Here as well, let's just use `any` as the generic parameter:

```suggestion
  writeTypeInfo: function(buffer, parameter: ParameterData<any>) {
```

src/data-types/date.ts
```diff
@@ -0,0 +1,58 @@
+import { DataType, ParameterData } from '../data-type';
+
+//This is to gain access to global objects in 'use strict' mode after typescript conversion to javascript. 
+var global = (0, eval)('this');
```
arthurschreiber: I don't think this workaround is required. 😬 In Node.js, there's always a `global` variable defined, that allows accessing the global object. This `var` declaration here shadows the _real_ `global`. If you just remove this line, the other references of `global` (and `global.Date`) will still continue working. 👍

arthurschreiber: As a bonus, your `global` variable has type `any`, while the proper `global` has a proper type and will allow proper typechecking for the usage of `global.Date`. 👍

IanChokS:
Thanks for the feedback! I agree, I think for now we should specify input type as 'any' for now as you've mentioned above, and circle back later once all the data-types have been converted to .ts.

src/data-types/date.ts
IanChokS: That's funny, I actually didn't realize 'global' was already defined in Node.js :sweat_smile:

PR #973, opened 2019-09-24
Speed up big-int parsing

SergeS:
Resolves speed bottleneck when there are a lot of bigint data coming from SQL server.

It tries to parse it into Number, and only if it will overflow MAX_SAFE_INTEGER it goes to string parsing

### DIFF
--- a/src/token/stream-parser.js
+++ b/src/token/stream-parser.js
@@ -1,4 +1,5 @@
 const Transform = require('readable-stream').Transform;
+const convertLEBytesToString = require('../tracking-buffer/bigint').convertLEBytesToString;
 const TYPE = require('./token').TYPE;
 
 const tokenParsers = {};
@@ -18,6 +19,9 @@ tokenParsers[TYPE.RETURNVALUE] = require('./returnvalue-token-parser');
 tokenParsers[TYPE.ROW] = require('./row-token-parser');
 tokenParsers[TYPE.NBCROW] = require('./nbcrow-token-parser');
 tokenParsers[TYPE.SSPI] = require('./sspi-token-parser');
+
+const pow2_32 = Math.pow(2, 32);
+const maxInt64HiVal = Math.pow(2, Math.floor(Math.log2(Number.MAX_SAFE_INTEGER)) - 32);
 
 module.exports = class Parser extends Transform {
   constructor(debug, colMetadata, options) {
@@ -62,6 +66,24 @@ module.exports = class Parser extends Transform {
     }
 
     done();
+  }
+
+  _convertUInt64(hiBits, lowBits) {
+    if (hiBits > maxInt64HiVal) {
+      return convertLEBytesToString(this.buffer.slice(this.position, this.position + 8));
+    }
+
+    return hiBits * pow2_32 + lowBits;
+  }
+
+  _convertInt64(hiBits, lowBits) {
+    const hiVal = hiBits & ~0x80000000;
+    if (hiVal > maxInt64HiVal) {
+      return convertLEBytesToString(this.buffer.slice(this.position, this.position + 8));
+    }
+    const isNegative = hiBits & 0x80000000 ? -1 : 1;
+
+    return (hiVal * pow2_32 + lowBits) * isNegative;
   }
 
   parseTokens() {
@@ -186,7 +208,7 @@ module.exports = class Parser extends Transform {
 
   readInt64LE(callback) {
     this.awaitData(8, () => {
-      const data = Math.pow(2, 32) * this.buffer.readInt32LE(this.position + 4) + ((this.buffer[this.position + 4] & 0x80) === 0x80 ? 1 : -1) * this.buffer.readUInt32LE(this.position);
+      const data = this._convertInt64(this.buffer.readUInt32LE(this.position + 4), this.buffer.readUInt32LE(this.position));
       this.position += 8;
       callback(data);
     });
@@ -194,7 +216,7 @@ module.exports = class Parser extends Transform {
 
   readInt64BE(callback) {
     this.awaitData(8, () => {
-      const data = Math.pow(2, 32) * this.buffer.readInt32BE(this.position) + ((this.buffer[this.position] & 0x80) === 0x80 ? 1 : -1) * this.buffer.readUInt32BE(this.position + 4);
+      const data = this._convertInt64(this.buffer.readUInt32BE(this.position), this.buffer.readUInt32BE(this.position + 4));
       this.position += 8;
       callback(data);
     });
@@ -202,7 +224,7 @@ module.exports = class Parser extends Transform {
 
   readUInt64LE(callback) {
     this.awaitData(8, () => {
-      const data = Math.pow(2, 32) * this.buffer.readUInt32LE(this.position + 4) + this.buffer.readUInt32LE(this.position);
+      const data = this._convertUInt64(this.buffer.readUInt32LE(this.position + 4), this.buffer.readUInt32LE(this.position));
       this.position += 8;
       callback(data);
     });
@@ -210,7 +232,7 @@ module.exports = class Parser extends Transform {
 
   readUInt64BE(callback) {
     this.awaitData(8, () => {
-      const data = Math.pow(2, 32) * this.buffer.readUInt32BE(this.position) + this.buffer.readUInt32BE(this.position + 4);
+      const data = this._convertUInt64(this.buffer.readUInt32BE(this.position), this.buffer.readUInt32BE(this.position + 4));
       this.position += 8;
       callback(data);
     });
--- a/src/value-parser.js
+++ b/src/value-parser.js
@@ -6,7 +6,6 @@ const guidParser = require('./guid-parser');
 const readPrecision = require('./metadata-parser').readPrecision;
 const readScale = require('./metadata-parser').readScale;
 const readCollation = require('./metadata-parser').readCollation;
-const convertLEBytesToString = require('./tracking-buffer/bigint').convertLEBytesToString;
 
 const NULL = (1 << 16) - 1;
 const MAX = (1 << 16) - 1;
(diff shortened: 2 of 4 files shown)


src/token/stream-parser.js
```diff
@@ -19,6 +20,9 @@ tokenParsers[TYPE.ROW] = require('./row-token-parser');
 tokenParsers[TYPE.NBCROW] = require('./nbcrow-token-parser');
 tokenParsers[TYPE.SSPI] = require('./sspi-token-parser');
 
+const pow2_32 = Math.pow(2, 32);
+const maxInt64HiVal = Math.pow(2, Math.floor(Math.log2(Number.MAX_SAFE_INTEGER)) - 32);
```
MichaelSun90: Hi @SergeS, Thanks for making this change. Just wondering, for this line, are you trying to extract the high bits for the MAX_SAFE_INTEGER, so you can use it later to compare with the inputs's high bits value? Is this the case, or I misunderstood your purpose for this?

SergeS: Yes, exactly, it is used to check high 32 bits of int64 later and depending on result either parse as number or as string

arthurschreiber:
@SergeS I have a set of changes that moves all custom `bigint` parsing to `jsbi` here:

https://github.com/tediousjs/tedious/compare/arthur/jsbi

I think we should introduce a new config option that would stop the conversion to `string`s for bigint and just return JSBI instances instead. What do you think?

SergeS:
Hi, that is good option. But in my case it may not work, as I need to fetch number from DB and send out in json, focused on speed. I will test it

arthurschreiber:
`JSBI` should mostly match the "native" `BigInt`s that are available in newer V8 versions, and I think that the `toString` method of `JSBI` objects should be equally fast, and definately faster than our custom string conversion code.

I can try and create a benchmark for us to use and compare. 👍

arthurschreiber:
Here's a very basic benchmark:

```js
var WritableTrackingBuffer = require('../../lib/tracking-buffer/writable-tracking-buffer');
var convertLEBytesToString = require('../../lib/tracking-buffer/bigint').convertLEBytesToString;

var JSBI = require('jsbi');

const jsbi32n = JSBI.BigInt(32);

const pow2_32 = Math.pow(2, 32);
const maxInt64HiVal = Math.pow(2, Math.floor(Math.log2(Number.MAX_SAFE_INTEGER)) - 32);

function convertInt64(buffer) {
  const hiBits = buffer.readUInt32LE(4);
  const lowBits = buffer.readUInt32LE(0);

  const hiVal = hiBits & ~0x80000000;
  if (hiVal > maxInt64HiVal) {
    return convertLEBytesToString(buffer.slice(0, 8));
  }
  const isNegative = hiBits & 0x80000000 ? -1 : 1;

  return (hiVal * pow2_32 + lowBits) * isNegative;
}

function convertLEBytesToStringWithJSBI(buffer) {
  return JSBI.add(
    JSBI.leftShift(
      JSBI.BigInt(
        buffer[4] +
        buffer[5] * 2 ** 8 +
        buffer[6] * 2 ** 16 +
        (buffer[7] << 24) // Overflow
      ),
      jsbi32n
    ),
    JSBI.BigInt(buffer.readUInt32LE(0))
  ).toString();
}

function convertLEBytesToJSBI(buffer) {
  return JSBI.add(
    JSBI.leftShift(
      JSBI.BigInt(
        buffer[4] +
        buffer[5] * 2 ** 8 +
        buffer[6] * 2 ** 16 +
        (buffer[7] << 24) // Overflow
      ),
      jsbi32n
    ),
    JSBI.BigInt(buffer.readUInt32LE(0))
  );
}

function convertLEBytesToBigInt(buffer) {
  return buffer.readBigInt64LE(0);
}

function convertLEBytesToNumber(buffer) {
  return Math.pow(2, 32) * buffer.readInt32LE(4) + ((buffer[4] & 0x80) === 0x80 ? 1 : -1) * buffer.readUInt32LE(0);
}

var Benchmark = require('benchmark');

var buffer = new WritableTrackingBuffer(50);
buffer.writeUInt64LE(Number.MAX_SAFE_INTEGER);

var data = buffer.data;

console.log(convertInt64(data));
console.log(convertLEBytesToString(data));
console.log(convertLEBytesToStringWithJSBI(data));
console.log(convertLEBytesToJSBI(data));
console.log(convertLEBytesToNumber(data));
console.log(convertLEBytesToBigInt(data));

var suite = new Benchmark.Suite();
suite.add('convertLEBytesToString', function() {
  convertLEBytesToString(data);
});
suite.add('JSBI toString()', function() {
  convertLEBytesToStringWithJSBI(data);
});
suite.add('convertInt64', function() {
  convertInt64(data);
});
suite.add('JSBI', function() {
  convertLEBytesToJSBI(data);
});
suite.add('Number', function() {
  convertLEBytesToNumber(data);
});
suite.add('BigInt', function() {
  convertLEBytesToBigInt(data);
});
suite.on('cycle', function(event) {
  console.log(String(event.target));
});
suite.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
});
suite.run();
```

And the result when running on my machine:

```
convertLEBytesToString x 645,822 ops/sec ±0.52% (94 runs sampled)
JSBI toString() x 1,460,602 ops/sec ±0.68% (86 runs sampled)
convertInt64 x 183,980,877 ops/sec ±0.52% (93 runs sampled)
JSBI x 11,764,772 ops/sec ±0.55% (93 runs sampled)
Number x 855,808,788 ops/sec ±0.94% (90 runs sampled)
BigInt x 6,870,208 ops/sec ±0.39% (93 runs sampled)
```

Based on this, by far the fastest implementation is just using  plain old `Number`s. Unfortunately, not all `bigint`s can be represented with `Number`, so that's not really a feasible solution.

Next comes the implementation @SergeS suggested, which uses `Number`s for values below `Number.MAX_SAFE_INTEGER`, and switches to strings for values that can not be represented with a `Number`. The benchmark did not test values bigger than `Number.MAX_SAFE_INTEGER`, I'd assume this would perform a lot worse once values would be returned as strings.

Next come `JSBI` and native `BigInt`, with `JSBI` actually being faster than native `BigInt`s 🤷‍♂ - the perf difference might actually be due to something in the `Buffer.readBigInt64LE` method being slower.

And at the very end, we have the implementations that return strings for large values, with the implementation based on `JSBI` being at least twice as fast compared to the one currently shipping with `tedious`.

arthurschreiber:
@SergeS What are your speed requirements here? Would be moving to string serialization via `JSBI` as a first step, with an option to return `JSBI` options directly as a second step be good enough?

The option that automatically switch between `number` or `string` based on the value seems to have best performance, but it's also confusing from an API perspective, as consumers would always have to check whether data was returned as number or string, and be ready to handle both these data types.

My favorite option would be to just return JSBI values for bigint types in the future, and have the user perform whatever conversion they need on their own.

SergeS:
Hi, yes, I'm aware about this drawback ... We actually use jsbi on front-end then, and it is able to catch it. Maybe we can introduce an option if we prefer speed or consistency? What is your opinion on this. Anyway we are working around with real time data. So any delay is bad for me.

arthurschreiber:
I did some additional benchmarking, this time less focused on the raw parsing performance and instead using a full "real-world" example, involving querying an actual server.

Here's the benchmarking code (I'll open a PR to add it to the repo):

```js
var tedious = require('../../lib/tedious');
var Request = tedious.Request;
var TYPES = tedious.TYPES;
var async = require('async');

var common = require('../common');

var connection;
common.createBenchmark({
  name: 'Many bigint rows',
  profileIterations: 10,

  setup: function(cb) {
    common.createConnection(function(_connection) {
      connection = _connection;

      var request = new Request('CREATE TABLE #benchmark ([number] bigint)', function(err) {
        if (err) return cb(err);

        async.timesSeries(10000, function(n, next) {
          var request = new Request('INSERT INTO #benchmark ([number]) VALUES (@number)', next);

          request.addParameter('number', TYPES.BigInt, Number.MAX_SAFE_INTEGER - n);

          connection.execSql(request);
        }, cb);
      });

      connection.execSqlBatch(request);
    });
  },

  exec: function(cb) {
    var request = new Request('SELECT * FROM #benchmark', cb);
    connection.execSql(request);
  },

  teardown: function(cb) {
    var request = new Request('DROP TABLE #benchmark', function(err) {
      if (err) {
        return cb(err);
      }

      connection.close();
    });
    connection.execSqlBatch(request);
  }
});
```

This essentially creates a table with a single `bigint` column and fills it with 10000 rows. I ran this benchmark against different implementations again:

---

Current `master`:

```
Many bigint rows x 43.16 ops/sec ±0.64% (68 runs sampled)
Memory: 45.21875 MiB
```

---

`SergeS/master` (return`number` / `string`, depending on the integer value)

```
Many bigint rows x 169 ops/sec ±2.51% (77 runs sampled)
Memory: 47.015625 MiB
```

---

`arthur/jsbi` (return `string` from `JSBI`)

```
Many bigint rows x 70.79 ops/sec ±7.72% (83 runs sampled)
Memory: 43.64453125 MiB
```

---

`arthur/jsbi` (modified to always return `JSBI`)

```
Many result rows x 144 ops/sec ±1.44% (82 runs sampled)
Memory: 42.38671875 MiB
```

arthurschreiber:
@SergeS so, here's what I plan to do, and would love to get your feedback on this:

* Bring `arthur/jsbi` into a mergeable state - this will remove a bunch of our custom bigint code and already provide a small performance boost without changing any user-noticeable behavior.
* Switch `bigint` result type from `string` to `JSBI`.
* Add an option to provide custom parsing functions for different data-types. This would allow you to provide your own parsing code that then would get passed in a `Buffer`, and would return whatever fits your application best.

This way, you could easily implement special result handling like the mixed `number` / `string` handling in whatever way fits your application best. 🤔

SergeS:
I think option to add manual parsing fits my needs :-)

... yet I ended up working on even faster C++ message parser within tedious

arthurschreiber:
> ... yet I ended up working on even faster C++ message parser within tedious

That sounds interesting! Looking forward to see this in action! 👍

MichaelSun90:
Hi @arthurschreiber, I think it is safe to close this PR since the JSBI change is merged.

PR #974, opened 2019-09-25
More Conversion

IanChokS:
- add more conneciton.option types
- js->ts datetime.ts, & datetime2.ts


src/data-types/datetime2.ts
```diff
@@ -1,14 +1,16 @@
+import { DataType, ParameterData } from '../data-type';
+
 const YEAR_ONE = new Date(2000, 0, -730118);
 const UTC_YEAR_ONE = Date.UTC(2000, 0, -730118);
 
-module.exports = {
+const DateTime2: DataType = {
```
arthurschreiber: If you change the definition here to be this:

```suggestion
const DateTime2: DataType & { resolveScale: NonNullable<DataType['resolveScale']> } = {
```

TypeScript won't think `resolveScale` might be undefined for this type - and that'll allow you to revert the checks you added to see of `this.resolveScale` is defined or not. 👍

src/data-types/datetime2.ts
```diff
@@ -53,8 +56,11 @@ module.exports = {
       } else {
         timestamp = ((time.getHours() * 60 + time.getMinutes()) * 60 + time.getSeconds()) * 1000 + time.getMilliseconds();
       }
-
-      timestamp = timestamp * Math.pow(10, parameter.scale - 3);
+      if(parameter.scale){
+        timestamp = timestamp * Math.pow(10, parameter.scale - 3);
```
arthurschreiber: In cases like this, where a value like `parameter.scale` is optional based on it's type definition but we know it's definately defined at this point in the code, you can add a `!` at the end to hint to TypeScript that the value is not null or undefined:

```suggestion
    timestamp = timestamp * Math.pow(10, parameter.scale! - 3);
```

That'll allow you to remove the `if`...`else` statement again.

arthurschreiber:
I updated https://github.com/tediousjs/tedious/pull/972 to fix a bunch of style related issues, so you'll probably want to merge the latest changes from there back into your branch.

Are you using VSCode? If yes, you can set up ESLint to run on `.ts` files as well and have that feedback in your editor as described here: https://javascriptplayground.com/typescript-eslint/#enabling-eslint-on-ts-files-in-vscode

src/connection.d.ts
```diff
@@ -3,11 +3,62 @@ import BulkLoad from './bulk-load';
 import { Metadata } from './metadata-parser';
 
 export type ConnectionOptions = {
-  tdsVersion: string,
-  camelCaseColumns: boolean,
-  useColumnNames: boolean,
-  useUTC: boolean,
+  abortTransactionOnError?: false,
```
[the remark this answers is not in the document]
IanChokS: Oh ok, I've changed the name ``` ConnectionOptions ``` to ```InternalConnectionOptions```  for future reference.

src/data-type.ts
```diff
@@ -71,7 +71,7 @@ export interface DataType {
   writeParameterData(buf: any, data: ParameterData, options: ConnectionOptions, callback: () => void) : void,
   validate(value: any) : any, //TODO: Refactor 'any' and replace with more specific type. 
 
-  dataLengthFromScale?: (scale: number) => number,
+  dataLengthFromScale?: (scale: number) => number | undefined,
```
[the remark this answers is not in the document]
IanChokS: Sorry I'm not sure what you mean. I've tried describing the output type (in datetime2 for e.g.,) but it won't accept it. 
image

image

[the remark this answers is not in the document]
arthurschreiber: I think this should look like:

```ts
  dataLengthFromScale?: (scale: number) => (number | undefined),
```

Maybe? 🤔 😅

src/data-types/datetime2.ts
```diff
@@ -80,7 +81,7 @@ module.exports = {
         buffer.writeUInt24LE(Math.floor((+parameter.value - UTC_YEAR_ONE) / 86400000));
       } else {
         const dstDiff = -(parameter.value.getTimezoneOffset() - YEAR_ONE.getTimezoneOffset()) * 60 * 1000;
-        buffer.writeUInt24LE(Math.floor((+parameter.value - YEAR_ONE + dstDiff) / 86400000));
+        buffer.writeUInt24LE(Math.floor((+parameter.value - +YEAR_ONE + dstDiff) / 86400000));
```
IanChokS: @arthurschreiber I just wanted to double check with you if this change is still valid. Without the '+', TypeScript shows an error

arthurschreiber: Yeah, this is correct. It's not an error in JavaScript itself, because JavaScript would automatically perform the conversion from `Date` to `number`, but TypeScript wants us to be explicit about this (to help catch potential bugs due to typos).

PR #977, opened 2019-09-30
refactor: convert data-types files to TypeScript

MichaelSun90:
Contains file type changes from js to ts for ntext.js - xml.js under data-types folder


src/data-types/numeric.ts
```diff
@@ -1,18 +1,19 @@
+import { DataType, ParameterData } from '../data-type';
 const NumericN = require('./numericn');
 
-module.exports = {
+const Numeric: & DataType = {
```
arthurschreiber: Is this a typo? 🤔

IanChokS: (The '&' is unnecessary)

MichaelSun90: oh, I will remove the &. I think I have few other &s need to be removed.

src/data-types/numeric.ts
```diff
@@ -1,18 +1,19 @@
+import { DataType, ParameterData } from '../data-type';
 const NumericN = require('./numericn');
 
-module.exports = {
+const Numeric:DataType = {
```
arthurschreiber: If you let TypeScript know that `resolvePrecision` and `resolveScale` are always defined on this datatype, you can get rid of all the `!` on `this.resolvePrecision` and `this.resolveScale`:

```suggestion
const Numeric: DataType & { resolveScale: NonNullable<DataType['resolveScale']>, resolvePrecision: NonNullable<DataType['resolvePrecision']> }= {
```

This is not super beautiful, but nicer than sprinkling the `!` in all these places. 🤷‍♂ Similar to `any`, `!` can hide bugs and thus I'd try to reduce it's usage as much as possible.

src/data-types/nvarchar.ts
```diff
@@ -29,23 +33,24 @@ module.exports = {
...
-        return parameter.value.toString().length || 1;
+        return value.toString().length || 1;
       }
     } else {
       return this.maximumLength;
     }
   },
 
-  writeTypeInfo: function(buffer, parameter) {
+  writeTypeInfo: function(buffer, parameter:ParameterData<any>) {
```
arthurschreiber: I think you don't have to specify `ParameterData<any>` here (and in other places) anymore:

```suggestion
  writeTypeInfo: function(buffer, parameter) {
```

src/data-types/smallint.ts
```diff
@@ -1,6 +1,8 @@
+import { DataType } from '../data-type';
+
 const IntN = require('./intn');
 
-module.exports = {
+const SmallInt: & DataType = {
```
arthurschreiber: I think there's still a leftover `&` here. 😬 

```suggestion
const SmallInt: DataType = {
```

src/data-types/text.ts
```diff
@@ -1,4 +1,6 @@
-module.exports = {
+import { DataType } from '../data-type';
+
+const Text: { hasTextPointerAndTimestamp:boolean, dataLengthLength: number } & DataType = {
```
arthurschreiber:
```suggestion
const Text: { hasTextPointerAndTimestamp: boolean, dataLengthLength: number } & DataType = {
```

src/data-types/smallmoney.ts
```diff
@@ -1,6 +1,8 @@
+import { DataType } from '../data-type';
+
 const MoneyN = require('./moneyn');
 
-module.exports = {
+const SmallMoney:DataType = {
```
arthurschreiber:
```suggestion
const SmallMoney: DataType = {
```

src/data-types/nvarchar.ts
MichaelSun90: Should we keep the parameter:ParameterData<any> for the places that use "parameter.value.length"? Otherwise, the parameter. the value here will be unknown, and it will complain about calling length based on unknow. 

or we should do: 
const value =parameter.value as any.  instead?

src/data-types/numeric.ts
MichaelSun90: wandering, can 'precision' and 'scale' 'possibly undefined issue' under 'writeTypeInfo', 'writeParameterData' be solved using this similar fashion?

PR #980, opened 2019-10-02
Use `jsbi` for internal 64bit integer calculations

arthurschreiber:
This switches the code that is used for doing internal 64bit number calculations from our own custom code over to `jsbi`. `jsbi` is based on the `BigInt` type available in Node.js 12 and newer, but works in all Node.js versions that we want to be supported by `tedious`.

### Context

Some parts of the TDS protocal use 64-bit integers. Examples for this are the `rowcount` in the `DONE` token, or everything related to the `bigint` data type.

The code that was used for this was adapted from different other libraries (e.g. `node-int64`), but was never optimized for speed nor correctness (see https://github.com/tediousjs/tedious/pull/973). By replacing this code with `jsbi`, we should be able to improve performance, reduce the code we need to maintain, and also ensure correctness.

Note: This only changes internals. All external code reading / writing to `bigint` columns will be unaffected (but should be much faster than before).


src/token/done-token-parser.js
```diff
@@ -30,7 +31,15 @@ function parseToken(parser, options, callback) {
...
           curCmd: curCmd
         });
-      });
+      };
+
+      if (options.tdsVersion < '7_2') {
+        parser.readUInt32LE(next);
+      } else {
+        parser.readBigUInt64LE((rowCount) => {
+          next(JSBI.toNumber(rowCount));
```
arthurschreiber: If more than `Number.MAX_SAFE_INTEGER` (`9007199254740991`) rows are transferred, this conversion might result in an incorrect rowcount. I'm not sure if that's something we have to care about - this limit seems hardly reachable. :shrug:

src/token/done-token-parser.ts
```diff
@@ -42,7 +44,15 @@ function parseToken(parser: Parser, options: InternalConnectionOptions, callback
...
           curCmd: curCmd
         });
-      });
+      };
+
+      if (options.tdsVersion < '7_2') {
+        parser.readUInt32LE(next);
+      } else {
+        parser.readBigUInt64LE((rowCount) => {
+          next(JSBI.toNumber(rowCount));
```
arthurschreiber: If more than `Number.MAX_SAFE_INTEGER` (`9007199254740991`) rows are transferred, this conversion might result in an incorrect rowcount. I'm not sure if that's something we have to care about - this limit seems hardly reachable. 🤷‍♂

PR #981, opened 2019-10-03
Value parser .ts

IanChokS:
- Changed value-parser.js -> value-parser.ts
- Add DataTypeN interface
- add new interface for MetaData


arthurschreiber:
@IanChokS Thanks for preparing this PR! ❤️ 🙇 

I took the chance to extend the changes you proposed with some cleanups:

* The `DataTypeN` type you proposed was removed and instead dummy methods were added to types like `BitN` to make them proper `DataType`. Now we only have `DataType`, which seems a bit more straightforward to me.
* The `sql_variant` values parsing code was completely refactored. Previously, that code was really hard to understand, especially the `isVariantValue` checking and the flow of how the variant value's `dataLength` was passed around. The new `sql_variant` value parsing is now a lot more straightforward. 👍 
* A lot of properties from `DataType` were dropped: `dataLengthFromScale`, `fixedDataLength`, `dataLengthLength`, `hasPrecision`, `hasScale`, `hasCollation`, `hasSchemaPresent` and `hasUDTInfo` were all removed. Instead of relying on these properties, the behaviour they triggered is now simply hardcoded inside the value parsing code.
* `MetaData` was changed back to `Metadata`, more in line with Microsoft's [Capitalization Conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

PR #988, opened 2019-10-08
refactor: random test failures refactor

MichaelSun90:
refactor some pervious mocha test which is causing some random failures during either travis and appveryor.


test/unit/instance-lookup-test.js
```diff
@@ -346,11 +344,9 @@ describe('parseBrowserResponse', function() {
...
-    new InstanceLookup().instanceLookup(options, () => {
-      assert.ok(spy.called, 'Failed to call dns.lookup on hostname');
-      assert.ok(spy.calledWithMatch(options.server), 'Unexpected hostname passed to dns.lookup');
-
-      done();
-    });
+    new InstanceLookup().instanceLookup(options, () => {});
+    assert.ok(spy.called, 'Failed to call dns.lookup on hostname');
+    assert.ok(spy.calledWithMatch(options.server), 'Unexpected hostname passed to dns.lookup');
+    done();
```
arthurschreiber: It looks like `done` is no longer called asynchronously, so it'd be better to make the tests synchronous by removing `done` from the test function completely. What do you think? 🙇

MichaelSun90: Sure thing. Is this behvaior change with the done function that happened in recendt mocha updates or it just behave differently with nodeunit version done function? 

By the way, there were some random failures happen within sender tests as well. I have reviewed them, and I have not identified any significant differences with the previous version.  The only thing I changed is moving one function to its corresponding test. 
I will paste the error logs here, see if you can share some insight on this. Thanks! 🙇

1) Sender send to hostnam
[00:01:14]        Send
[00:01:14]          "after each" hook:
[00:01:14]      Uncaught AssertionError: Unexpected hostname passed to dns.lookup: expected false to be truthy
[00:01:14]       at ok (test\unit/instance-lookup-test.js:335:14)
[00:01:14]       at callback (src/instance-lookup.ts:59:13)
[00:01:14]       at GetAddrInfoReqWrap.cb [as callback] (src/sender.ts:138:16)
[00:01:14]       at GetAddrInfoReqWrap.onlookupall [as oncomplete] (dns.js:79:17)
[00:01:14] 

[00:01:14] 
[00:01:14]   3) Sender send to hostnam
[00:01:14]        Send
[00:01:14]          "before each" hook in "Send":
[00:01:14]      TypeError: Cannot read property 'call' of undefined
[00:01:14]   
[00:01:14] 
[00:01:14]   4) Parallel Send Strategy
[00:01:14]        Send
[00:01:14]          "after each" hook for "should send all IPs success":
[00:01:14]      TypeError: Cannot read property 'testSockets' of undefined
[00:01:14]       at Context.testSockets (test\unit/sender-test.js:329:30)

MichaelSun90: Could these sender test failures also because there are done() function called during the process which is causing conflicts?

MichaelSun90:
Hi @arthurschreiber, Did a bit investigation on the random test failures under sender-test. Seems it is caused by an issue within dgram.js.  To fully avoid this random failure, according to this ticket: https://github.com/nodejs/node/issues/7061, I think we could either try to modify the sender.ts based on the workarounds mentioned in the ticket or checking whether the node side going to fix this.  People have created a PR: https://github.com/nodejs/node/pull/7066 for this fix, but the merging process for this PR got complicated on the node side.

PR #990, opened 2019-10-16
Add AAD Service Principal Secret authorization method

funkydev:
It adds Azure Active Directory *Service Principal Secret* Authorization method.

Usage:
```javascript
  server: 'your-azure-database-server',
  options: {
    database: 'database_name',
    encrypt: true
  },
  authentication: {
    type: 'azure-active-directory-service-principal-secret',
    options: {
      clientId: '', // SERVICE PRINCIPAL CLIENT ID
      clientSecret: '', // SERVICE PRINCIPAL CLIENT SECRET
      tenantId: '', // SERVICE PRINCIPAL TENANT ID (domain)
    }
  }
});
```

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -4,7 +4,7 @@ const os = require('os');
 const constants = require('constants');
 const { createSecureContext } = require('tls');
 
-const { loginWithUsernamePassword, loginWithVmMSI, loginWithAppServiceMSI } = require('@azure/ms-rest-nodeauth');
+const { loginWithUsernamePassword, loginWithVmMSI, loginWithAppServiceMSI, loginWithServicePrincipalSecret } = require('@azure/ms-rest-nodeauth');
 
 const BulkLoad = require('./bulk-load');
 const Debug = require('./debug');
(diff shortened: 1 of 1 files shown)


funkydev:
@arthurschreiber here is a small PR which adds support for AAD Service Principal Secret Auth.

Would you like to review the changes and merge this pull request? I saw that you're during the refactor process into TypeScript, but we need the service principal authentication as soon as possible, because our team wants to use that in production app.

arthurschreiber:
We're planning on moving away from `@azure/ms-rest-nodeauth` as soon as we drop Node.js 6.x support in favor of `@azure/identity`. `@azure/ms-rest-nodeauth` has absolutely abysmal performance, and I'm wary of adding a new auth method based on it.

Can this authentication method be mapped to `@azure/identity`, without changes to the config options?

funkydev:
Yes, using the `ClientSecretCredential` (more details are [here](https://azure.github.io/azure-sdk-for-js/identity/classes/clientsecretcredential.html#constructor))

`tenantId: string` - The Azure Active Directory tenant (directory) ID.
`clientId: string` - The client (application) ID of an App Registration in the tenant.
`clientSecret: string` - A client secret that was generated for the App Registration.

funkydev:
@arthurschreiber You right. Our application did dramatically slow down after we changed authentication using MSI. It had problems with longer CPU times after the change.

Someone should decide to drop support for Node 6.x because even the Node.js team does not maintain that version. You can check it [here](https://nodejs.org/en/about/releases/). Then we should rewrite azure authentication methods using `@azure/identity` – as I saw, they implemented interactive authentication as well, so they made progress.

PR #999, opened 2019-11-03
FIX No longer rely on removed data-type props

### LINKED ISSUE #998: Removal of hasScale prop on data-types breaks bulk-load
In [v6.5.0](https://github.com/tediousjs/tedious/tree/v6.5.0) data-types were converted to TS and as part of this the `hasScale` property from `DateTime2` has been removed.

This is breaking [bulk-load](https://github.com/tediousjs/tedious/blob/master/src/bulk-load.ts#L166-L170) which relies on that prop. This is causing tests to fail on the mssql library.

When using `BulkLoad` with `DataType`s that used to have a `hasScale` property, the following error is now thrown: `Received an invalid column length from the bcp client for colid x`

dhensby:
Fixes #998 

This is a slightly different approach to fixing the problem than reinstating the `hasScale` flag (and other removed flags) from the `DataType` type; instead we infer the need to use `resolvePrecision` and `resolveScale` from the fact they are present.

This fixes a bug with `BulkLoad` where the error `Received an invalid column length from the bcp client for colid x` is thrown when scales are nor resolved for certain columns.

### DIFF
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -7,7 +7,7 @@ import { TYPE as TOKEN_TYPE } from './token/token';
 import Message from './message';
 import { TYPE as PACKET_TYPE } from './packet';
 
-import { Parameter } from './data-type';
+import { DataType, Parameter } from './data-type';
 import { RequestError } from './errors';
 
 const FLAGS = {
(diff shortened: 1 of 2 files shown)


src/bulk-load.ts
```diff
@@ -131,7 +131,7 @@ class BulkLoad extends EventEmitter {
     this.bulkOptions = { checkConstraints, fireTriggers, keepNulls, lockTable };
   }
 
-  addColumn(name: string, type: any, { output = false, length, precision, scale, objName = name, nullable = true }: ColumnOptions) {
```
arthurschreiber: 🤦‍♂for the use of `any` here...

PR #1004, opened 2019-11-13
fix error message for (typeof config.options.isolationLevel !== 'number')

OriHoch:


src/connection.ts
```diff
@@ -780,7 +780,7 @@ class Connection extends EventEmitter {
 
       if (config.options.isolationLevel !== undefined) {
         if (typeof config.options.isolationLevel !== 'number') {
-          throw new TypeError('The "config.options.language" property must be of type numer.');
+          throw new TypeError('The "config.options.isolationLevel" property must be of type numer.');
```
arthurschreiber: Can you also fix the `numer` -> `number` typo? 🙇

OriHoch: fixed

Issue #510, opened 2017-02-21
Validate isolationLevel value passed by the user.


tvrprasad:
There are four places where this validation needs to be done:

Connection.config.options.connectionIsolationLevel
Connection.config.options.isolationLevel
Connection.transaction()
Connection.beginTransaction()

@arthurschreiber Why do we have two different config options on Connection - isolationLevel and connectionIsolationLevel? As I read the code, it seems like isolationLevel always overrides connectionIsolationLevel.

PR #1016, opened 2019-12-17
510 validate isolation level

IanChokS:
Fix for #510


arthurschreiber:
@IanChokS Thanks for preparing this! There is a bunch of unrelated changes in this PR (all the changes in `package-lock.json`) - can you get that cleaned up? 🙇

IanChokS:
@arthurschreiber done :)

IanChokS:
Also Codecov seems to be irritatingly inconsistent. I'll try to look for alternative code coverage tools

IanChokS:
Node.js 6 environment will fail the test because the PR includes the `Object.values()` function which I don't believe was part of JavaScript at the time Node.js 6 came out

src/connection.ts
```diff
@@ -1865,6 +1865,11 @@ class Connection extends EventEmitter {
   }
 
   beginTransaction(callback: (err: Error | null | undefined, transactionDescriptor?: Buffer) => void, name = '', isolationLevel = this.config.options.isolationLevel) {
+    const isoValues = Object.values(ISOLATION_LEVEL);
```
arthurschreiber: How do you feel about extracting this check to the `Transaction` class, as a static method? There's multiple methods that take isolation level strings, and centralising the validation seems useful to me. Something along the lines of `Transaction.validateIsolationLevel`.

Additionally, how do you feel about changing the way the check is performed? `Object.values` always allocates a new `Array` object that holds the same numbers (`0`, `1`, `2`, `3`, `4`, `5`), and calling `.includes` will always iterate over these values to check if the given `isolationLevel` can be found. This can be simplified by doing a check that verifies `isolationLevel >= 0 && isolationLevel <= 5`, or even a simple `switch` statement like:

```js
switch (isolationLevel) {
  case ISOLATION_LEVEL.NO_CHANGE:
  case ISOLATION_LEVEL.READ_UNCOMMITTED:
  case ISOLATION_LEVEL.READ_COMMITTED:
  case ISOLATION_LEVEL.REPEATABLE_READ:
  case ISOLATION_LEVEL.SERIALIZABLE:
  case ISOLATION_LEVEL.SNAPSHOT:
    break;

  default:
    throw new Error(`Error: invalid isolationLevel ${isolationLevel}`);
}
```

test/integration/connection-test.js
```diff
@@ -359,6 +359,17 @@ describe('Encrypt Test', function() {
   });
 });
 
+describe('BeginTransaction Tests', function() {
+  it('should throw invalid IsolationLevel Error', function() {
+    const config = getConfig();
+    config.options.isolationLevel = 9;
+    const connection = new Connection(config);
```
arthurschreiber: You probably should wait for the connection to be established before you call `connection.beginTransaction`. 🤔

IanChokS:
Moved to #1030

PR #1019, opened 2019-12-19
fix #1018

### LINKED ISSUE #1018: DateTime parameter,the Day is wrong when time >'23:59:16'
[https://github.com/tediousjs/node-mssql/issues/962#issue-540075836](url)

mrliuc:
**Before submitting a PR :**

### DIFF
--- a/src/data-types/bigint.ts
+++ b/src/data-types/bigint.ts
@@ -1,4 +1,4 @@
-import { DataType, ParameterData } from '../data-type';
+import { DataType } from '../data-type';
 import IntN from './intn';
 
 const BigInt: DataType = {
@@ -15,13 +15,10 @@ const BigInt: DataType = {
     buffer.writeUInt8(8);
   },
 
-  writeParameterData: function(buffer, parameter: ParameterData<null | unknown>, _options, cb) {
-    const value = parameter.value;
-
-    if (value != null) {
-      const val = typeof value !== 'number' ? parseInt(value as string) : value;
+  writeParameterData: function(buffer, parameter, _options, cb) {
+    if (parameter.value != null) {
       buffer.writeUInt8(8);
-      buffer.writeInt64LE(val);
+      buffer.writeInt64LE(Number(parameter.value));
     } else {
       buffer.writeUInt8(0);
     }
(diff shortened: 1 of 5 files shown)


arthurschreiber:
Oh, this is great, thanks! ❤️ I'll review, merge and release this over the next few days.

dhensby:
@arthurschreiber - I'm not familiar with UTC dates in JS so I don't know if this is the right fix, but this seems to be caused by a problem with timezones (possibly?)

The node-mssql lib's test suite has a test for this already that fails locally on my mac unless I provide the `TZ=UTC` env var. Is there a more elegant way of correcting this than checking if ms is >= `86357000` which seems like hard coding a value that just happens to be there without truly understanding what's going on and why?

Won't think have implications for any date where the time is `>= 23:59:17` ? I'd like to see a test to make sure it doesn't start removing days from dates just because the time is late

see https://github.com/tediousjs/node-mssql/issues/962#issuecomment-568189352

mrliuc:
I did some tests and found this value 86357000 ，but this solution is really temporary 。

I am looking forward to the official solution...

dhensby:
> I did some tests and found this value 86357000 ，but this solution is really temporary 。

did you look into whether there are any timezone discrepancies?

arthurschreiber:
@dhensby I'm not sure if that's relevant here, but timezone support is basically non-existant in `Date` objects. The support is basically limited to parsing strings with a timezone identifier, but the resulting `Date` object won't carry the timezone information with it, and will have the timezone of the node process instead, I think. And then any calculations that are made on that object will be made in the process' timezone, so things like DST changes will probably end up incorrect.

I'm not sure if that's the issue here. I'll read up on the discussion over in https://github.com/tediousjs/node-mssql/issues/962#issuecomment-568189352 and try to understand what this code is supposed to fix and get back to this PR soon. 🙇

arthurschreiber:
@mrliuc Which timezone is your SQL Server and your Node.js app running in? I can't reproduce this at all. 🤔

mrliuc:
> @mrliuc Which timezone is your SQL Server and your Node.js app running in? I can't reproduce this at all. 🤔

The timezone is (UTC+08:00),and the configuration set `"useUTC": false

mrliuc:
I think I fixed the issue this time,

The time part  is always right, just the day is wrong,so I thought was there any problem with the day algorithm ?

[fix #1018](https://github.com/tediousjs/tedious/pull/1019/commits/ac48d86c996abff6fd57ec100fff1b8903fdfce3)

Issue #883, opened 2019-03-29
Incorrect time from Types.DateTime


bitBrainResearch:
When calling a usp which does no conversion and returns sql server's GetDate() function value as-is, the time returned using Types.DateTime is incorrect.

It's 10am, the (incorrect but doesn't matter) time zone is reported as GMT-0400.  The time returned is 6am. 

It appears as if someone was expecting GMT, modified it by the GMT offset and returned that.  When instead, they are getting the correct time already modified and then are adding the offset a second time.

I'm getting: 
Fri Mar 29 2019 06:37:14 GMT-0400 (Eastern Daylight Time)
From tedious

On sql server: 
2019-03-29 10:37:38.943

On further experimentation, even pulling down the straight date without using a usp and thus not going through Types.DateTime still causes some parsing.  So this might be a javascript thing although it seems weird that JS is just going to parse random values it might recognize.  It makes more sense that tediousjs is parsing values like that?

I thought it might just be a ToString() style conversion but when I debugged I found that my returned date was the long string plus a whole bunch of time functions.

MichaelSun90:
Hi @bitBrainResearch 
I am not familiar with usp, but as you explained, it passes through the sql.getdate()'s value. From your example, I assume it is '2019-03-29 10:37:38.943'.  I run some test on my side, create a data using JS date constructor  'new Date('2019-03-29 10:37:38.943')'. Then, call the writeParameterData under datetime.js from Tedious side. It seems to return the correct time 'Fri Mar 29 2019 10:37:38'  on my side. 

I am not sure this is the reproduce step that you expect, or you can provide me come repo steps, so I can try it.  Also, I remembered there were some discussion happened last year, and there was a fix mentioned. You can also try the lastest Tedious version is you are not currently using the latest one. See if that helps. 

For the last part of your issue: I am not sure I understood everything.   "So this might be a javascript thing although it seems weird that JS is just going to parse random values it might recognize." What do you mean about the random values?  For this part, "I thought it might just be a ToString() style conversion but when I debugged I found that my returned date was the long string plus a whole bunch of time functions.", which part you expect is a ToString()  conversion? Some clarification will be extremely 
appreciated.

IanChokS:
@bitBrainResearch Do you have a reproducible case for us to try out?

IanChokS:
@bitBrainResearch This issue should be fixed by #1023

PR #1023, opened 2019-12-26
Fix date calculations for `date`, `datetime`, `datetime2` and `datetimeoffset`

arthurschreiber:
This changes the date calculations to not be based on Unix Epoch milliseconds, but on actual days between the data types base time and the given value.

This should fix all weirdness around DST changes and other timezone related shenanigans.

Fixes: https://github.com/tediousjs/tedious/issues/1018

References: https://github.com/tediousjs/tedious/issues/883, https://github.com/tediousjs/tedious/issues/755


arthurschreiber:
@mrliuc can you check this out? I believe it should fix the weird issues you've been running into around `datetime` values. 🙇

arthurschreiber:
For some more background information here - I believe the way we calculated the number of days between the epoch dates for `datetime` (`01.01.1900`) and `datetime2`/`date`/`datetimeoffset` (`01.01.0001`) and the passed in `Date` object was having problems with changing UTC offsets due to timezones changing their offsets over the years. I think this could lead to the days calculation being off by one in some cases.

See the following example based on the `Asia/Hong_Kong` timezone:

```
$ TZ=Asia/Hong_Kong node
Welcome to Node.js v13.5.0.
Type ".help" for more information.
> new Date(1900, 00, 01)
1899-12-31T16:23:18.000Z
> new Date(1900, 00, 01).toString()
'Mon Jan 01 1900 00:00:00 GMT+0736 (Hong Kong Standard Time)'
> new Date().toString()
'Fri Dec 27 2019 06:43:12 GMT+0800 (Hong Kong Standard Time)'
```

This behavior only changed semi-recently, with Node 10. Earlier versions had different behaviour, as can be seen here when running the same commands as before on Node.js 8:

```
Arthurs-MacBook-Pro:tedious arthurschreiber$ TZ=Asia/Hong_Kong node
> new Date(1900, 00, 01).toString()
'Mon Jan 01 1900 00:00:00 GMT+0800 (HKT)'
> new Date().toString()
'Fri Dec 27 2019 06:45:20 GMT+0800 (HKT)'
```

This update in Node 10 was based on a change in V8 to better reflect timezone handling and be more compliant with the EcmaScript standard, but it caused the days calculation as implemented in `tedious` to be wrong in some cases.

This PR switches all these day calculations over to be based on `LocalDate` objects from the `@js-joda/core` module, which completely side-steps any timezone shenanigans and just calculates days based on the calendar dates.

This also removes the special code that was added in `tedious` for DST changes - I think that code was never correct to begin with, as DST changes are not even handled by any of the date types available in SQL Server.

mrliuc:
@arthurschreiber  I did some test, when I insert into DB and Millisecond>997,  The time part is wrong ,It's hard to describe, Here's the test code and results, when you see it,you will konw

code:
[code](https://github.com/mrliuc/node-mssql/blob/master/test/tedious/datetimedb.js)

results:
[mocha result](https://github.com/mrliuc/node-mssql/blob/master/test/tedious/datetime-db-test-result.png)     
[db result](https://github.com/mrliuc/node-mssql/blob/master/test/tedious/datetime-db-test-result2.png)

arthurschreiber:
@mrliuc This is expected behaviour in SQL Server - the `datetime` data type's precision is not high enough to store milliseconds with the resolution you require.

This behaviour is documented here: https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver15#rounding-of-datetime-fractional-second-precision

As mentioned at the top of the linked page, you should probably move to `datetime2` or `datetimeoffset` if you require higher precision than what is offered by `datetime`.

arthurschreiber:
@dhensby Can you run this against the `node-mssql` test suite? I believe this will fix all weirdness around `date`, `datetime`, `datetime2` and `datetimeoffset` data types.

mrliuc:
@arthurschreiber datetime2 & datetimeoffset has a different but similar issue ,

I inserted into DB '2019-12-19 23:59:59:998',the data in DB was '2019-12-19 23:59:59.9980000',when I selected,I got  '2019-12-19 23:59:59:997'

I saw the test suite in node-mssql and I think it's not comprehensive enough, the test suite just select the variable like this:
`select @dt1 as dt1, @dt2 as dt2`,  [code](https://github.com/tediousjs/node-mssql/blob/6de9375f1f4a2d308befacd2ccf3aa5d41e453d6/test/common/times.js#L125)

It's not enough,you must insert into the really db like this:
`insert into [TestCreateTime] values(@dt1);select CreateTime dt1 from TestCreateTime` [code](https://github.com/mrliuc/node-mssql/blob/e69f5031801d64867c2dc881a6b0178a3c4ccf4d/test/tedious/datetimedb.js#L49)

Then this line of code will get the result unequal
`assert.strictEqual(+result.recordset[0].dt1, date.getTime())`

PS:the test code about [datetime2](https://github.com/mrliuc/node-mssql/blob/master/test/tedious/datetime2db.js),[datetimeoffset](https://github.com/mrliuc/node-mssql/blob/master/test/tedious/datetimeOffsetdb.js)

mrliuc:
It seems like :

> datetime: insert wrong select right

> datetim2: insert right select wrong

I would rather use the datetime,unless you're going to be compatible with this issue.

arthurschreiber:
@mrliuc If you need millisecond accuracy, then `datetime2` is the correct datatype to use, not `datetime`.

You can try the following to see this is an issue on SQL Server that can't be fixed in `tedious`:

```sql
SELECT CONVERT(varchar, CAST('2019-11-19 23:59:59:998' as datetime), 127);
-- 2019-11-19T23:59:59.997

SELECT CONVERT(varchar, CAST('2019-11-19 23:59:59:998' as datetime2), 127);
-- 2019-11-19T23:59:59.9980000
```

I also added additional test cases in https://github.com/tediousjs/tedious/pull/1023/commits/37c63ed8f372bcb09b71213076f03c836cf7871f that should show that we're handling `datetime2` values correctly.

Did you update the `mssql` version you used for testing to use a version of `tedious` with the fixes in this PR?  🤔 I really can't explain how or why the `datetime2` values are wrong, as the test cases I just added all pass.

mrliuc:
@arthurschreiber  You are right,I find a bug in my test case , It's all pass now.Thank you very much

dhensby:
> @dhensby Can you run this against the `node-mssql` test suite? I believe this will fix all weirdness around `date`, `datetime`, `datetime2` and `datetimeoffset` data types.

I have tested it locally and the previously failing test is now passing 🎉

Issue #983, opened 2019-10-04
Documentation inaccurate for `trustServerCertificate` config value


FSQuaratiello:
In the [documentation for `Connection` objects](http://tediousjs.github.io/tedious/api-connection.html#function_newConnection), it states that the default value of the `trustServerCertificate` option is `false`. 

However, [in the actual code](https://github.com/tediousjs/tedious/blob/master/src/connection.js#L245), it appears that this value defaults to `true`. Is there a reason for this inconsistency? The security implications (by default, connections will skip certificate verification, while users of Tedious will be entirely justified in believing that certificate verification is taking place) seem rather significant.

IanChokS:
Hi @FSQuaratiello, thanks for raising this issue! For now I've updated the docs to show that `trustServerCertificate = true` to better reflect the actual default value in Tedious. We will take a deeper look into this issue with possible changes coming soon.

PR #1029, opened 2020-01-04
Deprecate default value for `trustServerCertificate ` being `true`

arthurschreiber:
Defaulting `trustServerCertificate` to `true` is a potential security issue, so we should switch to it being set to false, and maybe even update our documentation to guide users to use the correct `cryptoCredentialsDetails` values to verify server certificates instead of relying on `trustServerCertificate`.

References: https://github.com/tediousjs/tedious/issues/983

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -915,6 +915,8 @@ class Connection extends EventEmitter {
         }
 
         this.config.options.trustServerCertificate = config.options.trustServerCertificate;
+      } else {
+        deprecate('The default value for `config.options.trustServerCertificate` will change from `true` to `false` in the next major version of `tedious`. Set the value to `true` or `false` explicitly to silence this message.');
       }
 
       if (config.options.useColumnNames !== undefined) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,8 @@ before_install:
         },
         "options": {
           "port": 1433,
-          "database": "master"
+          "database": "master",
+          "trustServerCertificate": true
         }
       }
     }' > ~/.tedious/test-connection.json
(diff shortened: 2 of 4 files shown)


src/connection.ts
```diff
@@ -912,6 +912,8 @@ class Connection extends EventEmitter {
         }
 
         this.config.options.trustServerCertificate = config.options.trustServerCertificate;
+      } else {
+        deprecate('The default value for `config.options.trustServerCertificate` will change from `true` to `false` in the next major version of `tedious`. Set the value to `true` or `false` explicitly to silence this message.');
```
MichaelSun90: Hi @arthurschreiber. I just had a question. Is this change only a message for users to get aware of the change ahead? The actual default value change for line 537 will be in another PR?

arthurschreiber: Yes exactly. If we change the default, a lot of applications that rely on `trustServerCertificate` defaulting to `true` will break, so I'd like to do that change together with a major version bump.

By emitting a value when `trustServerCertificate` is not set (and not emitting the message if a value is provided), we can nudge users to set a value explicitly. If we then change the default, they will be unaffected (because they provided an explicit value).

IanChokS:
@arthurschreiber Will the breaking change be merged along side the dropping Node.js 6/8 support?

arthurschreiber:
Yeah, I'd just group all pending breaking changes (that are ready) together, so we only bump the major version once.

arthurschreiber:
> Yeah, I'd just group all pending breaking changes (that are ready) together, so we only bump the major version once.

Just a note that this is not a hard requirement. I'm also fine bumping the major version multiple times, if that means we can move a bit quicker and don't get hold up. 🤷‍♀

PR #1031, opened 2020-01-06
refactor: return specific data type to users

IanChokS:


IanChokS:
Yup sounds good for both. I'll go over the changes and refactor them.

arthurschreiber:
@IanChokS I pushed some additional changes to your PR. I moved the mapping into the parser, where we now map the internal types to external types whenever a new column metadata list is assigned to the parser. This way, the mapping only happens once (after we parsed the column metadata).

When we parse `RowToken`s or `NBCRowToken`s, we use the internal column metadata when parsing the data, put use the external column metadata when building the tokens. This should fix all cases where we previously returned types like `IntN` in a consistent fashion.

What do you think?

IanChokS:
Sounds good 👍 

Thanks!

IanChokS:
@arthurschreiber Any plans on dropping node.js 6 or 8 support soon?

PR #1033, opened 2020-01-06
refactor: cancelling before buffer

IanChokS:
A fix for issue #948

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2005,18 +2005,15 @@ class Connection extends EventEmitter {
       let message: Message;
 
       request.once('cancel', () => {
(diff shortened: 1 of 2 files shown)


arthurschreiber:
@IanChokS Can you use `git commit --amend` (with a force-push) to change the commit message to `fix: properly cancel requests in `BUILDING_CLIENT_REQUEST` state`? 🙇 `refactor` commits won't trigger a new release, and the fix won't be listed in the release changelog otherwise.

arthurschreiber:
@IanChokS Actually, I just squash merged this with a fixed up commit message, so all is good! 🙇

PR #1037, opened 2020-01-17
refactor: generate request payloads using generator functions

IanChokS:
When parameters or bulk load rows are converted from JavaScript values to actual TDS request payload data, a method called `writeParameterData` is currently used.

This method takes a `WritableTrackingBuffer` argument and stores the converted data type information in this tracking buffer. As part of https://github.com/tediousjs/tedious/issues/1038, we want to move to a model where this parameter data is not fully stored in one `WritableTrackingBuffer` before a request is written to the network, but instead we want to generate data in a streaming fashion and start sending data over the network as soon as possible.

This changeset adds a new `generate` generator method to all data types, which generates individual `Buffer`s containing the same data as has been written by `WritableTrackingBuffer` so far. `writeParameterData` methods were update to make use of the generator objects returned by `generate`, until all of the request payload generation functionality gets updated to make use of the generator objects directly.


src/data-type.ts
```diff
@@ -69,14 +69,16 @@ export interface DataType {
 
   declaration(parameter: Parameter) : string,
   writeTypeInfo(buf: any, data: ParameterData, options: InternalConnectionOptions) : void,
-  writeParameterData(buf: any, data: ParameterData, options: InternalConnectionOptions, callback: () => void) : void,
+  writeParameterData(buf: any, data: ParameterData, options: InternalConnectionOptions, callback: () => void): void,
   validate(value: any) : any, // TODO: Refactor 'any' and replace with more specific type.
+  generate(parameter: ParameterData, _options: InternalConnectionOptions): any,
```
arthurschreiber: The proper type for this function would be:

```suggestion
  generate(parameter: ParameterData, _options: InternalConnectionOptions): Generator<Buffer, void>,
```

src/data-types/datetime.ts
```diff
@@ -55,10 +66,13 @@ const DateTime: DataType = {
       buffer.writeUInt8(8);
       buffer.writeInt32LE(days);
       buffer.writeUInt32LE(threeHundredthsOfSecond);
+
+      yield buffer.buffer
```
arthurschreiber: I think this is a typo?  🤔 `WritableTrackingBuffer.buffer` contains the whole pre-allocated buffer, which can be bigger than the data that was written in the `WritableTrackingBuffer` (and would be returned via `WritableTrackingBuffer.data`). I think that's what's causing the remaining test failures. 😬 

```suggestion
      yield buffer.data;
```

IanChokS: Oh oops yeah that was a typo

test/unit/data-type.js
```diff
@@ -75,9 +75,322 @@ describe('Data Types', function() {
...
+    const type = TYPES.typeByName.BigInt;
+
+    for (const [value, expected] of [
+      [123456789, Buffer.from('0815cd5b0700000000', 'hex')],
+      [null, Buffer.from('00', 'hex')],
+    ]) {
+      const buffer = new WritableTrackingBuffer(0);
+      const parameterValue = { value };
+      type.writeParameterData(buffer, parameterValue, { useUTC: false }, () => { });
+      assert.isTrue(buffer.data.equals(expected));
```
arthurschreiber: I think it would be better to use `assert.deepEqual(buffer.data, expected);` here. 

In case of a failure `assert.isTrue` will only tell you that something was `false` instead of `true`, while `assert.deepEqual` will show the individual bytes, making it a bit easier to understand which bytes were different (and what values they had).

You should always try to use the most specific assertion functions instead of more generic ones.

test/unit/data-type.js
```diff
@@ -75,9 +75,322 @@ describe('Data Types', function() {
...
+    for (const [value, expected] of [
+      [Buffer.from([0x12, 0x34, 0x00, 0x00]), Buffer.from('040012340000', 'hex')],
+      [null, Buffer.from('ffff', 'hex')]
+    ]) {
+      const buffer = new WritableTrackingBuffer(0);
+      const parameterValue = { value, length: 4 };
+      type.writeParameterData(buffer, parameterValue, { useUTC: false }, () => { });
+      assert.isTrue(buffer.data.equals(expected));
+    }
+  });
```
arthurschreiber: All these test cases that call `writeParameterData` are written in a synchronous fashion, while the `writeParameterData` is executing in a potentially asynchronous fashion - e.g. for `TVP`, the passed in callback is sometimes called after one or multiple `setImmediate` calls.

The reason why these test cases are written in a synchronous fashion is that initially, `writeParameterData` was a regular synchronous function and did not take a callback parameter, and when that was changed, these test cases were not updated correctly.

A better way to write these test cases would be something like this:

```js
  it('should writeParameterData binary (Buffer)', function(done) {
    const value = Buffer.from([0x12, 0x34, 0x00, 0x00]);
    const expected = Buffer.from('040012340000', 'hex');

    const type = TYPES.typeByName.Binary;

    const buffer = new WritableTrackingBuffer(0);
    const parameterValue = { value, length: 4 };
    type.writeParameterData(buffer, parameterValue, { useUTC: false }, () => {
      assert.isTrue(buffer.data.equals(expected));

      done();
    });
  });

  it('should writeParameterData binary (null)', function(done) {
    const value = null;
    const expected = Buffer.from('ffff', 'hex');

    const type = TYPES.typeByName.Binary;

    const buffer = new WritableTrackingBuffer(0);
    const parameterValue = { value, length: 4 };

    type.writeParameterData(buffer, parameterValue, { useUTC: false }, () => {
      assert.isTrue(buffer.data.equals(expected));

      done();
    });
  });
```

Notice the addition of `done` to the test function, and calling `done` inside the callback passed to `writeParameterData`.

Also, the assertion was moved into the callback to `writeParameterData` as well.

I also had to break the test into two separate test cases, because testing multiple input/output combinations inside an asynchronous test function is awkward, and it's easier to just create multiple test cases.

test/unit/data-type.js
```diff
@@ -75,9 +75,866 @@ describe('Data Types', function() {
...
+  it('should writeParameterData DateTimeOffSet (Buffer)', function(done) {
+    const value = new Date(Date.UTC(2014, 1, 14, 17, 59, 59, 999));
+    const expected = Buffer.from('0820fd002d380b20fe', 'hex');
+
+    const type = TYPES.typeByName.DateTimeOffset;
+
+    const buffer = new WritableTrackingBuffer(0);
+    const parameterValue = { value, scale: 0 };
+
+    type.writeParameterData(buffer, parameterValue, { useUTC: false }, () => {
```
arthurschreiber: I think this test is failing because `useUTC: false` instead of `useUTC: true` here.

The `expected` value is based on what the value would be in your timezone, while the system running the tests might be in a different timezone and thus sending a different value and offset.

Try changing `useUTC` to `true` and update the `expected` value to be UTC based (offset = 0), and then the tests should also pass on CI and other systems.

src/data-types/bigint.ts
```diff
@@ -15,15 +17,22 @@ const BigInt: DataType = {
     buffer.writeUInt8(8);
   },
 
-  writeParameterData: function(buffer, parameter, _options, cb) {
+  writeParameterData: function(buff, parameter, options, cb) {
+    buff.writeBuffer(Buffer.concat(Array.from(this.generate(parameter, options))));
+    cb();
+  },
+
+  generate: function* (parameter: ParameterData, options: InternalConnectionOptions) {
```
arthurschreiber: As the parameter types for `generate` are defined in the `DataType` definition, specifying the types again here is a bit repetitive.

This could be changed to the following:

```suggestion
  generate: function* (parameter, options) {
```

test/unit/data-type.js
IanChokS: Slicing the last two buffers off, which holds different values based on local machine timezone

PR #1044, opened 2020-01-29
Use generators to generate RPC payload data

IanChokS:


src/connection.ts
```diff
@@ -5,7 +5,7 @@ import { Socket } from 'net';
 import constants from 'constants';
 import { createSecureContext, SecureContext, SecureContextOptions } from 'tls';
 
-import { Readable } from 'readable-stream';
+import { Readable } from 'stream';
```
arthurschreiber:
```suggestion
import { Readable } from 'readable-stream';
```

src/rpcrequest-payload.ts
```diff
@@ -3,7 +3,8 @@ import { writeToTrackingBuffer } from './all-headers';
 import Request from './request';
 import { Parameter, ParameterData } from './data-type';
 import { InternalConnectionOptions } from './connection';
-import { Readable } from 'readable-stream';
+import { Readable } from 'stream';
```
arthurschreiber:
```suggestion
import { Readable } from 'readable-stream';
```

src/rpcrequest-payload.ts
```diff
@@ -34,19 +35,10 @@ class RpcRequestPayload {
...
-
-    return new Readable({
-      read() {
-        self.getData((buf) => {
-          this.push(buf);
-          this.push(null);
-        });
-      }
-    });
+    return Readable.from(this.generateData());
```
arthurschreiber:
```suggestion
    return Readable.from(this.generateData(), { objectMode: false });
```

According to https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_stream_readable_from_iterable_options, a `Readable` stream created by `Readable.from` has by default `objectMode: true`. But as we're working with `Buffer` data exclusively, changing it to `objectMode: false` is preferred.

src/rpcrequest-payload.ts
```diff
@@ -64,26 +56,32 @@ class RpcRequestPayload {
...
+    buffer.writeBuffer(Buffer.concat(buffers));
 
-      this._writeParameterData(parameters[i], buffer, () => {
-        setImmediate(() => {
-          writeNext(i + 1);
-        });
-      });
-    };
-    writeNext(0);
+    yield buffer.data;
```
arthurschreiber: I think by using `parameters.forEach`, you ran into a sort-of "limitation" of generator functions - that is, you can't `yield` from inside a closue.

I see you worked around this by consuming the generator and collecting all buffers in the `buffers` array first. This works, but there's a disadvantage here - we're back to having to collect all the request data we need to send off into a large buffer first, and we'll be back at having to pay with higher latency as well as higher memory usage.

A better solution would be to use a plain old `for` loop here, like this:

```js
// Yield data written to `buffer` so far
yield buffer.data;

// Yield parameter data
for(let i = 0; i < parameters.length; i++) {
  yield * this.generateParameterData(param, this.options);
}
```

arthurschreiber:
### Before

```
$ node benchmarks/query/insert-varbinary.js 
query/insert-varbinary.js size=10 n=10: 308.6890408599335
query/insert-varbinary.js size=1024 n=10: 311.47775919536076
query/insert-varbinary.js size=1048576 n=10: 18.102475918534246
query/insert-varbinary.js size=10485760 n=10: 2.109837591831322
query/insert-varbinary.js size=10 n=100: 497.5944393940821
query/insert-varbinary.js size=1024 n=100: 401.36903928896766
query/insert-varbinary.js size=1048576 n=100: 18.25438779354878
query/insert-varbinary.js size=10485760 n=100: 2.250578908715709

$ node benchmarks/query/select-varbinary.js 
query/select-varbinary.js size=10 n=10: 347.8905223315275
query/select-varbinary.js size=100 n=10: 343.6256665693634
query/select-varbinary.js size=1000 n=10: 355.023165971626
query/select-varbinary.js size=10000 n=10: 328.7983708435283
query/select-varbinary.js size=1048576 n=10: 28.922057533567365
query/select-varbinary.js size=10485760 n=10: 3.61266827520264
query/select-varbinary.js size=10 n=100: 524.0513003670094
query/select-varbinary.js size=100 n=100: 545.4001204723419
query/select-varbinary.js size=1000 n=100: 482.9299872299306
query/select-varbinary.js size=10000 n=100: 405.018117776717
query/select-varbinary.js size=1048576 n=100: 32.81065020423878
query/select-varbinary.js size=10485760 n=100: 3.524019660813264

$ node benchmarks/request/rpcrequest-payload.js 
request/rpcrequest-payload.js size=1048576 n=10: 321.78192015509376
request/rpcrequest-payload.js size=10485760 n=10: 49.40733090363232
request/rpcrequest-payload.js size=52428800 n=10: 8.52002443536192
request/rpcrequest-payload.js size=1048576 n=100: 597.4325503430123
request/rpcrequest-payload.js size=10485760 n=100: 62.95259916749355
request/rpcrequest-payload.js size=52428800 n=100: 11.697934463285144
```

### After

```
$ node benchmarks/query/insert-varbinary.js 
query/insert-varbinary.js size=10 n=10: 279.0921311592354
query/insert-varbinary.js size=1024 n=10: 312.6991209371042
query/insert-varbinary.js size=1048576 n=10: 20.249761407692457
query/insert-varbinary.js size=10485760 n=10: 2.1134659008144805
query/insert-varbinary.js size=10 n=100: 445.1387124563054
query/insert-varbinary.js size=1024 n=100: 435.0690013780289
query/insert-varbinary.js size=1048576 n=100: 22.43368865230221
query/insert-varbinary.js size=10485760 n=100: 2.366485339029834

$ node benchmarks/query/select-varbinary.js 
query/select-varbinary.js size=10 n=10: 338.18833119716675
query/select-varbinary.js size=100 n=10: 325.718979045586
query/select-varbinary.js size=1000 n=10: 333.5310950372841
query/select-varbinary.js size=10000 n=10: 320.3208743888598
query/select-varbinary.js size=1048576 n=10: 27.974499900564638
query/select-varbinary.js size=10485760 n=10: 3.343897240855387
query/select-varbinary.js size=10 n=100: 486.66409941620947
query/select-varbinary.js size=100 n=100: 541.0536085167336
query/select-varbinary.js size=1000 n=100: 501.0910380444101
query/select-varbinary.js size=10000 n=100: 415.54201777837517
query/select-varbinary.js size=1048576 n=100: 29.55069400842108
query/select-varbinary.js size=10485760 n=100: 3.5741086078328834

$ node benchmarks/request/rpcrequest-payload.js 
request/rpcrequest-payload.js size=1048576 n=10: 445.0352534676034
request/rpcrequest-payload.js size=10485760 n=10: 80.31625456905117
request/rpcrequest-payload.js size=52428800 n=10: 16.873847022688324
request/rpcrequest-payload.js size=1048576 n=100: 876.175028473717
request/rpcrequest-payload.js size=10485760 n=100: 106.22369527530356
request/rpcrequest-payload.js size=52428800 n=100: 21.44433932107644
```

### Analysis

Not much of a change here, performance wise. We're a bit slower for small payloads, because we're allocating more smaller buffers now in `generateData` and `generateParameterData` (at least one buffer per parameter, plus some other buffers that hold other request data). Before, a single `WritableTrackingBuffer` was used with an initial size that could hold parameter data for requests with small payloads without any additional allocations.

For larger parameter data, we're a bit faster because less parameters no longer share the same buffer which needed to be re-sized often.

I expect that this will improve (both for small requests as well as larger requests) with the other changes we're planning to do next. 👍

PR #1049, opened 2020-02-10
Improve Performance - Refactor plp

IanChokS:
**My Machine**
_Comparing #1044 (before) with this PR (after)_

**Before:** 
$ node benchmarks/query/insert-varbinary.js
query\insert-varbinary.js size=10 n=10: 719.5642319011607
query\insert-varbinary.js size=1024 n=10: 705.0743708921045
query\insert-varbinary.js size=1048576 n=10: 58.216095819036425
query\insert-varbinary.js size=10485760 n=10: 10.632852494477827
query\insert-varbinary.js size=10 n=100: 1,131.653130209141
query\insert-varbinary.js size=1024 n=100: 1,121.2849153686102
query\insert-varbinary.js size=1048576 n=100: 102.65458589586083
query\insert-varbinary.js size=10485760 n=100: 13.391209246951373

$ node benchmarks/query/select-varbinary.js
query\select-varbinary.js size=10 n=10: 960.9286414390867
query\select-varbinary.js size=100 n=10: 943.1914826790902
query\select-varbinary.js size=1000 n=10: 901.0876127485875
query\select-varbinary.js size=10000 n=10: 864.8050297060528
query\select-varbinary.js size=1048576 n=10: 97.16577066384849
query\select-varbinary.js size=10485760 n=10: 23.32439309346061
query\select-varbinary.js size=10 n=100: 1,326.780805196735
query\select-varbinary.js size=100 n=100: 1,310.4786039534335
query\select-varbinary.js size=1000 n=100: 1,290.2576349027902
query\select-varbinary.js size=10000 n=100: 1,190.9922871339486
query\select-varbinary.js size=1048576 n=100: 220.8057772047495
query\select-varbinary.js size=10485760 n=100: 30.848338247833514

$ node benchmarks/request/**rpcrequest-payload.js**
request\rpcrequest-payload.js size=1048576 n=10: **630.9426914753333**
request\rpcrequest-payload.js size=10485760 n=10: **72.67632002000052**
request\rpcrequest-payload.js size=52428800 n=10: **16.142013535171973**
request\rpcrequest-payload.js size=1048576 n=100: **697.1779630412019**
request\rpcrequest-payload.js size=10485760 n=100: **81.02967322839558**
request\rpcrequest-payload.js size=52428800 n=100: **16.49164115396037**

**After:**
$ node benchmarks/query/insert-varbinary.js
query\insert-varbinary.js size=10 n=10: 705.8507474359442
query\insert-varbinary.js size=1024 n=10: 699.0954194638931
query\insert-varbinary.js size=1048576 n=10: 59.53206605439921
query\insert-varbinary.js size=10485760 n=10: 10.636543974131925
query\insert-varbinary.js size=10 n=100: 1,180.315640008451
query\insert-varbinary.js size=1024 n=100: 1,097.2540121092952
query\insert-varbinary.js size=1048576 n=100: 106.0414461868382
query\insert-varbinary.js size=10485760 n=100: 14.768045833580125

$ node benchmarks/query/select-varbinary.js
query\select-varbinary.js size=10 n=10: 968.9546916786171
query\select-varbinary.js size=100 n=10: 957.42338219384
query\select-varbinary.js size=1000 n=10: 932.0185658098309
query\select-varbinary.js size=10000 n=10: 863.4386322648539
query\select-varbinary.js size=1048576 n=10: 98.86630013633662
query\select-varbinary.js size=10485760 n=10: 24.82001764206854
query\select-varbinary.js size=10 n=100: 1,348.8013876468676
query\select-varbinary.js size=100 n=100: 1,253.9326462618387
query\select-varbinary.js size=1000 n=100: 1,259.0668551909437
query\select-varbinary.js size=10000 n=100: 1,181.2916479136618
query\select-varbinary.js size=1048576 n=100: 217.26005093879152
query\select-varbinary.js size=10485760 n=100: 32.37861637170988

$ node benchmarks/request/**rpcrequest-payload**
request\rpcrequest-payload.js size=1048576 n=10: **2,083.5069589132427**
request\rpcrequest-payload.js size=10485760 n=10: **2,155.916297843286**
request\rpcrequest-payload.js size=52428800 n=10: **1,988.6250646303145**
request\rpcrequest-payload.js size=1048576 n=100: **6,880.607221844057**
request\rpcrequest-payload.js size=10485760 n=100: **6,671.69223003381**
request\rpcrequest-payload.js size=52428800 n=100: **6,594.827017687326**

**Summary**
_rpcrequest-payload_ **increased** **ops/sec** by **230%** for smaller bytes and up to **40,000%** for larger bytes.

### DIFF
--- a/src/data-types/varbinary.ts
+++ b/src/data-types/varbinary.ts
@@ -3,6 +3,8 @@ import WritableTrackingBuffer from '../tracking-buffer/writable-tracking-buffer'
 
 const NULL = (1 << 16) - 1;
 const MAX = (1 << 16) - 1;
+const UNKNOWN_PLP_LEN = Buffer.from([0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
+const PLP_TERMINATOR = Buffer.from([0x00, 0x00, 0x00, 0x00]);
 
 const VarBinary: { maximumLength: number } & DataType = {
   id: 0xA5,
@@ -41,29 +43,53 @@ const VarBinary: { maximumLength: number } & DataType = {
     }
   },
 
-  writeTypeInfo: function(buffer, parameter) {
-    buffer.writeUInt8(this.id);
+  generateTypeInfo: function(parameter) {
+    const buffer = Buffer.alloc(3);
+    buffer.writeUInt8(this.id, 0);
+
     if (parameter.length! <= this.maximumLength) {
-      buffer.writeUInt16LE(this.maximumLength);
+      buffer.writeUInt16LE(this.maximumLength, 1);
     } else {
-      buffer.writeUInt16LE(MAX);
+      buffer.writeUInt16LE(MAX, 1);
     }
+
+    return buffer;
   },
 
-  writeParameterData: function(buff, parameter, options, cb) {
-    buff.writeBuffer(Buffer.concat(Array.from(this.generate(parameter, options))));
-    cb();
-  },
+  *generateParameterData(parameter, options) {
+    let value = parameter.value;
 
-  generate: function* (parameter, options) {
-    if (parameter.value != null) {
-      const buffer = new WritableTrackingBuffer(0);
+    if (value != null) {
       if (parameter.length! <= this.maximumLength) {
-        buffer.writeUsVarbyte(parameter.value);
-      } else {
-        buffer.writePLPBody(parameter.value);
+        const buffer = new WritableTrackingBuffer(0);
+        buffer.writeUsVarbyte(value);
+        yield buffer.data;
+
+      } else { // writePLPBody
+        let length;
+        if (value instanceof Buffer) {
+          length = value.length;
+        } else {
+          value = value.toString();
+          length = Buffer.byteLength(value, 'ucs2');
+        }
+
+        yield UNKNOWN_PLP_LEN;
+        if (length > 0) {
+          const buffer = Buffer.alloc(4);
+          buffer.writeUInt32LE(length, 0);
+          yield buffer;
+
+          if (value instanceof Buffer) {
+            yield value;
+          } else {
+            yield Buffer.from(value, 'ucs2');
+          }
+        }
+
+        yield PLP_TERMINATOR;
       }
-      yield buffer.data;
+
     } else if (parameter.length! <= this.maximumLength) {
       const buffer = new WritableTrackingBuffer(2);
       buffer.writeUInt16LE(NULL);
--- a/benchmarks/request/rpcrequest-payload-tvp.js
+++ b/benchmarks/request/rpcrequest-payload-tvp.js
@@ -3,6 +3,8 @@ const { createBenchmark } = require('../common');
 const { Request, TYPES } = require('../../lib/tedious');
 const RpcRequestPayload = require('../../lib/rpcrequest-payload');
 
+const { Readable } = require('readable-stream');
+
 const bench = createBenchmark(main, {
   n: [10, 100],
   size: [10, 100, 1000, 10000]
@@ -35,7 +37,7 @@ function main({ n, size }) {
     }
 
     const payload = new RpcRequestPayload(request, Buffer.alloc(0), {});
-    const stream = payload.getStream();
+    const stream = Readable.from(payload, { objectMode: false });
     const chunks = [];
     stream.on('data', (chunk) => {
       chunks.push(chunk);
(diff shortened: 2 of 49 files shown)


src/data-types/varbinary.ts
```diff
@@ -56,14 +56,43 @@ const VarBinary: { maximumLength: number } & DataType = {
...
+            buffer = Buffer.concat([buffer, value], buffer.length + value.length);
+
+          } else {
+            const buffer2 = Buffer.from(value, 'ucs2');
+            buffer = Buffer.concat([buffer, buffer2], buffer.length + buffer2.length);
+          }
+        }
+
+        const end = Buffer.from([0x00, 0x00, 0x00, 0x00]);
+        yield Buffer.concat([UNKNOWN_PLP_LEN, buffer, end], UNKNOWN_PLP_LEN.length + buffer.length + end.length);
```
arthurschreiber: Whenever `Buffer.concat` is called, a new `Buffer` gets allocated with a size that equals the length of all passed in buffers summed together, and then data is copied over from the passed in buffers into the new result buffer.

You can improve both performance as well as memory usage here by not doing that, and instead `yield`ing each chunk that you passed to `Buffer.concat` instead:

```js
        if (length > 0) {
          let buffer = Buffer.alloc(4);
          buffer.writeUInt32LE(length, 0);
          yield buffer;

          if (value instanceof Buffer) {
            yield value;
          } else {
            yield Buffer.from(value, 'ucs2');
          }
        }

        const end = Buffer.from([0x00, 0x00, 0x00, 0x00]);
        yield end;
```

benchmarks/request/rpcrequest-payload.js
```diff
@@ -16,8 +16,9 @@ function main({ n, size }) {
   const buf = Buffer.alloc(size);
   buf.fill('x');
 
-  const request = new Request('INSERT INTO #benchmark ([value]) VALUES (@value)', () => {});
+  const request = new Request('INSERT INTO #benchmark ([value], [val]) VALUES (@value, @val)', () => {});
   request.addParameter('value', TYPES.VarBinary, buf);
+  request.addParameter('val', TYPES.VarChar, 'Hello World');
```
arthurschreiber: I think it would be better to split this up into separate benchmark runs. That way running the benchmarks will give numbers that can easier be understood.

We'd have to change the `createBenchmark` call to this:

```js
const bench = createBenchmark(main, {
  n: [10, 100],
  type: ['varbinary', 'varchar']
  size: [
    1024 * 1024,
    10 * 1024 * 1024,
    50 * 1024 * 1024
  ]
});
```

and `main` would need to be updated to add the `type`:

```js
function main({ n, type, size }) {
```

and then the code here on these lines could be changed like this:

```js
  const request = new Request('INSERT INTO #benchmark ([value]) VALUES (@value)', () => {});
  switch (type) {
    case 'varchar': {
      const value = 'x'.repeat(size);
      request.addParameter('value', TYPES.VarChar, value);
      break;
    }

    case 'varbinary': {
      const value = Buffer.alloc(size);
      value.fill('x');
      request.addParameter('value', TYPES.VarBinary, value);
      break;
    }

    default:
      throw new Error('unknown type ${type}');
  }
```

src/data-types/varbinary.ts
IanChokS: Oh yup, varbinary jumped from **16.49 ops/sec** to **6,594.82 ops/sec** 😅

src/data-types/varbinary.ts
```diff
@@ -56,14 +56,42 @@ const VarBinary: { maximumLength: number } & DataType = {
...
       if (parameter.length! <= this.maximumLength) {
-        buffer.writeUsVarbyte(parameter.value);
-      } else {
-        buffer.writePLPBody(parameter.value);
+        const buffer = new WritableTrackingBuffer(0);
+        buffer.writeUsVarbyte(value);
+        yield buffer.data;
+
+      } else { // writePLPBody
+        const UNKNOWN_PLP_LEN = Buffer.from([0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
```
arthurschreiber: What do you think about making this a top-level constant? This value never changes, and that way we could save ourselves from creating this `Buffer` over and over again for each `VarBinary` parameter.

src/data-types/varbinary.ts
```diff
@@ -56,14 +56,42 @@ const VarBinary: { maximumLength: number } & DataType = {
...
+          yield buffer;
+
+          if (value instanceof Buffer) {
+            yield value;
+          } else {
+            yield Buffer.from(value, 'ucs2');
+          }
+        }
+
+        const end = Buffer.from([0x00, 0x00, 0x00, 0x00]);
```
arthurschreiber: What do you think about making this a top-level constant too? Same reasoning as for https://github.com/tediousjs/tedious/pull/1049/files#r377817446. I think `PLP_TERMINATOR` would be a fitting name for this.

PR #1054, opened 2020-02-23
Split generation of parameter length and data

arthurschreiber:
This splits the generation of parameter length and parameter data into two separate methods.

This supersedes https://github.com/tediousjs/tedious/pull/1053, where a new `toBuffer` method was proposed to perform essentially the same task.

@fredwes Does this cover all the use cases of the `toBuffer` method?

### DIFF
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -177,6 +177,7 @@ class RowTransform extends Transform {
         value: row[i]
       };
 
+      this.push(c.type.generateParameterLength(parameter, this.mainOptions));
       for (const chunk of c.type.generateParameterData(parameter, this.mainOptions)) {
         this.push(chunk);
       }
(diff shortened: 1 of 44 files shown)


fredwes:
@arthurschreiber this covers most of the use cases of `toBuffer` except that the following data types must always return a buffer that is 8 bytes long:
- datetime2
- bit
- int
- smallint
- smallmoney
- tinyint

Although tedious.js will be able to read/write values without these normalizations, values written by tedious.js may not be able to be read by other drivers such as mssql-jdbc.

arthurschreiber:
@fredwes So if I understand this correctly, the buffer that is sent into the cipher has to be 8 bytes, so we can take the output of `generateParameterData` of these types and just pad it with zero bytes at the end so it is 8 bytes long before giving it to the cipher?

fredwes:
@arthurschreiber For only those types I listed, yes, they have to be normalized to 8 bytes before they are encrypted. Padding zeroes the end of the buffer will work for those types, except for `datetime2`. Since, for `datetime2` the first 5 bytes should be the timestamp and the next 3 bytes should be the days, always, regardless of the scale. So you would have to pad the timestamp differently based on the scale used.

PR #1063, opened 2020-03-04
Small improvements to payload generation

arthurschreiber:
This contains a few small improvements to the payload generation code that should slightly improve performance.

### DIFF
--- a/src/rpcrequest-payload.ts
+++ b/src/rpcrequest-payload.ts
@@ -65,7 +65,7 @@ class RpcRequestPayload implements Iterable<Buffer> {
   }
 
   * generateParameterData(parameter: Parameter, options: any) {
-    const buffer = new WritableTrackingBuffer(0);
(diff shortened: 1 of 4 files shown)


src/rpcrequest-payload.ts
```diff
@@ -65,7 +65,7 @@ class RpcRequestPayload implements Iterable<Buffer> {
   }
 
   * generateParameterData(parameter: Parameter, options: any) {
-    const buffer = new WritableTrackingBuffer(0);
+    const buffer = new WritableTrackingBuffer(1 + 2 + Buffer.byteLength(parameter.name, 'ucs-2') + 1);
```
arthurschreiber: This ensures the `Buffer` used by `WritableTrackingBuffer` is big enough to contain the data that will be written to it without causing any resizing,

arthurschreiber:
CI build for Node.js 13.x is currently failing because of an incompatibility with one of the `babel` related packages. https://github.com/tediousjs/tedious/pull/1064 will fix this.

PR #1065, opened 2020-03-05
feat: allow validating bulk load parameters

MichaelSun90:
Validate function will report error is input data do not match the data type check. However, the current bulk logic does not expose this error. Added some logic to expose these errors. 

Related issue: #889

### DIFF
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -151,7 +151,7 @@ class RowTransform extends Transform {
   /**
    * @private
    */
-  _transform(row: Array<any>, _encoding: string, callback: () => void) {
+  _transform(row: Array<any>, _encoding: string, callback: (error?: Error) => void) {
     if (!this.columnMetadataWritten) {
       this.push(this.bulkLoad.getColMetaData());
       this.columnMetadataWritten = true;
@@ -163,6 +163,13 @@ class RowTransform extends Transform {
 
     for (let i = 0; i < this.columns.length; i++) {
       const c = this.columns[i];
+      if (this.bulkLoad.options.validateBulkLoadParameters) {
+        const error = c.type.validate(row[i]);
+
+        if (error instanceof TypeError) {
+          return callback(error);
+        }
+      }
       const parameter = {
         length: c.length,
         scale: c.scale,
@@ -268,6 +275,7 @@ class BulkLoad extends EventEmitter {
    * @private
    */
   rowToPacketTransform: RowTransform;
+  message: Message;
 
   /**
    * @private
@@ -331,6 +339,28 @@ class BulkLoad extends EventEmitter {
     this.streamingMode = false;
 
     this.rowToPacketTransform = new RowTransform(this); // eslint-disable-line no-use-before-define
+    this.message = new Message({ type: PACKET_TYPE.BULK_LOAD });
+    this.rowToPacketTransform.pipe(this.message);
+
+    this.rowToPacketTransform.once('finish', () => {
+      this.removeListener('cancel', onCancel);
+    });
+
+    this.rowToPacketTransform.once('error', (err) => {
+      this.rowToPacketTransform.unpipe(this.message);
+
+      this.error = err;
+
+      this.message.ignore = true;
+      this.message.end();
+    });
+
+    const onCancel = () => {
+      this.rowToPacketTransform.emit('error', RequestError('Canceled.', 'ECANCEL'));
+      this.rowToPacketTransform.destroy();
+    };
+
+    this.once('cancel', onCancel);
 
     this.bulkOptions = { checkConstraints, fireTriggers, keepNulls, lockTable };
   }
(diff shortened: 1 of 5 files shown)


arthurschreiber:
@MichaelSun90 Can you add a test case for this?

arthurschreiber:
Thanks for providing a test case! 🙇 I think we should also add a test case for streaming bulk loads? I have a hunch that the change will not work for streaming bulk loads in its current form. 🤔

src/connection.ts
```diff
@@ -935,6 +938,15 @@ class Connection extends EventEmitter {
         this.config.options.useUTC = config.options.useUTC;
       }
 
+      if (config.options.validateBulkLoadParameters !== undefined) {
+        if (typeof config.options.validateBulkLoadParameters !== 'boolean') {
+          throw new TypeError('The "config.options.validateBulkLoadParameters" property must be of type boolean.');
+        }
+
+        deprecate('This "validateBulkLoadParameters" property will be depercated in the next major version of `tedious`.');
```
MichaelSun90: I am not sure that I suppose to add the message here. I thought that users may want to know that this property is going to be deprecated when they are actually using it which means that when "config.options.validateBulkLoadParameters !== undefined".  However, I am not sure whether I am correct about this.

test/unit/bulk-load-test.js
```diff
@@ -1,13 +1,24 @@
...
   });
 
+  it('add row with a value has a wrong data type', function() {
+    const request = new BulkLoad('tablename', { tdsVersion: '7_2', validateBulkLoadParameters: true }, { });
+    request.addColumn('columnName', TYPES.Date, { nullable: true });
+    try {
+      request.addRow({ columnName: 'Wrong Input' });
+    } catch (err) {
+      assert.strictEqual(err.message, 'Invalid date.');
+    }
```
arthurschreiber, line 19: I think this should use the `assert.throws` helper:
```suggestion
    assert.throws(() => {
      request.addRow({ columnName: 'Wrong Input' });
    }, TypeError, 'Invalid date.');
```

MichaelSun90: Oh, sure. There are few other tests in this test file that also using strictEqual, should I also modify them as well?

MichaelSun90: Nvm, those are under "\test\integration\bulk-load-test.js" 😅

src/connection.ts
```diff
@@ -936,6 +939,16 @@ class Connection extends EventEmitter {
...
       }
 
+      if (config.options.validateBulkLoadParameters !== undefined) {
+        if (typeof config.options.validateBulkLoadParameters !== 'boolean') {
+          throw new TypeError('The "config.options.validateBulkLoadParameters" property must be of type boolean.');
+        }
+
+        this.config.options.validateBulkLoadParameters = config.options.validateBulkLoadParameters;
+      } else {
+        deprecate('The default value for "config.options.validateBulkLoadParameters" will change from `false` to `true` in the next major version of `tedious`. Set the value to `true` or `false` explicitly to silence this message.');
```
MichaelSun90, line 949: Will this be a better message ? "validateBulkLoadParameters" property will be depercated in the next major version of `tedious`."

src/bulk-load.ts
```diff
@@ -92,6 +92,13 @@ class RowTransform extends Transform {
 
     for (let i = 0; i < this.columns.length; i++) {
       const c = this.columns[i];
+      if (this.bulkLoad.options.validateBulkLoadParameters) {
+        const error = c.type.validate(row[i]);
+
+        if (error instanceof TypeError) {
+          return callback(error);
```
MichaelSun90, line 99: Hi @arthurschreiber, since the error is not thrown here, we need additional handling for the error under "addRow" method, right?

src/connection.ts
arthurschreiber: I think we'll want this to be a 2 step process:

* First, we introduce the `validateBulkLoadParameters` flag, but keep defaulting to `false`. (minor release)
* Then, we switch the default to `true`, and deprecate setting the `validateBulkLoadParameters` flag. (major release)
* Eventually, we drop the `validateBulkLoadParameters` flag. (major release)

oanylund:
Hi, was just about to report an issue but found this feature that looks like it will fix it. 
The added callback in the transform stream looks promising. I just spent one hour trying to catch an unhandled exception from the stream. 😄 
The compiled lib today adds a try catch around the "for of chunk" wich if fails and eventually throws an _iteratorError in the _write method of the RowTransform. Adding a callback with error allows better error handling in general from outside the library.

Compiled part:
`try {
        for (var _iterator = c.type.generateParameterData(parameter, this.mainOptions)[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
          const chunk = _step.value;
          this.push(chunk);
        }
      } catch (err) {
        _didIteratorError = true;
        _iteratorError = err;
      } finally {
        try {
          if (!_iteratorNormalCompletion && _iterator.return != null) {
            _iterator.return();
          }
        } finally {
          if (_didIteratorError) {
            **throw _iteratorError;**
          }
        }
      }`

In the Node.js documentation this is mentioned: "Throwing an Error from within these methods or manually emitting an 'error' event results in undefined behavior."

test/integration/bulk-load-test.js
```diff
@@ -351,7 +351,7 @@ describe('Bulk Load Tests', function() {
   });
 
   it('should test stream bulk load', function(done) {
-    this.timeout(50000);
+    this.timeout(200000);
```
MichaelSun90, line 354: change this because of that this test always gives me a time out error. Increase this make the test pass.

PR #730, opened 2018-04-15
Feature: initialise `Connection` without opening a connection and open function exposed

joshstevens19:
Hi guys, love the library just starting integrating with it the create full blown node package for SQL server. You probably will see me expanding on this library a lot. 

The feature i have put in is the ability to create a `Connection` instance without creating a connection straight away. I know there is times when i may want to create the connection instance but not open the connection until i physically have too. I think we should allow people to make that judgement in when there database connection is going to be opened. This ability gives flexibility for people who want to do both.

I wanted this because my library is going to work similar theme of c# `system.data` where you have to physically open the connection when you want too. I made it a optional parameter with default true so it will not break anything for anyone. 

I also want to add an `open` method to your types library (if you have one).

Thanks guys!
Josh

PR #1069, opened 2020-03-15
Deprecate automatically opening a connection when building a `Connection` instance

arthurschreiber:
In https://github.com/tediousjs/tedious/pull/730, the ability to build a new `Connection` object without immediately opening the actual connection to the server was requested.

I think this requested behavior should actually be the **default**. This is also how `net.Socket` in NodeJS works - calling `new net.Socket` won't actually open a connection to the remove server, unless the `.connect` method is called on the returned socket object. The `net` module also provides a module level `.connect` convenience method that creates and returns a new `Socket` instance and calls `.connect` on it.

This pull request adds a similar method on the `tedious` module, that creates a new `Connection`, calls `.connect` on it and returns the created connection object.

Calling `new Connection` without immediately calling `.connect` on it will emit a new deprecation warning, but will automatically call `.connect` to emulate the existing behavior. This existing behavior will be removed in the future, and not calling `.connect` will cause the connection to never be opened.

Fixes: https://github.com/tediousjs/tedious/pull/730

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -204,7 +204,7 @@ interface State {
   };
 }
 
-interface ConnectionConfiguration {
+export interface ConnectionConfiguration {
   server: string;
   options?: ConnectionOptions;
   authentication?: {
@@ -290,6 +290,7 @@ class Connection extends EventEmitter {
   ntlmpacketBuffer: undefined | Buffer;
 
   STATE!: {
+    INITIALIZED: State;
     CONNECTING: State;
     SENT_PRELOGIN: State;
     REROUTING: State;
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -979,6 +980,26 @@ class Connection extends EventEmitter {
...
+      if (this.state === this.STATE.INITIALIZED) {
+        const message = 'In the next major version of `tedious`, creating a new ' +
+          '`Connection` instance will no longer establish a connection to the ' +
+          'server automatically. Please use the new `connect` helper function or ' +
+          'call the `.connect` method on the newly created `Connection` object to ' +
+          'silence this message.';
+        deprecate(message);
+        this.connect();
+      }
+    });
```
arthurschreiber: This is a neat "trick" - if `new Connection()` is called without a call to `.connect` in the same event loop tick, `.connect` will be called automatically, but there will also be a deprecation message emitted. This ensures existing code continues working as-is.

The deprecation message can easily be silenced by calling `.connect` right after calling `new Connection()`, allowing users to migrate their code to the new API before upgrading to the next major version.

src/connection.ts
```diff
@@ -979,17 +980,59 @@ class Connection extends EventEmitter {
...
+  }
+
+  connect(connectListener?: (err?: Error) => void) {
+    if (this.state !== this.STATE.INITIALIZED) {
+      throw new ConnectionError('`.connect` can not be called on a Connection in `' + this.state.name + '` state.');
+    }
```
arthurschreiber: This is something I'm not 100% sure about. `net.Socket` allows calling `.connect` multiple times (for whatever reason 🤷‍♂).

I'd argue that calling `.connect` more than once is a programmer error, and thus throwing an exception is the correct way to handle this.

PR #1071, opened 2020-03-19
AE: col metadata parser

fredwes:
Implement parsing for crypto metadata. These changes will allow for parsing of Always Encrypted metadata as part of the column metadata parser, if the Always Encrypted feature is enabled on the connection.
This is another PR of the effort to break down the large Always Encrypted PR into smaller, more manageable ones for review. #1020

*Re-opened


src/token/colmetadata-token-parser.ts
```diff
@@ -56,48 +57,230 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+}
+
+function readCryptoMetadataOrdinal(parser: Parser, columnEncryptionKeys: EncryptionKeyInfo[][] | undefined, callback: (ordinal: number) => void) {
+  if (!columnEncryptionKeys) {
+    // Ordinal is not provided for SQL function return values
+    callback(0);
+  } else {
+    parser.readUInt16LE((ordinal) => {
+      callback(ordinal);
+    });
```
arthurschreiber, line 144:
```suggestion
    parser.readUInt16LE(callback);
```

src/token/colmetadata-token-parser.ts
```diff
@@ -56,48 +57,230 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+  });
+}
+
+function readKeyStoreName(parser: Parser, callback: (keyStoreName: string) => void) {
+  parser.readUInt8((keyStoreNameLength) => {
+    parser.readBuffer(2 * keyStoreNameLength, (keyStoreNameBuffer) => {
+      callback(keyStoreNameBuffer.toString('ucs2'));
+    });
+  });
```
arthurschreiber, line 161:
```suggestion
  parser.readBVarChar(callback);
```

src/token/colmetadata-token-parser.ts
```diff
@@ -56,48 +57,230 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+}
+
+function readKeyPath(parser: Parser, callback: (keyPath: string) => void) {
+  parser.readUInt8((keyPathLength) => {
+    parser.readBuffer(2 * keyPathLength, (keyPathBuffer) => {
+      callback(keyPathBuffer.swap16().toString('ucs2'));
+    });
+  });
```
arthurschreiber, line 169:
```suggestion
  parser.readBVarChar(callback);
```

src/token/colmetadata-token-parser.ts
```diff
@@ -56,48 +57,230 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+}
+
+function readAlgorithmName(parser: Parser, callback: (algorithmName: string) => void) {
+  parser.readUInt16BE((algorithmNameLength) => {
+    parser.readBuffer(2 * algorithmNameLength, (algorithmNameBuffer) => {
+      callback(algorithmNameBuffer.toString('ucs2'));
+    });
+  });
```
arthurschreiber, line 177:
```suggestion
  parser.readUsVarChar(callback);
```

src/connection.ts
```diff
@@ -175,6 +175,7 @@ export interface InternalConnectionOptions {
   requestTimeout: number;
   rowCollectionOnDone: boolean;
   rowCollectionOnRequestCompletion: boolean;
+  serverSupportsColumnEncryption: boolean;
```
arthurschreiber, line 178: Is this connection level option required? 🤔 I see there is another flag returned with the column metadata flags that should be enough to distinguish whether encrypted data is send or not.

src/metadata-parser.ts
```diff
@@ -102,223 +107,247 @@ function readUDTInfo(parser: Parser, callback: (udtInfo: UdtInfo | undefined) =>
...
-                scale: undefined,
-                dataLength: undefined,
-                schema: undefined,
-                udtInfo: udtInfo
-              });
-            });
+function parseFlags(parser: Parser, callback: (flags: number) => void) {
+  parser.readUInt16LE((flags) => {
+    callback(flags);
+  });
```
arthurschreiber, line 330:
```suggestion
  parser.readUInt16LE(callback);
```

src/token/colmetadata-token-parser.ts
```diff
@@ -56,48 +57,230 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+  // AEAD_AES_256_CBC_HMAC_SHA512
+  if (cipherAlgorithmId !== 0) {
+    return callback();
+  }
+
+  parser.readUInt8((nameSize) => {
+    parser.readBuffer(nameSize, (algorithmNameBuffer) => {
+      callback(algorithmNameBuffer.toString('ucs2'));
+    });
+  });
```
arthurschreiber, line 134:
```suggestion
  parser.readBVarChar(callback);
```

src/token/colmetadata-token-parser.ts
fredwes: `parser.readBVarChar(callback);` doesn't do the `swap16` needed here

src/token/colmetadata-token-parser.ts
fredwes: `parser.readUsVarChar` doesn't do `readUInt16BE` needed here

src/connection.ts
fredwes: That flag is returned on a per column basis (is this an encrypted column or not).

This option is at connection level because we need to know to parse the encryption keys (for the result set as a whole) before we parse the metadata for each column. If the server has negotiated the always encrypted feature to be enabled when the connection initializes then this flag should be set to true on the connection.

PR #1084, opened 2020-04-18
fix: DateTime2 nanosecondsDelta property name

bitman-decodelog:
nanosecondsDelta property name was invalid


arthurschreiber:
Thanks! The change looks good from a quick glance. I'll double check why the build is failing.

hlobil (approved):

Issue #1074, opened 2020-04-06
RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: Attempt to write outside buffer bounds


arpitgupta2405:
Hi @IanChokS , I guess you haven't read my issue clearly (https://github.com/tediousjs/tedious/issues/1035).

I'm trying to connect to MSSQL Server only. But if I specify the port number as 3306, I get an uncaught RangeError. How do I catch that error?

IanChokS:
@arpitgupta2405 You mentioned that 

> I have Mysql running on the port 3306

That's why I closed the issue, because Tedious isn't built for MySQL Server.

arpitgupta2405:
@IanChokS , the problem is we take the connection config from the end user and there are times when end user specifies port as 3306 by mistake for the mssql connection, so in that case my server breaks.

So it will be better if you can add try and catch for this error, so it becomes easy for me to handle this case and my server doesn't break.

Let me know if there is still some confusion.

IanChokS:
Have you implemented error handling in your application to catch errors that tedious throws? 

I think the error is being thrown when establishing a `new connection(config)` in tedious. Maybe try to do a `try catch` around that to see if your able to catch it?

arpitgupta2405:
Hi @IanChokS , I have all the error handlers in place for tedious and even tried using try and catch for new connection(config) but none of those worked.

IanChokS:
Can you show what error handlers you have for tedious?

arpitgupta2405:
const { Request, Connection } = require('tedious')

const connectionConfig = {
  authentication: { 
    type: 'default',
    options: {
      userName: '<my user name>',
      password: '<my password>'
    }
  },
  server: '<sever host>',
  options: {
    database: '<db>',
    encrypt: true,
    port: 3306,
    packetSize: 32768
  }
}

const connection = new Connection(connectionConfig)

function rowHandler () {
  // save the result
}

function errorHandler () {
  // handle the error
}

let queryToRun = 'sql query'

let request = new Request(queryToRun, (err, rowCount) => {
  if (err) {
    // handle error
  }
  // cleaning up the connection
})

request.on('error', errorHandler)

request.on('row', rowHandler)

connection.on('error', errorHandler)

connection.execSql(request)

arpitgupta2405:
Is there any update?

arpitgupta2405:
@IanChokS , is there nay update on this?

IanChokS:
[...]

PR #1101, opened 2020-05-28
Added an additional check to prevent ERR_OUT_OF_BOUNDS error

arpitgupta2405:
**Before submitting a PR :**

### DIFF
--- a/src/incoming-message-stream.ts
+++ b/src/incoming-message-stream.ts
@@ -49,7 +49,7 @@ class IncomingMessageStream extends Transform {
       // Get the full packet length
       const length = this.bl.readUInt16BE(2);
 
-      if (this.bl.length >= length) {
+      if (this.bl.length >= length && length >= HEADER_LENGTH) {
         const data = this.bl.slice(0, length);
         this.bl.consume(length);
 


arthurschreiber:
Thanks for this contribution! Is there an issue open already that describes when this was happening? Or can you provide a reproducible test case that shows what this is fixing exactly?

arthurschreiber:
Ah, perfect, I see you referenced this PR in an existing issue. Thanks!

arpitgupta2405:
I was facing https://github.com/tediousjs/tedious/issues/1074 and I did some research on tedious and found that this PR fixes the issue.

arpitgupta2405:
How do I restart the checks?

arthurschreiber:
@arpitgupta2405 👋 Usually the build restarts whenever you push a new change (or perform force push). Don't worry about the failures due to commit message linting - I can fix this on my side when merging the changes.

src/incoming-message-stream.ts
```diff
@@ -49,7 +49,7 @@ class IncomingMessageStream extends Transform {
       // Get the full packet length
       const length = this.bl.readUInt16BE(2);
 
-      if (this.bl.length >= length) {
+      if (this.bl.length >= length && length >= HEADER_LENGTH) {
```
arthurschreiber, line 52: I'm wondering if it would make sense to call `callback` with an error here, if `length` is either smaller than `HEADER_LENGTH` or larger than the previously negotiated packet length? 🤔

arpitgupta2405: I don't think we need a callback with an error at this place.

arthurschreiber: The stream API expects errors to be returned back via the `callback` that is passed to the `_transform` method. This callback is passed from `_transform` down to `processBufferedData`.

If the length we read from the packet is invalid, this means the whole stream is "corrupted" and the connection should be shut down and for sure can't be used anymore. The proper way to signal this back is to call `callback` with an `Error` parameter and return from  any further processing in `processBufferedData`. This way the error gets bubbled up to wherever the `IncomingMessageStream` is used.

Optimally, this should then cause the connection to be closed properly, but I'm not sure if that's going to happen without testing this properly.

With the fix you put in place here, the initial error you encountered is no longer happening (so the Node process is not going to crash), but I think this is silently corrupting the connection and making it stuck, until the underlying socket times out or the server side closes the connection forcefully.

Does this make sense? 🙇‍♂️

arpitgupta2405: Making sense now. I had to manually close the connection on the server side.

arthurschreiber:
Closing this one out in favor of #1102.

PR #1102, opened 2020-05-28
fix: correctly handle invalid packet headers

IanChokS:
When the server sends invalid packet header data, we will now correctly pass a `ConnectionError` to the `callback` given to `TranformStream._transform`.

Fixes: #1074
References: #1101

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1703,7 +1703,18 @@ class Connection extends EventEmitter {
     }
 
     if (connectListener) {
-      this.once('connect', connectListener);
+      const onConnect = () => {
+        this.removeListener('error', onError);
+        connectListener();
+      };
+
+      const onError = (err: Error) => {
+        this.removeListener('connect', onConnect);
+        connectListener(err);
+      };
+
+      this.once('connect', onConnect);
+      this.once('error', onError);
     }
 
     this.transitionTo(this.STATE.CONNECTING);
@@ -2217,6 +2228,9 @@ class Connection extends EventEmitter {
       this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
       this.messageIo.on('message', () => { this.dispatchEvent('message'); });
       this.messageIo.on('secure', (cleartext) => { this.emit('secure', cleartext); });
+      this.messageIo.on('error', (error) => {
+        this.socketError(error);
+      });
 
       this.socket = socket;
       this.socketConnect();
--- a/src/incoming-message-stream.ts
+++ b/src/incoming-message-stream.ts
@@ -4,6 +4,7 @@ import { Transform } from 'readable-stream';
 import Debug from './debug';
 import Message from './message';
 import { Packet, HEADER_LENGTH } from './packet';
+import { ConnectionError } from './errors';
 
 /**
   IncomingMessageStream
@@ -43,11 +44,14 @@ class IncomingMessageStream extends Transform {
     return this;
   }
 
-  processBufferedData(callback: () => void) {
+  processBufferedData(callback: (err?: ConnectionError) => void) {
     // The packet header is always 8 bytes of length.
     while (this.bl.length >= HEADER_LENGTH) {
       // Get the full packet length
       const length = this.bl.readUInt16BE(2);
+      if (length < HEADER_LENGTH) {
+        return callback(new ConnectionError('Unable to process incoming packet'));
+      }
 
       if (this.bl.length >= length) {
         const data = this.bl.slice(0, length);
--- a/test/integration/invalid-packet-stream-test.js
+++ b/test/integration/invalid-packet-stream-test.js
@@ -1,0 +1,58 @@
+const { assert } = require('chai');
+const net = require('net');
+const Connection = require('../../src/tedious').Connection;
+const ConnectionError = require('../../src/errors').ConnectionError;
+
+describe('Connecting to a server that sends invalid packet data', function() {
+  let server, sockets;
+
+  beforeEach(function(done) {
+    sockets = [];
+    server = net.createServer();
+    server.listen(0, '127.0.0.1', done);
+    server.on('connection', (socket) => {
+      sockets.push(socket);
+    });
+  });
+
+  afterEach(function(done) {
+    // Clean up all leftover sockets
+    for (const socket of sockets) {
+      socket.destroy();
+    }
+    server.close(done);
+  });
+
+  it('should throw Connection Error ', function(done) {
+    server.once('connection', (socket) => {
+      const packetData = Buffer.from('test1234');
+
+      const packetHeader = Buffer.alloc(8);
+      let offset = 0;
+      offset = packetHeader.writeUInt8(0x11, offset);
+      offset = packetHeader.writeUInt8(0x01, offset);
+      offset = packetHeader.writeUInt16BE(5, offset);
+      offset = packetHeader.writeUInt16BE(0x0000, offset);
+      offset = packetHeader.writeUInt8(1, offset);
+      packetHeader.writeUInt8(0x00, offset);
+
+      const packet = Buffer.concat([packetHeader, packetData]);
+      socket.write(packet);
+    });
+
+    const addressInfo = server.address();
+    const connection = new Connection({
+      server: addressInfo.address,
+      options: {
+        port: addressInfo.port,
+      }
+    });
+
+    connection.connect((err) => {
+      assert.instanceOf(err, ConnectionError);
+      assert.equal(err.message, 'Connection lost - Unable to process incoming packet');
+
+      done();
+    });
+  });
+});
--- a/test/unit/incoming-message-stream-test.js
+++ b/test/unit/incoming-message-stream-test.js
@@ -4,6 +4,7 @@ const { assert } = require('chai');
 const IncomingMessageStream = require('../../src/incoming-message-stream');
 const Message = require('../../src/message');
 const Debug = require('../../src/debug');
+const ConnectionError = require('../../src/errors').ConnectionError;
 
 describe('IncomingMessageStream', function() {
   it('extract messages from packet data', function(done) {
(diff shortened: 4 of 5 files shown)


src/incoming-message-stream.ts
```diff
@@ -43,13 +43,13 @@ class IncomingMessageStream extends Transform {
...
 
-  processBufferedData(callback: () => void) {
+  processBufferedData(callback: (err?: Error) => void) {
     // The packet header is always 8 bytes of length.
     while (this.bl.length >= HEADER_LENGTH) {
       // Get the full packet length
       const length = this.bl.readUInt16BE(2);
 
-      if (this.bl.length >= length) {
+      if (this.bl.length >= length && length >= HEADER_LENGTH) {
```
arthurschreiber, line 52: 🤔 What happens if `this.bl.length` is smaller than `length` if not enough data has been received from the server side yet?

src/incoming-message-stream.ts
```diff
@@ -82,7 +82,7 @@ class IncomingMessageStream extends Transform {
           return;
         }
       } else {
-        break;
+        return callback(new Error('Unable to process incoming buffer'));
```
arthurschreiber, line 85: I think we should return a `ConnectionError` here to be a bit more specific, instead of a generic `Error`.

IanChokS:
@arthurschreiber Tests have been added ☝️

src/incoming-message-stream.ts
arthurschreiber: Ok, so I double checked, and `this.bl.length` is always greater than `HEADER_LENGTH` (due to the condition in the `while` statement above).

I still think there's something not _quite_ right with this condition. Let's say the so-far received buffer length (`this.bl.length`) is `20` and `length` read from the header is `40`. In this case, we would also incorrectly call `callback` with an error, while before we correctly used `break` to exit the loop and wait for more data to come in.

I think what we need to do here is not change this `if` (or it's `else` block) at all, and instead put a condition right before it that checks that `length` is at least `HEAD_LENGTH`.

Something like this:

```js
if (length < HEADER_LENGTH) {
  return callback(new ConnectionError('Unable to process incoming packet'));
}
```

test/unit/incoming-message-stream-test.js
```diff
@@ -175,4 +176,49 @@ describe('IncomingMessageStream', function() {
...
+
+    const packet = Buffer.concat([packetHeader, packetData]);
+
+    const incoming = new IncomingMessageStream(new Debug());
+
+    assert.throws(() => {
+      incoming.end(packet);
+    }, ConnectionError, 'Unable to process incoming packet');
+
+    done();
```
arthurschreiber, line 199: So, this test "works", but there are some slight issues with it.

For one, we want to test whether an exception is emitted via the `error` event, because that is what happens if the `callback` is called with an error.

The reason why catching the error via `assert.throws` works here is because this ends up being an uncaught exception. When the packet is parsed, the callback is called with the error, but because there is no handler on the `error` event, the error is being thrown, which is caught by `assert.throws`, and the test passes.

If you'd change the `callback` invocation to `throw` directly inside `IncomingMessageStream`, the test would still pass, but if that condition would be hit in a real application, the application would still break due to an uncaught exception! So the test is giving us a false sense of confidence into our code being "correct".

What would be better is if we'd test that the `error` event is emitted, like this:

```suggestion
    incoming.on('error', (err) => {
      assert.instanceOf(err, ConnectionError);
      assert.equal(err.message, 'Unable to process incoming packet');

      done();
    });

    incoming.end(packet);
```

test/unit/incoming-message-stream-test.js
```diff
@@ -175,4 +176,49 @@ describe('IncomingMessageStream', function() {
...
+    offset = packetHeader.writeUInt8(1, offset);
+    packetHeader.writeUInt8(0x00, offset);
+
+    const packet = Buffer.concat([packetHeader, packetData]);
+
+    const incoming = new IncomingMessageStream(new Debug());
+
+    assert.throws(() => {
+      incoming.end(packet);
+    }, ConnectionError, 'Unable to process incoming packet');
```
arthurschreiber, line 220: Same as above.

test/unit/incoming-message-stream-test.js
```diff
@@ -175,4 +176,49 @@ describe('IncomingMessageStream', function() {
...
+
+    done();
+  });
```
arthurschreiber, line 223: I think this test here is actually showing the issue I mentioned over in https://github.com/tediousjs/tedious/pull/1102/files#r432504620.

The packet that is written here is fine. The packet says it's length is 18 bytes, but we're only writing 16 bytes of data here. In this case, the `IncomingMessageStream` should actually not emit an error - it should be waiting for more data coming in from the network instead.

This was working before, because of the `break` statement which would stop processing and wait for more data to come in.

test/unit/incoming-message-stream-test.js
IanChokS: I did try incoming.on('error'....) but the test gets hanged and the event 'error' doesn't get caught or isn't being emitted.

test/unit/incoming-message-stream-test.js
IanChokS: I'm just a bit confused. If the packet length specifies 18 bytes, wouldn't that mean that the current packet must be 18 bytes long? I'm assuming that the next data coming in is a new packet with a new packet header.

> 2.2.3.1.3 Length
> Length is the size of the packet including the 8 bytes in the packet header. It is the number of bytes
> from the start of this header _to the start of the next packet header._
>

src/connection.ts
```diff
@@ -1369,6 +1369,9 @@ class Connection extends EventEmitter {
       this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
       this.messageIo.on('message', () => { this.dispatchEvent('message'); });
       this.messageIo.on('secure', (cleartext) => { this.emit('secure', cleartext); });
+      this.messageIo.on('error', (message) => {
+        this.emit('error', message);
+      });
```
arthurschreiber, line 1374: I think we need to also dispatch a `socketError` event here so the connection gets cleaned up properly? 🤔 

```suggestion
      this.messageIo.on('error', (error) => {
        this.emit('error', error);
        this.dispatchEvent('socketError', error);
      });
```

test/unit/incoming-message-stream-test.js
arthurschreiber: > If the packet length specifies 18 bytes, wouldn't that mean that the current packet must be 18 bytes long?

Yes, but a packet of 18 bytes could be delivered to the stream in any number of chunks. It could be two chunks being 9 bytes, or 4 chunks (1 byte, 2 bytes, 3 bytes, 13 bytes), or 18 chunks each containing 1 byte, or whatever other combination you could come up with. This depends on the underlying transport and the operating systems kernel. The only guarantee we have is that the data comes in in the same order as sent by the server, but there is no guarantee around the "chunkiness" of the arriving data.

test/integration/invalid-packet-stream-test.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      });
+      connection.connect((err) => {});
+  
+      connection.on('error', (err) => {
+        assert.instanceOf(err, ConnectionError);
+        assert.equal(err.message, 'Unable to process incoming packet');
+        connection.close();
+        done();
+      })
```
arthurschreiber, line 47: Something is not quite right here. 🤔 

There should be an `err` being passed to the callback given to `connect`, but no error emitted via the  `error` event. We currently only emit `error` events if the connection was established successfully, and I think we have to continue doing that.

Maybe instead of doing

```ts
        this.emit('error', error);
        this.dispatchEvent('socketError', error);
```

we should be just calling `socketError`, like this:

```ts
        this.socketError(error);
```

Can you check if that would change the behavior to what we want? 🙇‍♂️

IanChokS: `connection.connect(cb)` calls the callback when the first `connect` event is emitted. So `this.emit('connect', error)` seems to do the trick.

src/connection.ts
```diff
@@ -1369,6 +1369,9 @@ class Connection extends EventEmitter {
       this.messageIo.on('data', (data) => { this.dispatchEvent('data', data); });
       this.messageIo.on('message', () => { this.dispatchEvent('message'); });
       this.messageIo.on('secure', (cleartext) => { this.emit('secure', cleartext); });
+      this.messageIo.on('error', (error) => {
+        this.emit('connect', error)
```
arthurschreiber, line 1373: @IanChokS I think there's still a problem with this. 🙇‍♂️ 

If a protocol error happens when establishing a connection, this will emit a `connect` event, but I don't think this will correctly close the network socket or perform any other cleanup.

On top of that, if a protocol error happens _after_ establishing a connection initially, this will fire a second, unexpected `connect` event, and most likely lead to really weird behavior in code that uses `tedious`.

I still think calling the `socketError` is probably the correct thing to do. It will emit a `connect` event with an error if the connection was not fully established yet, and will emit an `error` if the connection was established already. It will also dispatch the `socketError` internal event that will cause the connection to switch to the `FINAL` state, which will ensure the underlying network socket will get closed correctly.

test/integration/invalid-packet-stream-test.js
```diff
@@ -0,0 +1,48 @@
...
+        server: addressInfo.address,
+        options: {
+          port: addressInfo.port,
+        }
+      });
+
+      connection.connect((err) => {
+        assert.instanceOf(err, ConnectionError);
+        assert.equal(err.message, 'Unable to process incoming packet');
+        connection.close();
```
arthurschreiber, line 44: You shouldn't have to call `connection.close`. As mentioned in https://github.com/tediousjs/tedious/pull/1102/files#r441797042, the correct behavior will be that the connection gets closed automatically.

IanChokS:
@arthurschreiber ☝️ Updated to call `socketError`

PR #1103, opened 2020-05-28
Stop stubbing `dns` module

arthurschreiber:
This removes the stubbing of `dns.lookup` we had in place in some of our test cases.

This stubbing was error prone and lead to a whole bunch of flaky test failures on CI.

Instead of stubbing functions, this changes relevant code to allow passing in a custom `lookup` function, but falling back to `dns.lookup` if that function is `undefined`.

Closes: https://github.com/tediousjs/tedious/pull/988

### DIFF
--- a/src/connector.ts
+++ b/src/connector.ts
@@ -45,9 +45,11 @@ export class ParallelConnectionStrategy {
     }
 
     for (let i = 0, len = addresses.length; i < len; i++) {
-      const socket = sockets[i] = net.connect(Object.create(this.options, {
-        host: { value: addresses[i].address }
-      }));
+      const socket = sockets[i] = net.connect({
+        ...this.options,
+        host: addresses[i].address,
+        family: addresses[i].family
+      });
 
       socket.on('error', onError);
       socket.on('connect', onConnect);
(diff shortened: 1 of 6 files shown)


arthurschreiber:
I _think_ this fixes the flakiness that https://github.com/tediousjs/tedious/pull/988 also tried to fix, but in a different way. 😅 

The problem that https://github.com/tediousjs/tedious/pull/988 worked around was that sometimes the `callback` given to the `InstanceLookup` or `Connector` classes could be fired twice. On the first call, the current test would fail (because it did not expect an error), and then after some time the `callback` was called again causing whatever test was running at that time to fail as well. https://github.com/tediousjs/tedious/pull/988 worked around this by moving the assertions in the test into the non-async part of the test case.

This fixed the test flakyness, but did not fix the problems in the underlying code.

I believe the changes in this PR in the `src/instance-lookup.ts` fix the problem in the code because they slightly change how and when `setTimeout`/`clearTimeout` is called, and this should prevent the double callback execution.

But I think there is still problems with this code in general.

There is no way to cancel any of the in-flight operations (e.g. the dns lookups) if the `connectTimeout` has been reached or the connection is prematurely closed. So what users sometimes see is that the timeout fires, and then some time later another error is thrown again. I'll try to fix this in follow up PRs.

Issue #1081, opened 2020-04-16
Hostname/IP does not match certificates when trustServerCertificate=false


heyjared:
As per the recent deprecation notice introduced in #1029, I have made the change to explicitly set this to false. However, when attempting to connect to an Azure SQL Server in an Elastic Pool, tedious reports the following error:

`HostnameIP does not match certificates altnames Host <redacted>.database.windows.net is not in the certs altnames <redacted>.worker.database.windows.net <redacted>.worker.database.windows.net <redacted>.worker.database.windows.net <redacted>.worker.database.windows.net <redacted>.worker.database.windows.net`

The connection string recommended by the Azure portal includes this option which is missing from the available options in connection options in tedious.

`hostNameInCertificate=*.database.windows.net`

arthurschreiber:
@heyjared 👋 

Thanks for letting us know about this!

I think this is an issue with the re-routing code. I think Elastic Pools work by re-routing the connection from the initial server (`*.database.windows.net`) to target hosts (`*.worker.database.windows.net`), but we don't correctly set the target hosts as the expected host name for the TLS connection, which in turn causes the issue you're seeing.

A fix should not be too hard to do, I'll see if I can find some time to fix this!

Sebbish:
I just ran into this issue with Azure SQL not running elastic pool after making the same change. Changing server connection policy from redirect to proxy fixed it, but is slower. The altnames I got was workers as well.

arthurschreiber:
https://github.com/tediousjs/tedious/pull/1116 is running into the same issue. The problem is that when we follow a server redirect, we still expect the initial server name when performing the TLS connection, which then fails, as this does not match the hostname that the redirect points at.

The fix should be straightforward. I'll open a PR soon. 🙇‍♂️

clshortfuse:
I'm on v9.2.1 and I still have to force `trustServerCertificate: true`. The documentation still wrong, it seems.

Without the option set, I get: `unable to get local issuer certificate`, but enabled it works.

PR #1116, opened 2020-06-22
feat: change default value of `trustServerCertificate`

arthurschreiber:
BREAKING CHANGE: This changes the default value from `true` to `false`. This might lead to encrypted connections not being established if the server's certificate can not be validated against the default list of CAs.

You can either explicitly set `trustServerCertificate` to `true` (not recommended, as it disables certificate validation) or specify a custom CA list via the `cryptoCredentialsDetails` connection option.

References: https://github.com/tediousjs/tedious/issues/983


arthurschreiber:
This is currently blocked by the same issue as described here: https://github.com/tediousjs/tedious/issues/1081

PR #1117, opened 2020-06-22
fix: use proper servername for TLS after redirect

arthurschreiber:
This fixes an issue where after a server redirect, we would continue to use the original servername when trying to establish a TLS connection. This issue is only visible when `trustServerCertificate` was explicitly set to `false`, as otherwise the hostname returned by the certificate was never verified.

Fixes: https://github.com/tediousjs/tedious/issues/1081

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2201,7 +2201,7 @@ Connection.prototype.STATE = {
             return this.close();
           }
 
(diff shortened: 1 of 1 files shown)


arthurschreiber:
@IanChokS @MichaelSun90 I'm not sure we can write tests for this easily, but I verified this fixes the issue described in https://github.com/tediousjs/tedious/issues/1081 and should make the tests for https://github.com/tediousjs/tedious/pull/1116 pass.

PR #1086, opened 2020-04-23
Add error listenter to readable stream

dhensby:
Fixes #1085 

Currently the readstream that is created doesn't have any error handlers attached, so if the value parsing errors the emitted error is not caught causing node to crash.

This catches the error and passes it to the request so it can be handled.

Issue #1085, opened 2020-04-22
Invalid uniqueidentifiers cause node to crash


dhensby:
At the moment if the database contains an invalid UniqueIdentifier value the TDS request errors but the error it not passed to the callback.

This causes an uncaught error which means node crashes and the developer cannot handle this exception elegantly.

I've not been able to find the source of the problem in the TS code (sorry) but this affects v6+.

Please see https://github.com/tediousjs/node-mssql/issues/1001 for context.

This may be a wider problem in `writeParameterData` for all `DataType` objects so there may be a fairly generic fix for this.

PR #1122, opened 2020-06-28
fix: improve `UniqueIdentifier` data type validation

arthurschreiber:
Fixes: https://github.com/tediousjs/tedious/issues/1085
Closes: https://github.com/tediousjs/tedious/pull/1086

---

@dhensby Can you take a look? 🙇‍♂️

### DIFF
--- a/src/data-types/uniqueidentifier.ts
+++ b/src/data-types/uniqueidentifier.ts
@@ -34,12 +34,19 @@ const UniqueIdentifier: DataType = {
     if (value == null) {
       return null;
     }
+
     if (typeof value !== 'string') {
       if (typeof value.toString !== 'function') {
         return TypeError('Invalid string.');
       }
+
       value = value.toString();
     }
+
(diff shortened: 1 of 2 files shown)


dhensby:
This looks like it would solve the specific problem that is being reported regarding GUIDs, but not the more generic issue of errors being thrown in the stream, which my original patch tried to handle.

arthurschreiber:
> This looks like it would solve the specific problem that is being reported regarding GUIDs, but not the more generic issue of errors being thrown in the stream, which my original patch tried to handle.

@dhensby I think I mentioned this before, but the patch you provided would've left the connection in an undefined state. I'll see what I can do about better error handling, but I think the proper approach is to actually fix the code so that no errors happen in the first place.

arthurschreiber:
@dhensby Would you like to see this PR to be backported to some older versions? 🙇‍♂️

dhensby:
Thanks @arthurschreiber! In terms of backporting, mssql is still using v6 and our v7 alpha is using v8 (but that could be bumped to v9 before release).

I doubt you really want to back port this to v6, but it would allow us to tick this off as fixed in our current stable version. Otherwise I'll just direct people to use v7.

PR #1131, opened 2020-07-27
feat: add options.workstationId

aklietz:
Add `options.workstationId` for client logging in SQL Server when the actual client is behind a proxy. The value is reported by the TSQL function `HOST_NAME()`. 

This is useful for websites that want to pass through the hostname or IP address of the HTTP request for logging in SQL Server.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -367,6 +367,7 @@ export interface InternalConnectionOptions {
   trustServerCertificate: boolean;
   useColumnNames: boolean;
   useUTC: boolean;
+  workstationId: undefined | string;
   lowerCaseGuids: boolean;
 }
 
@@ -805,6 +806,15 @@ interface ConnectionOptions {
   useUTC?: boolean;
 
   /**
+   * The workstation ID (WSID) of the client, default os.hostname().
+   * Used for identifying a specific client in profiling, logging or
+   * tracing client activity in SQLServer.
+   *
+   * The value is reported by the TSQL function HOST_NAME().
+   */
+  workstationId?: string | undefined;
+
+  /**
    * A boolean determining whether to parse unique identifier type with lowercase case characters.
    *
    * (default: `false`).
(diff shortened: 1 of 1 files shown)


src/connection.ts
```diff
@@ -654,6 +655,13 @@ interface ConnectionOptions {
    */
   fallbackToDefaultDb?: boolean;
 
+  /**
+   * The hostname of the client, default os.hostname(). Used for
+   * logging in SQL Server when the client is behind a proxy.
+   * The value is reported by the TSQL function HOST_NAME().
+   */
+  hostname?: string | undefined;
```
arthurschreiber, line 663: This option is called `workstation id` in other drivers (e.g. in the Java Driver for MSSQL). I think `hostname` as an option name can be confused easily with `servername`, so maybe `workstationId` would be more fitting here?

@IanChokS @MichaelSun90 What do you think?

aklietz: I agree with @arthurschreiber .  Using hostname will likely cause confusion.  It should be workstationId.

For consistency the patch should also change hostname to workstationId in the TDS packet internals, login7-payload.ts. The latter  is not user-visible anywhere so I don't think it would will break anything, but I want to mention it just in case.

Do you want me to close the pull request and resubmit it?

arthurschreiber: It is specified as `hostname` in the `LOGIN7` part of the MS-TDS specification: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/773a62b6-ee89-4c02-9e5e-344882630aac

Because that part of the code is internal, leaving it as `hostname` is probably preferable, just to keep closter to the specification? What do you think?

aklietz: Yeah leave it.

arthurschreiber: Ok, just to recap:

* Change the option to `workstationId` in `ConnectionOptions` and `InternalConnectionOptions`. This will make sure we stick with the name used by other drivers and there is no confusion with `serverName`.
* Leave the name of the `hostname` option for the `login7-payload.ts` code as-is. It's internal code (so less room for confusion) and we match with what's specified in `MS-TDS`.

If you change the PR accordingly, I'll merge it asap.

aklietz: Done

PR #1140, opened 2020-08-10
feat: allow parameters to have collation options

arthurschreiber:
This was extracted from https://github.com/tediousjs/tedious/pull/1020

I've no idea how this could be tested. I tried sending `varchar` parameter values with different collations set, but it's not clear to me how to check what the collation of the parameter is on the server side.

@MichaelSun90 @IanChokS Any ideas how this could be tested?

---

@fredwes Would you mind taking a look?

### DIFF
--- a/test/integration/datatypes-in-results-test.ts
+++ b/test/integration/datatypes-in-results-test.ts
@@ -584,3 +584,90 @@ describe('Datatypes in results test', function() {
     ]
   });
 });
+
+describe('Collation LCID test', function() {
+  let connection: Connection;
+
+  beforeEach(function(done) {
+    connection = new Connection(config);
+
+    connection.on('errorMessage', function(error) {
+      console.log(`${error.number} : ${error.message}`);
+    });
+
+    connection.on('debug', function(message) {
+      if (debug) {
+        console.log(message);
+      }
+    });
+
+    connection.connect(done);
+  });
+
+  afterEach(function(done) {
+    if (!connection.closed) {
+      connection.on('end', done);
+      connection.close();
+    } else {
+      done();
+    }
+  });
+
+  it('should recieve correct column collation after sending different collation char and varchar', function(done) {
+    const remove_table_request = new Request('IF OBJECT_ID(\'dbo.temp_collation\', \'U\') IS NOT NULL DROP TABLE dbo.temp_collation;', (err) => {
+      if (err) {
+        return done(err);
+      }
+      const create_table_request = new Request('create table temp_collation (CharCol char(4) COLLATE Latin1_General_100_CI_AI_SC, CharCol2 varchar(4) COLLATE Latin1_General_100_CI_AI_SC)', (err) => {
+        if (err) {
+          return done(err);
+        }
+
+        const insert_sql = new Request('insert into temp_collation values (@p1, @p2)', (err) => {
+          if (err) {
+            return done(err);
+          }
+
+          const select_sql = new Request('select * from temp_collation ', (err) => {
+            if (err) {
+              return done(err);
+            }
+            return done();
+          });
+
+          select_sql.on('row', (columns) => {
+            columns.forEach((column: any) => {
+              assert.strictEqual(1033, column.metadata.collation.lcid);
+              assert.strictEqual('S?ao', column.value);
+            });
+          });
+
+          connection.execSql(select_sql);
+        });
+
+        insert_sql.addParameter('p1', TYPES.Char, 'こんにちは', {
+          collation: {
+            lcid: 1041,
+            flags: 0,
+            version: 0,
+            sortId: 0,
+          }
+        });
+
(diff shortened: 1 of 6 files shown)


test/integration/datatypes-in-results-test.ts
```diff
@@ -584,3 +584,90 @@ describe('Datatypes in results test', function() {
...
+              assert.strictEqual('S?ao', column.value);
+            });
+          });
+
+          connection.execSql(select_sql);
+        });
+
+        insert_sql.addParameter('p1', TYPES.Char, 'こんにちは', {
+          collation: {
+            lcid: 1041,
```
jeremy-w, line 650: it would be helpful when reading the test to have the referenced lcids use hex literals defined to a constant for the language tag. i've been converting them to hex to reference https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/63d3d639-7fd2-4afb-abbe-0d5b5551eef8 .

test/integration/datatypes-in-results-test.ts
```diff
@@ -584,3 +584,90 @@ describe('Datatypes in results test', function() {
...
+            if (err) {
+              return done(err);
+            }
+            return done();
+          });
+
+          select_sql.on('row', (columns) => {
+            columns.forEach((column: any) => {
+              assert.strictEqual(1033, column.metadata.collation.lcid);
+              assert.strictEqual('S?ao', column.value);
```
jeremy-w, line 641: is there any way to arrive at this result from reading docs & the TDS spec, or does it come from observed behavior?

i think this test might benefit from a column that enables an identity test: if `temp_collation` had a column `MatchingCollationCol varchar(15) COLLATE Japanese_CI_AS`, then inserting and reading back the parameter value should preserve the value without change.

test/integration/datatypes-in-results-test.ts
```diff
@@ -584,3 +584,90 @@ describe('Datatypes in results test', function() {
...
+  afterEach(function(done) {
+    if (!connection.closed) {
+      connection.on('end', done);
+      connection.close();
+    } else {
+      done();
+    }
+  });
+
+  it('should recieve correct column collation after sending different collation char and varchar', function(done) {
```
jeremy-w, line 616: typo: recieve => receive

```suggestion
  it('should receive correct column collation after sending different collation char and varchar', function(done) {
```

test/integration/datatypes-in-results-test.ts
arthurschreiber: I'm pretty sure there is something wrong with the tests here - I don't think that value is right.

The test you're describing is how this "should" be behaving - but it's currently not. 😞

arthurschreiber:
Okay, I think what happens is this:

We send the parameter as `varchar` with the `Chinese_PRC_CI_AS` collation (encoded with code page 936) to the server. The server takes this value, and converts it to `varchar` with the server's default collation `SQL_Latin1_General_CP1_CI_AS` (encoded with code page 1252), replacing all characters that can't be converted with `?`. The result is written to the `data` column, converting back from code page 1252 into code page 936.

I "guess" this is expected behaviour? 🙈

arthurschreiber:
Okay, some more testing. It looks like the input parameter collation is totally ignored. SQL Server "expects" all `varchar` data to be send in the encoding of the current database. So whether we send a collation value or just zeroes for `varchar` columns makes no difference.

So I guess we don't even have to expose the `collation` option for parameters, as it's useless anyway. 🤷

jeremy-w:
yes, that matches my understanding. tempdb always using the server's collation is a fun gotcha here.

it looks like the collation for var/char input params should always be the database's current collation. i think it might be practicable to use microsoft's own JDBC driver as a cross-check. some spelunking shows:

- it stores the current databaseCollation: https://github.com/microsoft/mssql-jdbc/blob/8172599fd735a5481b784ed9ef0237b9ab6f7ea2/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L1186-L1191
- and uses it when setting a parameter: https://github.com/microsoft/mssql-jdbc/blob/dev/src/main/java/com/microsoft/sqlserver/jdbc/Parameter.java#L358-L370
- and this gets used as the encoding of the string: https://github.com/microsoft/mssql-jdbc/blob/8172599fd735a5481b784ed9ef0237b9ab6f7ea2/src/main/java/com/microsoft/sqlserver/jdbc/dtv.java#L305-L325 

(ok that last one was clob not plain old var/char but i'd be very surprised if the encoding gymnastics are different in those cases.)

and luckily it's MIT-licensed, so no harm in peeking. 😆 

i think we're already well set up to track that collation, so i will include that in [my work on utf-8](https://github.com/jeremy-w/tedious/tree/jeremy-w/feat/support-utf-8-1143). (i bet my having a DB using a utf-8 collation by default and trying to send varchar params up is where some things were getting mangled…)

Issue #1146, opened 2020-08-21
App Service user-assigned MSI authentication


sharonchoong:
Hello, I'm having issues making App Service MSI authentication work. I went through [this tutorial](https://docs.microsoft.com/en-us/azure/app-service/app-service-web-tutorial-connect-msi) to set up a user-assigned managed identity.  But I still can't make it work. 

The authentication part of my tedious config is 
```
type: 'azure-active-directory-msi-app-service',
options: {
            clientId: client_ID_of_the_user_assigned_MSI
        } 
```
The error I'm getting at runtime is:
```
ConnectionError: Security token could not be authenticated or authorized.
     at ConnectionError (/home/site/wwwroot/node_modules/tedious/lib/errors.js:13:12)
     at /home/site/wwwroot/node_modules/tedious/lib/connection.js:2843:61
     at getTokenFromCredentials (/home/site/wwwroot/node_modules/tedious/lib/connection.js:2809:24)
     at /home/site/wwwroot/node_modules/@azure/ms-rest-nodeauth/dist/lib/login.js:645:24
     at /home/site/wwwroot/node_modules/@azure/ms-rest-js/dist/msRest.node.js:363:13
     at processTicksAndRejections (internal/process/task_queues.js:97:5) {
   code: 'EFEDAUTH'
 }
```

I saw that @azure/ms-rest-nodeauth is being deprecated for @azure/identity. Could that be why it's not working?

IanChokS:
Hi @sharonchoong I'll take a look!

IanChokS:
Actually it looks like there's a bug with with @azure/ms-rest-nodeauth https://github.com/Azure/ms-rest-nodeauth/issues/102
I'll check if using @azure/identity will be able to authenticate using MSI

sharonchoong:
@IanChokS, appreciate it!  Thanks for looking into this, at least I know where the problem likely is. I can use `azure-active-directory-password` in the meantime, but it doesn't seem to be the recommended way to authenticate based on the Microsoft tutorials.  

Regarding @azure/identity, I think this [pull request](https://github.com/tediousjs/tedious/pull/939) is related.

IanChokS:
No problem, I've confirmed that User-Assigned MSI authentication works with the @azure/identity library so I'll try to get that fixed merged ASAP

ramya-rao-a:
Hello,

I am from the team that owns the `@azure/ms-rest-nodeauth` package. 

[...]

PR #1155, opened 2020-09-21
refactor: update @azure/ms-rest-nodeauth

IanChokS:
Fix for #1146 

Refactors `.loginWithAppServiceMSI()` function config to include `clientId` for user-assigned MSI Auth

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3548,6 +3548,8 @@ Connection.prototype.STATE = {
               }, getTokenFromCredentials);
             } else if (authentication.type === 'azure-active-directory-msi-app-service') {
               loginWithAppServiceMSI({
+                msiEndpoint: authentication.options.msiEndpoint,
+                msiSecret: authentication.options.msiSecret,
                 resource: fedAuthInfoToken.spn,
                 clientId: authentication.options.clientId
               }, getTokenFromCredentials);
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3530,7 +3530,7 @@ Connection.prototype.STATE = {
               if (err) {
                 return callback(err);
               }
-                credentials!.getToken().then((tokenResponse: { accessToken: string | undefined; }) => {
+                credentials!.getToken().then((tokenResponse: { accessToken: string | undefined }) => {
                   callback(null, tokenResponse.accessToken);
                 }, callback);
             };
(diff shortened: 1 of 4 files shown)


src/connection.ts
```diff
@@ -3549,9 +3548,8 @@ Connection.prototype.STATE = {
               }, getTokenFromCredentials);
             } else if (authentication.type === 'azure-active-directory-msi-app-service') {
               loginWithAppServiceMSI({
-                msiEndpoint: authentication.options.msiEndpoint,
-                msiSecret: authentication.options.msiSecret,
-                resource: fedAuthInfoToken.spn
+                resource: fedAuthInfoToken.spn,
+                clientId: authentication.options.clientId
```
arthurschreiber, line 3552: Is there a specific reason for dropping the `msiEndpoint` and `msiSecret` options here? 🤔

IanChokS: It doesn't appear to be needed according to the [docs](https://github.com/Azure/ms-rest-nodeauth) 
"`The code below works for both system managed and user-assigned managed identities. You can leave the options empty if you want to use system managed identity. If you want to use the user-assigned managed identity, you must at least provide the clientId in the options.`"

arthurschreiber: That is correct, and that is what made user identities not work correctly here. But the `loginWithAppServiceMSI` still has support for the `msiEndpoint` and `msiSecret`, and I think we want to keep supporting those two options, in addition to `clientId`. 🙇‍♂️

IanChokS: Ok 👍

sharonchoong (approved):
Tested that the user-assigned MSI authentication worked with this PR.  Haven't tested all Azure authentication types but password authentication also had no issues.

MichaelSun90 (approved):

72636c:
Hiya @IanChokS, should this be a `fix` given it changes `dependencies`?

A new release containing this PR would be great as it drops a transitive `axios` dependency that is flagged by vulnerability scanners: https://snyk.io/vuln/SNYK-JS-AXIOS-1038255

arthurschreiber:
This upgrade is causing a `UnhandledPromiseRejectionWarning: ConnectionError: Security token could not be authenticated or authorized.` error when trying to connect via Azure AD password authentication. I'm not sure what's going on exactly, but I'll try to figure out what's going on. 🤔

PR #1156, opened 2020-09-23
feat: Lite - significantly reduce the size of tedious

kristoffer-zliide:
We are building a modern serverless architecture on a Microsoft cloud platform but struggling a bit with cold starts. We have experienced and read [about](https://azure.microsoft.com/en-us/blog/understanding-serverless-cold-start/#:~:text=Write%20lightweight%20code) [others](https://mikhail.io/serverless/coldstarts/azure/#does-package-size-matter) who have experienced a correlation between code size and startup times, when we realized that tedious has increased in size quite significantly since around version 4. Specifically, we see an increase in minified size of an app that does (almost) nothing except from issuing a SQL query from around 60kb to 1,900kb.

This pull request introduces a "tedious lite" alternative entry point that includes a lot less code that users who do not need "extra" things like exotic code pages and federated authentication can use to reduce their code size. This is done as a non-breaking change. Switching from normal tedious to tedious lite takes the minimal app size from around 1,900kb to around 530kb.

Future improvements could include handling date/time without joda, which would take the code size down even further to around 340kb.

Feedback welcome.


arthurschreiber:
This is a great idea! Not loading dependencies that are not always required will help application startup times a lot, and is definitely something worth improving.

I do think we'll want to take a slightly different approach, namely making use of dynamic `import` calls to load some of the modules dynamically.

This should be a straightforward change. @IanChokS is this something you want to work on?

kristoffer-zliide:
Thanks, @arthurschreiber! I just re-read what I wrote, I'm sure I meant that including tedious takes the size from 60kb to 1,900kb, not from 600kb to 1,900kb.

Anyways, we actually employ bundling (currently using browserify) to remove a lot of cruft in node_modules and do an overall minification. The 1,900kb number is also the minified size. Doing dynamic imports tend to trip that up. I think that's why I went with the alternate entry point.

arthurschreiber:
@kristoffer-zliide You're right - dynamic imports might not actually solve much for the use case you have. I'll think about this a bit more.

kristoffer-zliide:
@arthurschreiber: did you think about it a bit more?

trgwii:
This is an excellent effort. In general I think mssql + tedious have become too bloated, and introducing alternatives for lighter modes or splitting out functionality into plugins might be a lot nicer in the future.

Here are some stats from my local tests:
Install of mssql in an empty project, no other dependencies: node_modules becomes 36.5MB
Install of tedious in an empty project, no other dependencies: node_modules becomes 36.0MB

Here's a breakdown of the worst offenders:
image

kristoffer-zliide:
Yeah, I also [experimented](https://github.com/zliide/tedious/commit/c7a4c5501b55f77789f9ee528dd7ccc1b7d66cb0) with removing joda ☺️

trgwii:
I find myself continuously rechecking this issue for updates. In one of my applications at work, the mssql dependency makes up 71% of the size of the entire app including all dependencies.

MichaelSun90:
> I was thinking a bit more about this. One thing is the shear size of the (minified) code, another is really the number of dependencies, which in increases the need for frequently updating packages and increases the number of issues reported by npm audit (which used to be a lot, but maybe not as much recently).
> 
> What if we took the `connection-lite.ts` from this PR and all the code it depended on, and all the packages it depends on and turned into a new package (perhaps called 'tedious-core') and published that. That package would change it's dependencies a lot less, would not need to be touched or updated when Azure-specific changes are made, and it wouldn't be impacted be audit issues in Azure packages and their dependencies, and, of course, it would be much smaller, both as deployed minified code, but also in dev's node_modules. The 'tedious' package could then depend on and extend 'tedious-core' in the way that connection.ts uses connection-lite.ts in this PR.
> 
> @arthurschreiber, @IanChokS: what do you think?

arthurschreiber:
@kristoffer-zliide That sounds good 👍 But it will require some re-architecting of library repo - I'd like us to keep all the code for the different packages in one repo. I guess npm workspaces could help with that? But I've never used that.

MichaelSun90:
@kristoffer-zliide , Sorry that that I closed this earlier by mistake, and Thank you for putting the effort and thoughts on this idea.

PR #1161, opened 2020-10-05
Always Encrypted for latest tedious version

IanChokS:
This rebases #1020 PR on top of tedious v9.2.1


IanChokS:
@fredwes this branch is also a PR for your repo as well. If you decide to update your master branch on your repo, feel free to close this PR!

IanChokS:
Hi @arthurschreiber, when you have the time, can you please review this PR? 
Thanks!

src/token/colmetadata-token-parser.ts
```diff
@@ -60,57 +69,256 @@ function readColumnName(parser: Parser, options: InternalConnectionOptions, inde
...
+          // console.log('>> streambuffer data loadedd')
+          parser.suspended = false;
+          whileLoopPaused = false;
+          readCEKTable(parser, parser.options, (cekList?: CEKEntry[]) => {
+            readCEKTableCb(columnCount, parser, resolved, resolve, cekList);
+          });
+        })
+      }
+    }
+  })
```
IanChokS, line 321: @arthurschreiber I added these changes while merging this branch with the latest master. Let me know what you think. These changes are to accommodate when there is not enough column metadata buffer and needs to wait for more ( Our original implementation didn't' take this scenario into account). The tests `always-encrypted/collation-support-test` and `always-encrypted/cek-cache-test` trigger this scenario).

src/token/stream-parser.ts
```diff
@@ -113,8 +175,17 @@ class Parser {
...
+          // console.log('>> entering colmetadataParser')
+          const token = await colMetadataParser(parser)
+          // console.log('>> done colMetadataParser, starting Promise.all(decryptSymmetricKey)..')
           parser.colMetadata = token.columns;
+          // console.log('colMetdata list = ', parser.colMetadata)
+          await Promise.all(parser.colMetadata.map(async (metadata) => {
+            if (metadata.cryptoMetadata) {
+              await decryptSymmetricKey(metadata.cryptoMetadata, options);
+            }
+          }));
```
IanChokS, line 187: @arthurschreiber I also added these lines here, as this is where the decryption of the keys need to occur. I'm not sure if the other places for this decrypting process is really necessary though 🤔

src/data-types/binary.ts
```diff
@@ -60,6 +60,17 @@ const Binary: { maximumLength: number } & DataType = {
...
 
+  toBuffer: function(parameter) {
+    const value = parameter.value as string | Buffer;
+
+    if (value != null) {
+      return Buffer.isBuffer(value) ? value : Buffer.from(value);
+    } else {
+      // PLP NULL
+      return Buffer.from([ 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF ]);
+    }
```
arthurschreiber, line 71: I noticed that `toBuffer` always has special handling for `null` values, like here, where the PLP NULL buffer value is returned.

I also noticed that `toBuffer` is never called when the parameter value is null, because in those cases the `varbinary` column that actually holds the encrypted information will simply be set to null instead. 🤔

I think that needs to be verified, and if I'm correct, we can remove all the `null` handling logic from `toBuffer`.

Potentially, we can remove `toBuffer` as a whole and instead re-use `generateParameterData`.

src/rpcrequest-payload.ts
```diff
@@ -12,13 +15,14 @@ import { Collation } from './collation';
...
-  DEFAULT_VALUE: 0x02
+  DEFAULT_VALUE: 0x02,
+  ENCRYPTED_VALUE: 0x08,
 };
 
 /*
   s2.2.6.5
  */
-class RpcRequestPayload implements Iterable<Buffer> {
+class RpcRequestPayload implements AsyncIterable<Buffer> {
```
arthurschreiber, line 25: @MichaelSun90 This change here (switching `RPCRequestPayload` from a `Iterable` to a `AsyncIterable`) should be extracted into a separate PR.

The change will allow us to e.g. have `AsyncIterable` objects as values for parameters to `nvarchar(max)`, `varchar(max)`, `varbinary(max)`, and TVPs rows.

src/rpcrequest-payload.ts
```diff
@@ -106,9 +126,76 @@ class RpcRequestPayload implements Iterable<Buffer> {
...
+
+        for (let i = 0, len = parameters.length; i < len; i++) {
+          const type = parameters[i].type;
+          const paramValue = parameters[i].value;
+          if (parameters[i].cryptoMetadata && (paramValue !== undefined && paramValue !== null)) {
+            if (!type.toBuffer) {
+              throw new Error(`Column encryption error. Cannot convert type ${type.name} to buffer.`);
+            }
+            const plainTextBuffer = type.toBuffer(parameters[i], this.options);
+            if (plainTextBuffer) {
```
arthurschreiber, line 164: `plainTextBuffer` can never be `null` here. It will always be a `Buffer` instance.

src/rpcrequest-payload.ts
```diff
@@ -106,9 +126,76 @@ class RpcRequestPayload implements Iterable<Buffer> {
       param.collation = this.collation;
     }
 
-    yield type.generateTypeInfo(param, this.options);
-    yield type.generateParameterLength(param, this.options);
-    yield * type.generateParameterData(param, this.options);
+    const typeINfo = type.generateTypeInfo(param, this.options);
+    yield typeINfo;
```
arthurschreiber, line 130:
```suggestion
    yield type.generateTypeInfo(param, this.options);
```

PR #1164, opened 2020-10-08
Fix `BulkLoad.setTimeout` to actually cancel the request

### LINKED ISSUE #1158: bulkLoad.setTimeout(timeout) doesn't do anything
Just noticed that the `bulkLoad.setTimeout(timeout)` function set's a timeout property, but it doesn't appear to be used at all. If I set the value to 1ms, I would expect almost any query to timeout immediately, but they do not. Looking at the tedious Connection object, `execBulkLoad` function looks like it constructs a new `Request` but doesn't seem to do anything with the timeout value.

tedious version 9.2.1

IanChokS:
Fixes #1158.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3161,6 +3161,8 @@ class Connection extends EventEmitter {
         this.createCancelTimer();
       });
 
+      this.createRequestTimer();
+
       if (request instanceof BulkLoad) {
         message = request.getMessageStream();
 
@@ -3175,8 +3177,6 @@ class Connection extends EventEmitter {
         this.messageIo.outgoingMessageStream.write(message);
         this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
       } else {
-        this.createRequestTimer();
-
         message = new Message({ type: packetType, resetConnection: this.resetConnectionOnNextRequest });
         this.messageIo.outgoingMessageStream.write(message);
         this.transitionTo(this.STATE.SENT_CLIENT_REQUEST);
--- a/test/integration/bulk-load-test.js
+++ b/test/integration/bulk-load-test.js
@@ -28,7 +28,6 @@ function getConfig() {
 }
 
 describe('Bulk Load Tests', function() {
-  this.timeout(60000);
   let connection;
 
   beforeEach(function(done) {
@@ -351,9 +350,7 @@ describe('Bulk Load Tests', function() {
   });
 
   it('should test stream bulk load', function(done) {
-    this.timeout(200000);
-
-    const totalRows = 500000;
+    const totalRows = 20;
     const tableName = '#streamingBulkLoadTest';
 
     connection.on('error', done);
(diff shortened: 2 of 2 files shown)


IanChokS:
@arthurschreiber Can you review this PR?

test/integration/bulk-load-test.js
```diff
@@ -521,6 +521,72 @@ describe('Bulk Load Tests', function() {
...
+  it('should throw `RequestError: Canceled` after 10ms', function(done) {
+
+    const bulkLoad = connection.newBulkLoad('#tmpTestTable5', { keepNulls: true }, function(err, rowCount) {
+      if (err) {
+        assert.strictEqual(err.name, 'RequestError');
+        assert.strictEqual(err.message, 'Canceled.');
+        return done();
+      }
+
+      done(new Error('Test did not throw error as expected'));
```
arthurschreiber, line 533: I think you can simplify these assertions here like this:

```suggestion
      assert.instanceOf(err, Error);
      assert.strictEqual(err.name, 'RequestError');
      assert.strictEqual(err.message, 'Canceled.');

      return done();
```

Can you also fix this similarly in the other tests you added? 🙇‍♂️

src/connection.ts
```diff
@@ -3619,6 +3629,10 @@ Connection.prototype.STATE = {
     events: {
       socketError: function() {
         this.transitionTo(this.STATE.FINAL);
+      },
+      data: function() {                        // Discards incoming data. Can occur if bulkLoad is canceled and data event is still emitted afterwards.
+      },
+      message: function() {                     // Like above, catch unused 'message' event. Can occur if bulkLoad is canceled and message event is still emitted afterwards
```
arthurschreiber, line 3635: I don't think this is correct? 🤔

We should never have to handle messages or data in the `LOGGED_IN`. If we do receive a message or data here, that means we've switched state too early.

I don't quite understand what's happening here that would require this. There's two ways a BulkLoad is transferred to the server:

### Streaming mode

Cancelling the `BulkLoad` while data is streamed to the server will end the message with the ignore bit set. We should remain in the `SENT_CLIENT_REQUEST` until we get a message back from the server that will acknowledge the bulk load was cancelled.

### Non-Streaming mode

The `BulkLoad` will be sent off in full (I think?) and a second `attention` message will be send off right after. The connection will be switched into `SENT_ATTENTION` state and we should wait for a message that acknowledges the bulk load was cancelled.

---

@IanChokS Do you remember what required this to be added? 🤔

arthurschreiber: Answering my own question, this change here https://github.com/tediousjs/tedious/pull/1164/files#diff-a6be0583428e46844273df76939f02077073da3075716fc57d291a5f2463eaf5R3177-R3181 looks to be the cause 😅 

I don't think we should be switching to the `LOGGED_IN` state at that point. Do you remember why this was added?

IanChokS: Ah I just thought that was the way to cancel a request either in flight or before the data was sent out. I wasn't aware of the ignore bit and forgot about the attention message of TDS 😅. Thanks for the pointer, I'll do some updates to make it more proper.

IanChokS:
@arthurschreiber I think this is ready now 🙏

arthurschreiber (approved):

PR #1177, opened 2020-11-21
feat: added error state details into connection-error

gautamsamal:
I am facing an [issue](https://github.com/tediousjs/node-mssql/issues/1120) where I am wondering if it's possible to distinguish the root cause of an `ELOGIN` error during the connection event. Although I know that in most cases SQL server doesn't send a state other than 1 [(ref)](https://docs.microsoft.com/en-us/sql/relational-databases/errors-events/mssqlserver-18456-database-engine-error?view=sql-server-ver15#additional-error-information) but I hope that's configurable.

Assuming the same, does it makes sense to pass those information along with `ConnectionError`? The PR simply adds three properties (number, state, class) to the Error instance.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1982,6 +1982,9 @@ class Connection extends EventEmitter {
         }
       } else {
         const error = ConnectionError(token.message, 'ELOGIN');
+        error.number = token.number;
+        error.state = token.state;
+        error.class = token.class;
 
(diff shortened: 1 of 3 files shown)


IanChokS:
Thanks for raising this PR! We'll see what we can do about adding more error information regarding ELOGIN

Issue #1179, opened 2020-11-26
Support for Table-Valued Parameters in execSql


mastodon0:
Calling execSql with a TVP results in the error 
> Could not find stored procedure 'sp_executesql'

Example code:

    var table = {
        schema: 'dbo',
        name: 'MyTVP',
        columns: [
            { name: 'Name', type: TYPES.VarChar, length: 50 },
            { name: 'Rating', type: TYPES.Int },
        ],
        rows: [['Foo', 1]]
    };

    var request = new Request('select * from @TVP', (err) => {
            if (err) {
                console.log(err);
            }

            connection.close();
        }
    );

    request.addParameter('TVP', TYPES.TVP, table);

    connection.execSql(request);

IanChokS:
Thanks for raising this issue, I'll take a closer look at it soon!

lostfields:
How is this going? Every time I want to use TVP I have to implement procedures where I really want to use ordinary SQL statements (TSQL).

MichaelSun90:
Hi @lostfields , this PR #1327  seems has the changes for this issue. Can you give it a try, see if it fixes your issue? If yes, we can work with @arthurschreiber , see if we can merge and release this fix.

MichaelSun90:
@lostfields , glad to know that works. We will try to get the fix out as soon as possible.

arthurschreiber:
Fixed via https://github.com/tediousjs/tedious/pull/1327.

PR #1180, opened 2020-11-26
fix: tvp in execSql

mastodon0:
Fix for #1179
We can use TVPs by using the ids for special procedures instead of names


arthurschreiber:
Fixed via https://github.com/tediousjs/tedious/pull/1327.

PR #1181, opened 2020-11-27
chore: add typing to cryptoCredentialDetails

elhaard:
The typing is defined for the internal interface, but not for the external interface.

Since newer node versions contains changes to the default TLS settings (e.g. minimum TLS version), more users would need to set the `cryptoCredentialDetails`.

Adding the typing would make that easier.


arthurschreiber:
Cool - this is a pretty handy change! We don't expose typing information yet (I think), so this won't be immediately useful - but it will be once we do so! 🙇‍♂️

arthurschreiber (approved):

PR #1193, opened 2020-12-22
fix: improve Azure AD authentication performance

arthurschreiber:
We use the `@azure/ms-rest-nodeauth` package to perform all Azure ActiveDirectory related authentication steps. That library builds on top of `node-adal` and provides a higher level API and authentication methods that `node-adal` does not provide (like MSI VM and Service authentication).

Because requesting an authentication token is pretty slow (it can take 500ms or much more), `node-adal` exposes a in-memory token caching mechanism. Each request can specify a token cache, or a global token cache will be used instead. Unfortunately, the high level API that `@azure/ms-rest-nodeauth` exposes and that we use, provides a new token cache for each authentication request. In theory, it allows passing in a custom token cache, but that logic is not working correctly (see https://github.com/Azure/ms-rest-nodeauth/pull/107 for a fix).

So, for the time being, this switches us over to a lower level API exposed by `@azure/ms-rest-nodeauth`, where specifying a custom token cache works correctly. This should greatly improve the performance of the `azure-active-directory-password` authentication method by only requesting a token if no token was cached before.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -224,6 +224,29 @@ jobs:
     - name: run integration tests
       run: npx nyc --reporter=lcov npm run test-integration && npx codecov
 
+    - name: Set up CI configuration (AD Service Principal Authentication)
+      run: |
+        echo '{
+          "config": {
+            "server": "${{ secrets.AZURE_SERVER }}",
+            "authentication": {
+              "type": "azure-active-directory-service-principal-secret",
+              "options": {
+                "clientId": "${{ secrets.AZURE_AD_SP_CLIENT_ID }}",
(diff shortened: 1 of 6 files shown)


arthurschreiber:
So, this improves the build time of the Azure CI job using AD authentication from 5+ minutes, down to ~1:30 minutes, and it should also fix the flakiness of that build. It previously timed out quite often, requiring full re-builds of all jobs. That will be a thing of the past soon. 🎉 

I'll see if we can also use hook up the cache for some of the other Azure related authentication methods.

arthurschreiber:
I also opened a pull request against the  `@azure/ms-rest-nodeauth` package, which should allow us to revert back to using the high-level helper functions exposed by the package: https://github.com/Azure/ms-rest-nodeauth/pull/107

Issue #1197, opened 2020-12-31
Add ORDER as a BulkLoad Option


dl-eric:
It would be really nice to have the BulkLoad option "ORDER" when doing bulk inserts. I'd imagine it's an easy couple-line change 😄 

ORDER is described in the Microsoft Docs [here](https://docs.microsoft.com/en-us/sql/t-sql/statements/bulk-insert-transact-sql?view=sql-server-ver15) and [here](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms177468(v=sql.105)).

I'm looking to use Tedious to run our bulk inserts, and a current Bulk Insert optimization we need to do is specifying ORDER as a way to tell SQL to not sort the data as they're already sorted.

IanChokS:
Sorry for the late reply, but thanks for the suggested feature! When we have a bit more free time, we'll take a look into this.

PR #1198, opened 2020-12-31
Add ORDER option to BulkLoad

dl-eric:
First time writing TS, let me know if there are any style issues!

Tests and linter pass locally

image

- Addresses my request #1197 
- Adds basic test to verify the SQL is constructed properly based on bulkLoadOption


dl-eric:
Gonna be honest guys, am giving up after five minutes of trying to appease the commit linter. What should my messages look like?

arthurschreiber:
@dl-eric https://www.conventionalcommits.org/en/v1.0.0/ is the commit message style that we _prefer_.

I think the linter is complaining because your message starts with an upper case letter. That's annoying. I'll see if we can relax that check - it seems to be overly pedantic.

If you don't want to fight the linter, that's fine too. We can fix up the commit history manually when we merge the changes. Sorry for the trouble! 🙇‍♂️

arthurschreiber:
Okay, now that the commit message stuff is out of the way, I had some time to look at the actual changes! 😅 

First up, thank you so much for this contribution! ❤️ It's a simple change but it's a nice improvement if you're bulk loading a lot of data that is pre-ordered. 👍 

I'll leave a few comments in the code for small nits, but this is looking solid otherwise. 🚀

src/bulk-load.ts
```diff
@@ -48,6 +48,7 @@ interface InternalOptions {
   fireTriggers: boolean;
   keepNulls: boolean;
   lockTable: boolean;
+  order: {columnName: string, direction: 'ASC' | 'DESC'}[];
```
arthurschreiber, line 51: I'm wondering if it would make sense to change this to an object, instead of an array of objects?

```suggestion
  order: { [columnName: string]:  'ASC' | 'DESC' };
```

It never makes sense to specify a column name twice in the order, and this would remove a bunch of intermediate objects. When iterating over object properties, they are ordered in the order that they are defined - this is guaranteed by the EcmaScript spec nowadays.

Using this would look something like this:

```js
connection.newBulkLoad('#tmpTestTable', { order: { 'nnn': 'ASC', 'sss': 'DESC' } }, ...
```

What do you think? 🙇‍♂️

src/bulk-load.ts
```diff
@@ -508,6 +515,21 @@ class BulkLoad extends EventEmitter {
       addOptions.push('TABLOCK');
     }
 
+    if (this.bulkOptions.order.length) {
+      const orderColumns = [];
+
+      for (const column of this.bulkOptions.order) {
+        // Ensure column exists as a specified column, otherwise we ignore.
+        if (column.columnName && this.columnsByName[column.columnName] !== null) {
```
arthurschreiber, line 523: I think if `column.columnName` is anything but a string, we should maybe throw an exception? 🤔 

Also, I don't think the strict null check here is required/correct? If the column is not set, `this.columnsByName[column.columnName]` should return `undefined` and not `null`.

src/bulk-load.ts
```diff
@@ -508,6 +515,21 @@ class BulkLoad extends EventEmitter {
...
+          orderColumns.push(`${column.columnName} ${column.direction}`);
```
arthurschreiber, line 524: Do we need to be more careful with the column names here, in regards to SQL injection problems?

src/bulk-load.ts
```diff
@@ -508,6 +515,21 @@ class BulkLoad extends EventEmitter {
...
+        }
+      }
+
+      if (orderColumns.length) {
+        addOptions.push(`ORDER (${orderColumns.join(',')})`);
```
arthurschreiber, line 529: Can we add a whitespace here? 😅 

```suggestion
        addOptions.push(`ORDER (${orderColumns.join(', ')})`);
```

src/bulk-load.ts
dl-eric: Your approach is much cleaner. I'm learning new things about this language every day. Thank you very much!

src/bulk-load.ts
dl-eric: I'd imagine there exists a risk/scenario where a malicious columnName can do unintended actions, so I chose to (almost on a whim) ignore columnNames that didn't exist. But looking at the situation from Tedious's perspective, it might not be the package's responsibility? Not sure - I'll wait for your guidance here 🙏

Also if we do chose to ignore 'invalid' columnNames, it might cause some developer frustration if we ignore their commands without throwing some warning message that they might've mistyped something. This might be an area where we can save some human-hours. Thoughts?

src/bulk-load.ts
dl-eric: I now check `order` to be an 'object' and `columnName` to be a 'string', and throw TypeError if they aren't. How does it look?

src/bulk-load.ts
dl-eric: @arthurschreiber Hey, friendly ping for this PR 😄 any action items for me?

src/bulk-load.ts
```diff
@@ -508,6 +519,22 @@ class BulkLoad extends EventEmitter {
...
+        }
+        // Ensure column exists as a specified column, otherwise we ignore.
+        orderColumns.push(`${column} ${this.bulkOptions.order[column]}`);
+      }
+
+      if (orderColumns.length) {
+        addOptions.push(`ORDER (${orderColumns.join(', ')})`);
+      }
+    }
+
```
arthurschreiber, line 537: Can you move the validation logic into the constructor, and just leave the logic that builds the order fragment in here?

Throwing an error in this method is a bit unergonomic, because the error is thrown far away from where the actual problem lies. 😅

arthurschreiber: Also, if you're iterating over the object key/values, I think it's better to use something like:

```js
for (const [column, order] of Object.entries(this.bulkOptions.order)) {
  ...
}
```

This ensures that prototype properties are skipped correctly.

dl-eric: Makes sense!

src/bulk-load.ts
```diff
@@ -335,6 +335,16 @@ class BulkLoad extends EventEmitter {
       throw new TypeError('The "options.order" property must be of type object.');
     }
 
+    for (const [column, direction] of Object.entries(order)) {
+      if (typeof column !== 'string') {
+        throw new TypeError('The "options.order.columnName" property must be of type string.');
+      }
+
+      if (direction !== 'ASC' && direction !== 'DESC') {
+        throw new TypeError('The direction of ORDER must either be ASC or DESC.');
```
dl-eric, line 344: Is it fine that this is a TypeError?

dl-eric:
@arthurschreiber Any chance we can get this merged this week or next?

PR #1208, opened 2021-01-19
update readme year to 2021

IanChokS:
Updates README to reflect 2021


arthurschreiber (approved):
I guess the proper commit label would be `chore`, as this is not changing any code. Other than that, this is looking good! 👍

PR #1213, opened 2021-01-28
fix: add connection.connect() to broken minimal example

Staff-d:
This PR fixes the `minimal` example, which is currently broken.

Running the example currently yields no output, as a `connect()` call is missing. Adding the missing call executes the query as expected after the connection is ready.


arthurschreiber:
Oh, thank you very much! 🙇‍♂️

PR #1215, opened 2021-01-30
refactor: improve message handling

arthurschreiber:
The TDS protocol is message based - the client sends a request message to the server, and the server sends a response message back. Messages are broken up into smaller packets, that are sent over the connection, but there is no interleaving of packets from different messages.

But the way that message handling in `tedious` is implemented, is not making this obvious. Different states have separate `data` (for message contents) and `message` (for the end of a message) handlers. This generally works fine. If a message is received, data is processed via `data` events and then once the message is fully handled, the `message` event should handle cleanup tasks, like switching to a different state.

Unfortunately, that model easily gets messed up if we switch the connection state outside of `message` handlers, and that has been leading to all sorts of confusion.

To simplify this, I'm merging the `message` and `data` event handlers into a single, async `message` event, that is responsible for consuming the message fully.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -390,10 +390,8 @@ interface State {
     socketError?(this: Connection, err: Error): void;
     connectTimeout?(this: Connection): void;
     socketConnect?(this: Connection): void;
-    data?(this: Connection, data: Buffer): void;
-    message?(this: Connection): void;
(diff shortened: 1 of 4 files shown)


arthurschreiber:
Closing this for now in favor of https://github.com/tediousjs/tedious/pull/1221.

Issue #1201, opened 2021-01-07
Possible EventEmitter memory leak


andrewsensus:
I think I'm seeing a memory leak in tedious, although I don't have a working example that is independent of my much larger application. My use pattern is to create a request object, prepare a statement, then run many queries using that prepared statement. I find that if I occasionally reset the connection, throw away the request object, and create a new one, my leak goes away.

I noticed this warning:
```
2021-01-07 05:22:41.060: 7011 (node:7011) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 cancel listeners added to [Request]. Use emitter.setMaxListeners() to increase limit
```

A core dump examined with llnode shows a lot of ReadableState, WritableState, and a few other types:
```
       6603     393232 Promise
      14066    1237808 (ArrayBufferView)
     275604    8819368 (Array)
     858961   48101784 RpcRequestPayload
     858976  109948928 PassThrough
     859216   41242368 CorkedRequest
     859344  185618304 ReadableState
     859348  206245344 WritableState
     859470   41258888 BufferList
    1374337    7734328 (String)
    3408785  189469752 Object
 ---------- ----------
   10246521  840970736
```

`connection.js` calls `makeRequest`. `makeRequest `calls `request.once('cancel', ...)`. I suspect the listener is never getting cleaned up and that the closure captures `message` which probably references a lot of things. I've confirmed with node --trace-warnings that `makeRequest` is the location the event emitter is complaining about.

In v5.0.0 a feature was added, "allow `BulkRequest`s to be canceled" (f78d745cbd79d87fa0de159bb19cf8ccb7c4a1e3). I see the introduction of the cancel listener, but I don't see where it is cleaned up on the non-bulk and non-error path.

IanChokS:
Hi @andrewsensus , what version of tedious are you using? I think we've addressed this problem before but I'm not 100% sure.

andrewsensus:
Hi @IanChokS I just realized I was running some of my tests against 9.2.1 and some against 9.2.3. Prior to that I was on 2-something.

arthurschreiber:
@andrewsensus Prepared statements reuse the same `Request`s, if I remember correctly? 🤔 If that's true, you're right, the `cancel` handlers are never cleared out and will accumulate over time.

Thank you for reporting this!

I can't estimate how involved a fix would be right now. I took a quick look, and I think we'll have to unsubscribe from the 'cancel' event in at least these two cases:
[...]

PR #1221, opened 2021-02-16
refactor: improve message handling

arthurschreiber:
The TDS protocol is message based - the client sends a request message to the server, and the server sends a response message back. Messages are broken up into smaller packets, that are sent over the connection, but there is no interleaving of packets from different messages.

But the way that message handling in tedious is implemented is not making this obvious. Different states have separate `data` (for message contents) and `message` (for the end of a message) handlers. This generally works fine. If a message is received, data is processed via `data` events and then once the message is fully handled, the `message` event should handle cleanup tasks, like switching to a different state.

Unfortunately, that model easily gets messed up if we switch the connection before a message has been fully consumed, and that has been leading to all sorts of confusion.

To simplify this, I'm merging the message and data event handlers into a single `message` event that is responsible for consuming the message fully. I've also changed all the uses of the token stream parser to correctly account for backpressure and to wait until all data has been processed fully before performing any state switching.

---

This is similar to https://github.com/tediousjs/tedious/pull/1215, but does not make use of `async`/`await` - instead it uses regular events to reach the same goal.

---

I believe this fixes the event emitter leak mentioned in https://github.com/tediousjs/tedious/issues/1201, because we now properly remove `cancel` event listeners before we move to a different state.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -390,14 +390,11 @@ interface State {
     socketError?(this: Connection, err: Error): void;
     connectTimeout?(this: Connection): void;
     socketConnect?(this: Connection): void;
-    data?(this: Connection, data: Buffer): void;
-    message?(this: Connection): void;
+    message?(this: Connection, message: Message): void;
     retry?(this: Connection): void;
-    routingChange?(this: Connection): void;
     reconnect?(this: Connection): void;
     featureExtAck?(this: Connection, token: FeatureExtAckToken): void;
     fedAuthInfo?(this: Connection, token: FedAuthInfoToken): void;
-    endOfMessageMarkerReceived?(this: Connection): void;
     loginFailed?(this: Connection): void;
     attention?(this: Connection): void;
   };
@@ -2029,7 +2026,6 @@ class Connection extends EventEmitter {
 
     tokenStreamParser.on('routingChange', (token) => {
       this.routingData = token.newValue;
-      this.dispatchEvent('routingChange');
     });
 
     tokenStreamParser.on('packetSizeChange', (token) => {
@@ -2107,7 +2103,7 @@ class Connection extends EventEmitter {
           if (this.config.options.rowCollectionOnDone) {
             request.rst!.push(token.columns);
           }
-          if (!(this.state === this.STATE.SENT_ATTENTION && request.paused)) {
+          if (!request.canceled) {
             request.emit('row', token.columns);
           }
         }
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -3645,29 +3721,78 @@ Connection.prototype.STATE = {
...
+        });
+
+        // If the request was canceled after the request was sent, but before
+        // we started receiving a message, we send an attention message, fully
+        // consume the current message, and then switch the next state.
+        if (this.request?.canceled) {
+          this.tokenStreamParser.once('endOfMessage', () => {
+            this.transitionTo(this.STATE.SENT_ATTENTION);
+          });
+
```
IanChokS, line 3757: Oh nevermind, found it! 😄

IanChokS (approved):

IanChokS (commented):
Codecov is complaining a lot of untested lines. Should we add more tests for them?

arthurschreiber:
I extracted the fix for the EventEmitter memory leak into https://github.com/tediousjs/tedious/pull/1222, to make it easier to review that change separately and to ensure a new release is cut once it's merged.

arthurschreiber:
I ran the `benchmarks/query/select-many-rows.js` benchmark with and without these changes, and could not really see a difference really. The number of iterations per second is still roughly in the same ballpark:

### `master`

```
$ node benchmarks/query/select-many-rows.js size=100
query/select-many-rows.js size=100 n=10: 285.0619011918438 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 471.0169005715399 (minor: 3 - 3.7271219999999996ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 621.2009778642647 (minor: 31 - 37.251939ms, major: 0 - 0ms, incremental: 0 - 0ms)
```

### `arthur/improve-message-handling`

```
$ node benchmarks/query/select-many-rows.js size=100
query/select-many-rows.js size=100 n=10: 278.1678463672308 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 452.0936091102541 (minor: 3 - 4.325525ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 628.57227089419 (minor: 31 - 32.194192ms, major: 0 - 0ms, incremental: 0 - 0ms)
```

PR #1222, opened 2021-02-16
Prevent memory leaks via event handlers

arthurschreiber:
We were not correctly cleaning up `cancel` event handlers when a request was not canceled. This was not an issue for single use requests, but when a request was prepared and reused many times, `cancel` event handlers were set up over and over again. This would cause a memory leak, and would actually break request cancellation.

This fixes the memory leak by making sure the `cancel` event handlers are cleaned up.

Fixes: #1201

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2821,7 +2821,14 @@ class Connection extends EventEmitter {
    */
   execBulkLoad(bulkLoad: BulkLoad) {
     bulkLoad.executionStarted = true;
+
+    const onCancel = () => {
+      request.cancel();
+    };
+
     const request = new Request(bulkLoad.getBulkInsertSql(), (error: (Error & { code?: string }) | null | undefined) => {
+      bulkLoad.removeListener('cancel', onCancel);
+
       if (error) {
         if (error.code === 'UNKNOWN') {
           error.message += ' This is likely because the schema of the BulkLoad does not match the schema of the table you are attempting to insert into.';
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -3129,22 +3134,21 @@ class Connection extends EventEmitter {
...
 
-          if (request instanceof Request && request.paused) {
-            // resume the request if it was paused so we can read the remaining tokens
-            request.resume();
-          }
+        if (request instanceof Request && request.paused) {
+          // resume the request if it was paused so we can read the remaining tokens
+          request.resume();
         }
-      });
```
arthurschreiber, line 3147: We no longer need the `message.writable` check here, because a message is guaranteed to be writable until the  `finish` event has been called, and that's also when the event handler will be removed.

src/connection.ts
```diff
@@ -2821,7 +2821,14 @@ class Connection extends EventEmitter {
    */
   execBulkLoad(bulkLoad: BulkLoad) {
     bulkLoad.executionStarted = true;
+
+    const onCancel = () => {
+      request.cancel();
+    };
+
     const request = new Request(bulkLoad.getBulkInsertSql(), (error: (Error & { code?: string }) | null | undefined) => {
+      bulkLoad.removeListener('cancel', onCancel);
```
arthurschreiber, line 2830: This ensures we remove the event handler that forwarded cancellation from a `BulkLoad` object to the internal `Request` object.

PR #1227, opened 2021-02-25
Convert token parser to an async generator function

arthurschreiber:
This pull request converts the token parsing code from a `Transform` stream into a async generator function that consumes a `AsyncIterable<Buffer>` and yields back parsed tokens.

All of the existing token parsing implementation is retained, this mostly just switches out how the parsing is "driven". This gives us the option to call other async code during the parsing (e.g. as is required for always encrypted to get encryption keys).

Right now, the async generator is wrapped in a `Readable.from` call, so we can keep most of the existing code as-is.

This depends on https://github.com/tediousjs/tedious/pull/1226 to be merged first.

---

### Performance

This change has a negative impact on performance, mostly because we introduce some additional overhead by having a async generator and wrapping it back into a `Readable` stream via `Readable.from`. This is an interim solution and will hopefully be refactored away soon, which should improve performance again. In the mean time, I believe the performance hit is acceptable.

#### Before

```
root ➜ ~/workspace/tedious (master) $ node benchmarks/query/select-many-rows.js 
query/select-many-rows.js size=10 n=10: 532.4236697461165 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=10: 284.16636656330303 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=10: 85.51023654585572 (minor: 3 - 3.2305240000000004ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=10: 15.934764679625893 (minor: 24 - 17.215927ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=100: 674.8609042503098 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 463.037634212558 (minor: 3 - 3.7732279999999996ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=100: 147.4007835928836 (minor: 27 - 15.749910000000002ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=100: 20.14078649029258 (minor: 235 - 112.42549800000005ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=1000: 895.6707557867154 (minor: 6 - 14.221501ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 635.4394020000775 (minor: 31 - 30.979024999999993ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=1000: 166.43689997146637 (minor: 267 - 136.60967799999997ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=1000: 20.153075716417455 (minor: 2371 - 1165.8097820000007ms, major: 0 - 0ms, incremental: 0 - 0ms)
```

#### After

```
root ➜ ~/workspace/tedious (arthur/token-parser-async-generator) $ node benchmarks/query/select-many-rows.js 
query/select-many-rows.js size=10 n=10: 502.91585584057805 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=10: 227.16272888405445 (minor: 1 - 2.704519ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=10: 81.94501407610478 (minor: 3 - 3.8326270000000005ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=10: 13.575519515447352 (minor: 32 - 22.000658ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=100: 524.0109713434697 (minor: 1 - 1.42641ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 404.50896107083423 (minor: 5 - 8.497362ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=100: 107.47463633511748 (minor: 34 - 35.75296500000001ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=100: 16.49632461244278 (minor: 320 - 232.55980199999993ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=1000: 854.1123887852357 (minor: 9 - 18.826438ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 539.9230060830604 (minor: 43 - 57.258018ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=1000: 143.33281997154975 (minor: 344 - 222.09660600000004ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=1000: 15.99416653285739 (minor: 3203 - 2346.3768180000034ms, major: 0 - 0ms, incremental: 0 - 0ms)
```

### DIFF
--- a/src/token/stream-parser.ts
+++ b/src/token/stream-parser.ts
@@ -2,7 +2,6 @@ import Debug from '../debug';
 import { InternalConnectionOptions } from '../connection';
 import JSBI from 'jsbi';
 
-import { Transform } from 'readable-stream';
 import { TYPE, Token, ColMetadataToken } from './token';
 
 import colMetadataParser, { ColumnMetadata } from './colmetadata-token-parser';
@@ -38,7 +37,7 @@ const tokenParsers = {
   [TYPE.SSPI]: sspiParser
 };
 
-class Parser extends Transform {
+class Parser {
   debug: Debug;
   colMetadata: ColumnMetadata[];
   options: InternalConnectionOptions;
@@ -48,9 +47,65 @@ class Parser extends Transform {
   suspended: boolean;
   next?: () => void;
 
-  constructor(debug: Debug, options: InternalConnectionOptions) {
-    super({ objectMode: true });
+  static async *parseTokens(iterable: AsyncIterable<Buffer> | Iterable<Buffer>, debug: Debug, options: InternalConnectionOptions, colMetadata: ColumnMetadata[] = []) {
+    const parser = new Parser(debug, options);
+    parser.colMetadata = colMetadata;
 
+    let token: Token | undefined;
+    const onDoneParsing = (t: Token | undefined) => { token = t; };
+
+    for await (const input of iterable) {
+      if (parser.position === parser.buffer.length) {
+        parser.buffer = input;
+      } else {
+        parser.buffer = Buffer.concat([parser.buffer.slice(parser.position), input]);
+      }
+      parser.position = 0;
+
+      if (parser.suspended) {
+        // Unsuspend and continue from where ever we left off.
+        parser.suspended = false;
+        const next = parser.next!;
+
+        next();
+
+        // Check if a new token was parsed after unsuspension.
+        if (!parser.suspended && token) {
+          if (token instanceof ColMetadataToken) {
+            parser.colMetadata = token.columns;
+          }
+
+          yield token;
+        }
+      }
+
+      while (!parser.suspended && parser.position + 1 <= parser.buffer.length) {
+        const type = parser.buffer.readUInt8(parser.position);
+
+        parser.position += 1;
+
+        if (tokenParsers[type]) {
+          tokenParsers[type](parser, parser.options, onDoneParsing);
+
+          // Check if a new token was parsed after unsuspension.
+          if (!parser.suspended && token) {
+            if (token instanceof ColMetadataToken) {
+              parser.colMetadata = token.columns;
+            }
+            yield token;
+          }
+        } else {
+          throw new Error('Unknown type: ' + type);
+        }
+      }
+    }
+
+    if (parser.suspended) {
+      throw new Error('unexpected end of data');
+    }
+  }
+
+  constructor(debug: Debug, options: InternalConnectionOptions) {
     this.debug = debug;
     this.colMetadata = [];
     this.options = options;
@@ -59,55 +114,6 @@ class Parser extends Transform {
     this.position = 0;
     this.suspended = false;
     this.next = undefined;
-  }
-
-  _transform(input: Buffer, _encoding: string, done: (error?: Error | undefined, token?: Token) => void) {
-    if (this.position === this.buffer.length) {
-      this.buffer = input;
-    } else {
-      this.buffer = Buffer.concat([this.buffer.slice(this.position), input]);
-    }
-    this.position = 0;
-
-    if (this.suspended) {
-      // Unsuspend and continue from where ever we left off.
-      this.suspended = false;
-      const next = this.next!;
-
-      next();
-    }
-
-    // If we're no longer suspended, parse new tokens
-    if (!this.suspended) {
-      // Start the parser
-      this.parseTokens();
-    }
-
-    done();
-  }
-
-  parseTokens() {
-    const doneParsing = (token: Token | undefined) => {
-      if (token) {
-        if (token instanceof ColMetadataToken) {
-          this.colMetadata = token.columns;
-        }
-
-        this.push(token);
-      }
-    };
-
-    while (!this.suspended && this.position + 1 <= this.buffer.length) {
-      const type = this.buffer.readUInt8(this.position);
-
-      this.position += 1;
-
-      if (tokenParsers[type]) {
-        tokenParsers[type](this, this.options, doneParsing);
-      } else {
-        this.emit('error', new Error('Unknown type: ' + type));
-      }
-    }
   }
 
   suspend(next: () => void) {
--- a/benchmarks/token-parser/colmetadata-token.js
+++ b/benchmarks/token-parser/colmetadata-token.js
@@ -7,16 +7,19 @@ const bench = createBenchmark(main, {
   tokenCount: [10, 100, 1000, 10000]
 });
 
-function main({ n, tokenCount }) {
-  const parser = new Parser({ token: function() { } }, {}, {});
+async function * repeat(data, n) {
+  for (let i = 0; i < n; i++) {
+    yield data;
+  }
+}
 
+function main({ n, tokenCount }) {
   const data = Buffer.from('810300000000001000380269006400000000000900e7c8000904d00034046e0061006d006500000000000900e7ffff0904d000340b6400650073006300720069007000740069006f006e00'.repeat(tokenCount), 'hex');
+  const parser = new Parser(repeat(data, n), { token: function() { } }, {}, {});
 
   bench.start();
 
-  for (let i = 0; i < n; i++) {
-    parser.write(data);
-  }
-
-  bench.end(n);
+  parser.on('end', () => {
+    bench.end(n);
+  });
 }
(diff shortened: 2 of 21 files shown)


src/token/stream-parser.ts
```diff
@@ -48,68 +47,85 @@ class Parser extends Transform {
   suspended: boolean;
   next?: () => void;
 
-  constructor(debug: Debug, options: InternalConnectionOptions) {
-    super({ objectMode: true });
+  static async *parseTokens(iterable: AsyncIterable<Buffer>, debug: Debug, options: InternalConnectionOptions) {
```
IanChokS, line 50: Would it make more sense to move this as it's own function and export this instead of the Parser class?

IanChokS (changes requested):
Some tests that used to call `parser.write` needs to be updated. I'll try to update those.

src/token/stream-parser.ts
arthurschreiber: Yes, I think that'd make sense. I skipped doing that, mostly because it will require us to update a bunch of `import` statements in multiple places and I wanted to save myself the headache of doing that. 😅 

Let's do that in a followup PR! 👍

arthurschreiber:
> Some tests that used to call `parser.write` needs to be updated. I'll try to update those.

Ah, yes, you're right!

To update these tests, I think you should update the `parseTokens` function to also accept an `Iterable<Buffer>`, like this:

```diff
diff --git a/src/token/stream-parser.ts b/src/token/stream-parser.ts
index 85f27e3..5da37ad 100644
--- a/src/token/stream-parser.ts
+++ b/src/token/stream-parser.ts
@@ -47,25 +47,12 @@ class Parser {
   suspended: boolean;
   next?: () => void;
 
-  static async *parseTokens(iterable: AsyncIterable<Buffer>, debug: Debug, options: InternalConnectionOptions) {
-    const iterator = iterable[Symbol.asyncIterator]();
-
+  static async *parseTokens(iterable: AsyncIterable<Buffer> | Iterable<Buffer>, debug: Debug, options: InternalConnectionOptions) {
     const parser = new Parser(debug, options);
 
     let token: Token | undefined;
 
-    while (true) {
-      const result = await iterator.next();
-      if (result.done) {
-        if (parser.position === parser.buffer.length) {
-          return;
-        }
-
-        throw new Error('unexpected end of data');
-      }
-
-      const input = result.value;
-
+    for await (const input of iterable) {
       if (parser.position === parser.buffer.length) {
         parser.buffer = input;
       } else {
@@ -113,6 +100,10 @@ class Parser {
         }
       }
     }
+
+    if (parser.suspended) {
+      throw new Error('unexpected end of data');
+    }
   }
 
   constructor(debug: Debug, options: InternalConnectionOptions) {
```

This change will allow you to pass in an array of buffers to `parseTokens` in the test code (arrays implement the `Iterable` interface). The test cases will need to be converted to async functions, which will allow you to just pull the parsed tokens out of the async generator. Something like this:

```diff
diff --git a/test/unit/token/colmetadata-token-parser-test.js b/test/unit/token/colmetadata-token-parser-test.js
index 64416a9..b441d22 100644
--- a/test/unit/token/colmetadata-token-parser-test.js
+++ b/test/unit/token/colmetadata-token-parser-test.js
@@ -4,7 +4,7 @@ const TokenStreamParser = require('../../../src/token/stream-parser');
 const assert = require('chai').assert;
 
 describe('Colmetadata Token Parser', () => {
-  it('should int', () => {
+  it('should int', async function() {
     const numberOfColumns = 1;
     const userType = 2;
     const flags = 3;
@@ -20,11 +20,15 @@ describe('Colmetadata Token Parser', () => {
     buffer.writeBVarchar(columnName);
     // console.log(buffer.data)
 
-    const parser = new TokenStreamParser({ token() { } }, {}, {});
-    parser.write(buffer.data);
-    const token = parser.read();
-    // console.log(token)
+    const parser = TokenStreamParser.parseTokens([buffer.data], { token() { } }, {}, {});
+    const tokens = [];
+    for await (const token of parser) {
+      tokens.push(token);
+    }
+
+    assert.lengthOf(tokens, 1);
 
+    const token = tokens[0];
     assert.isOk(!token.error);
     assert.strictEqual(token.columns.length, 1);
     assert.strictEqual(token.columns[0].userType, 2);
```

arthurschreiber:
> To update these tests, I think you should update the `parseTokens` function to also accept an `Iterable<Buffer>`, like this:

@IanChokS I pushed the change I described here - but I didn't update any of the tests as you mentioned you want to look into that.

IanChokS:
Oh I already started reworking the tests with the previous implementation of the parser, which looks like this:

```js
  it('should be of bad type', (done) => {
    const buffer = new WritableTrackingBuffer(50, 'ucs2');

    buffer.writeUInt8(0xe3);
    buffer.writeUInt16LE(0); // Length written later
    buffer.writeUInt8(0xff); // Bad type

    const data = buffer.data;
    data.writeUInt16LE(data.length - 3, 1);

    const asyncIterable = {
      async*[Symbol.asyncIterator]() {
        yield buffer.data;
      }
    };

    const parser = StreamParser.parseTokens(asyncIterable, {}, {});

    (async () => {
      for await (let token of parser) { // (4)
        assert.strictEqual(token, null);
        done();
      }
    })()
  });
```'
on top of that, I've added some slight changes to the old parser, one which includes adding an optional parser object instead of forcefully creating a new parser object in the `*parserTokens` method. This is to fix the `row-parser-tests` because the tests manually set `colMetadata` property in the `Parser` class which previously was only created inside the `*parseTokens` method and therefore was not possible to set the field. With this optional parameter, the tests now look something like this: 
```js
// Parameter change
static async *parseTokens(iterable: AsyncIterable<Buffer>, debug: StreamParser, options: InternalConnectionOptions, parser?: Parser) 
```

```js
// Updated row-token-parser test e.g.,
    const asyncIterable = {
      async*[Symbol.asyncIterator]() {
        yield buffer.data;
      }
    };

    const parserObj = new Parser({}, options);
    parserObj.colMetadata = colMetadata;
    const parser = Parser.parseTokens(asyncIterable, {}, options, parserObj);

    (async () => {
      for await (let token of parser) {
        assert.strictEqual(token.columns.length, 1);
        assert.strictEqual(token.columns[0].value, value);
        assert.strictEqual(token.columns[0].metadata, colMetadata[0]);
        done();
      }
    })()
```
Are these changes acceptable so far?

IanChokS:
Actually with the option `Iterable<Buffer>`, it would be easier to just pass in the Buffer arrays for the tests instead of creating an AsyncIterable object like 
```js
    const asyncIterable = {
      async*[Symbol.asyncIterator]() {
        yield buffer.data;
      }
    };
```
I'll pull your changes and update the tests accordingly

arthurschreiber:
@IanChokS that parameter change to allow injecting a pre-built parser object isn't great, but it's okay for now to get the tests fixed up! 👍 I hope that can be refactored away sooner or later anyway. 😄

arthurschreiber:
> Actually with the option `Iterable<Buffer>`, it would be easier to just pass in the Buffer arrays for the tests instead of creating an AsyncIterable object like

Yes, that's the main idea. 😅

A simpler version of the approach you've taken to build an async iterable object is to define an async generator function and call it like this:

```ts
const asyncIterable = (async function*() => {
  yield buffer.data;
})();
```

But yes, for the tests I think just using an array of buffers is just simpler and easier to read and understand. 👍

arthurschreiber:
I pushed one more change that _slightly_ improves performance by re-using the callback that's passed into the individual token parsing functions. The performance difference isn't super big, but it's an easy improvement.

PR #1232, opened 2021-03-03
refactor: wip: updating token unit tests

IanChokS:
This PR refactors current unit/token tests to call the updated stream-parser function properly. 
- There are currently 3 failing tests (all from row-stream-parser-test.js)
- For `token/stream-parser.ts`, I've added 
  - An overall `try catch` for the whole `*parseTokens` function
  - A `yield null` on line 101 (to keep the behavior consistent with master)
  - An optional premade `Parser` object in the `*parseTokens` function to get row-token-parser tests working properly

### DIFF
--- a/src/token/stream-parser.ts
+++ b/src/token/stream-parser.ts
@@ -1,4 +1,4 @@
-import Debug from '../debug';
+import StreamParser from '../debug';
 import { InternalConnectionOptions } from '../connection';
 import JSBI from 'jsbi';
 
@@ -17,6 +17,7 @@ import returnValueParser from './returnvalue-token-parser';
 import rowParser from './row-token-parser';
 import nbcRowParser from './nbcrow-token-parser';
 import sspiParser from './sspi-token-parser';
+import Debug from '../debug';
 
 const tokenParsers = {
   [TYPE.COLMETADATA]: colMetadataParser,
@@ -38,7 +39,7 @@ const tokenParsers = {
 };
 
 class Parser {
-  debug: Debug;
+  debug: StreamParser;
   colMetadata: ColumnMetadata[];
   options: InternalConnectionOptions;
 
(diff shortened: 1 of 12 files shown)


test/unit/token/env-change-token-parser-test.js
```diff
@@ -61,10 +62,10 @@ describe('Env Change Token Parser', () => {
...
 
-    const parser = new Parser({ token() { } }, {}, {});
-    parser.write(data);
-    const token = parser.read();
+    const parser = StreamParser.parseTokens([data], {}, {});
 
-    assert.strictEqual(token, null);
+    for await (let token of parser) { // (4)
+      assert.strictEqual(token, null);
+    }
```
arthurschreiber, line 69: I don't think this (and the addition of `yield null` in the `Parser.parseTokens` static method) is quite right. 😅 

The `.read()` in the NodeJS `Readable` stream API is a synchronous method that either returns the next element of the stream (which would be a token), or `null` if there is no data available yet.

There's no completely matching equivalent to this when using an async generator, but that's okay - we just need to replicate the meaning of this test correctly.

The old test was trying to make sure that if data that references an invalid envchange token type was parsed, **no token** was returned by the parser - the data written to the parser was basically skipped.

One way to translate this more closely to the async generator version would just be to fail immediately if a token is generated. Something like this:

```
ts
for await (let token of parser) {
  assert.fail('expected no token to be returned');
}
```

src/token/stream-parser.ts
```diff
@@ -1,4 +1,4 @@
-import Debug from '../debug';
+import StreamParser from '../debug';
```
arthurschreiber, line 1: I think this line was added by accident and needs to be removed.

src/token/stream-parser.ts
```diff
@@ -38,7 +39,7 @@ const tokenParsers = {
 };
 
 class Parser {
-  debug: Debug;
+  debug: StreamParser;
```
arthurschreiber, line 42:
```suggestion
  debug: Debug;
```

test/unit/token/token-stream-parser-test.js
```diff
@@ -29,29 +29,31 @@ describe('Token Stream Parser', () => {
...
   it('should split token across buffers', (done) => {
     var buffer = createDbChangeBuffer();
+    
+    const asyncIterable = {
+      async*[Symbol.asyncIterator]() {
+        yield buffer.slice(0, 6);
+        yield buffer.slice(6);
+      }
+    };
+    var parser = new Parser(asyncIterable, debug);
```
arthurschreiber, line 51: Using an array makes this a bit simpler:

```suggestion
    var parser = new Parser([ buffer.slice(0, 6), buffer.slice(6) ], debug);
```

test/unit/token/colmetadata-token-parser-test.js
```diff
@@ -20,20 +21,19 @@ describe('Colmetadata Token Parser', () => {
...
-    assert.strictEqual(token.columns[0].type.name, 'Int');
-    assert.strictEqual(token.columns[0].colName, 'name');
+    for await (let token of parser) { // (4)
+      assert.isOk(!token.error);
+      assert.strictEqual(token.columns.length, 1);
+      assert.strictEqual(token.columns[0].userType, 2);
+      assert.strictEqual(token.columns[0].flags, 3);
+      assert.strictEqual(token.columns[0].type.name, 'Int');
+      assert.strictEqual(token.columns[0].colName, 'name');
+    }
```
arthurschreiber, line 33: What do you think about changing this to explicit iteration, like this:

```suggestion
    let result = await parser.next();
    assert.isFalse(result.done);

    const token = result.value;
    assert.isOk(!token.error);
    assert.strictEqual(token.columns.length, 1);
    assert.strictEqual(token.columns[0].userType, 2);
    assert.strictEqual(token.columns[0].flags, 3);
    assert.strictEqual(token.columns[0].type.name, 'Int');
    assert.strictEqual(token.columns[0].colName, 'name');

    result = await parser.next();
    assert.isTrue(result.done);
```

I think this is a bit closer to the old test code. It also is more robust, e.g. most of the `for await` based tests would pass if we e.g. introduced a bug that accidentally returned each token twice or more times.

arthurschreiber:
I left a few comments, but overall this is looking great - thank you! 🙇‍♂️

PR #1242, opened 2021-03-21
Refactor way of verifying the authentication type

funkydev:
- export authentication interfaces
- create enum with supported authentication types
- simplify conditions for verifying authentication types


funkydev:
@arthurschreiber Could you look up to my refactor changes? I'm on the way to implement a new authentication type, but first, I'd like to clean up the code.

src/connection.ts
```diff
@@ -3362,7 +3379,15 @@ Connection.prototype.STATE = {
       },
       featureExtAck: function(token) {
         const { authentication } = this.config;
-        if (authentication.type === 'azure-active-directory-password' || authentication.type === 'azure-active-directory-access-token' || authentication.type === 'azure-active-directory-msi-vm' || authentication.type === 'azure-active-directory-msi-app-service' || authentication.type === 'azure-active-directory-service-principal-secret') {
+        const aadAuthentications = [
```
arthurschreiber, line 3418: What do you think about extracting this to a top-level constant (but not exporting it).

src/connection.ts
```diff
@@ -196,8 +196,18 @@ const DEFAULT_LANGUAGE = 'us_english';
...
-  type: 'azure-active-directory-msi-app-service';
+export enum SupportedAuthenticationTypes {
+  AzureActiveDirectoryMsiAppService = 'azure-active-directory-msi-app-service',
+  AzureActiveDirectoryMsiVm = 'azure-active-directory-msi-vm',
+  AzureActiveDirectoryAccessToken = 'azure-active-directory-access-token',
+  AzureActiveDirectoryPassword = 'azure-active-directory-password',
+  AzureActiveDirectoryServicePrincipalSecret = 'azure-active-directory-service-principal-secret',
+  Ntlm = 'ntlm',
+  Default = 'default',
+}
```
arthurschreiber: How do you feel about using an object here, instead of an enum? I find TypeScript enums to be weird - event the TypeScript docuementation suggests using objects instead:

```suggestion
export const SupportedAuthenticationTypes = {
  AzureActiveDirectoryMsiAppService: 'azure-active-directory-msi-app-service',
  AzureActiveDirectoryMsiVm: 'azure-active-directory-msi-vm',
  AzureActiveDirectoryAccessToken: 'azure-active-directory-access-token',
  AzureActiveDirectoryPassword: 'azure-active-directory-password',
  AzureActiveDirectoryServicePrincipalSecret: 'azure-active-directory-service-principal-secret',
  Ntlm: 'ntlm',
  Default: 'default',
} as const;
```

I'd also suggest changing the name to `AuthenticationTypes`, the fact that they're supported is sort of implicit. 😅

src/connection.ts
```diff
@@ -3497,7 +3564,14 @@ Connection.prototype.STATE = {
...
               };
 
-              if (authentication.type === 'azure-active-directory-password') {
+              if (authentication.type === SupportedAuthenticationTypes.AzureActiveDirectoryAccessTokenGenerator) {
+                authentication
+                  .options
+                  .tokenGenerator
+                  .getToken()
+                  .then((token: string) => callback(null, token))
+                  .catch((err) => callback(err));
```
arthurschreiber, line 3573: I don't think we need to introduce a new authentication type for this. The `AzureActiveDirectoryAccessToken` can check whether the given `token` is a function, and if it is, do `await options.token.call(null)` inside the `sendLogin7Packet` method.

Does that make sense?

src/connection.ts
funkydev: Sure, but as we exporting them, we have to make sure that it won't be changed, so I'll make it as non-mutable by freezing that, like that:

```typescript
export const AuthenticationTypes  = Object.freeze({
  AzureActiveDirectoryMsiAppService = 'azure-active-directory-msi-app-service',
  AzureActiveDirectoryMsiVm = 'azure-active-directory-msi-vm',
  AzureActiveDirectoryAccessToken = 'azure-active-directory-access-token',
  AzureActiveDirectoryAccessTokenGenerator = 'azure-active-directory-access-token-generator',
  AzureActiveDirectoryPassword = 'azure-active-directory-password',
  AzureActiveDirectoryServicePrincipalSecret = 'azure-active-directory-service-principal-secret',
  Ntlm = 'ntlm',
  Default = 'default',
});
```

src/connection.ts
funkydev: That shouldn't be a part of this PR 🙈

src/connection.ts
funkydev: **Update:**
We have to use enum, as because we won't able to use it as type in the interface. After the refactor it looks like: 
```typescript
export interface AzureActiveDirectoryMsiAppServiceAuthentication {
  type: SupportedAuthenticationTypes.AzureActiveDirectoryMsiAppService;
  options: {
    // ...
  };
}
```

When we change it as a const, it won't build.

arthurschreiber: We don't really do this with any other objects we export, so I'm not really sure this needs to be done. If people want to shoot themselves in the foot, it's on them. 🤷 

But if you want to go down this path, you need to define this like this:

```
export const AuthenticationTypes = Object.freeze({
  AzureActiveDirectoryMsiAppService: 'azure-active-directory-msi-app-service',
  AzureActiveDirectoryMsiVm: 'azure-active-directory-msi-vm',
  AzureActiveDirectoryAccessToken: 'azure-active-directory-access-token',
  AzureActiveDirectoryAccessTokenGenerator: 'azure-active-directory-access-token-generator',
  AzureActiveDirectoryPassword: 'azure-active-directory-password',
  AzureActiveDirectoryServicePrincipalSecret: 'azure-active-directory-service-principal-secret',
  Ntlm: 'ntlm',
  Default: 'default',
} as const);
```

I still do think that what I suggested should be enough - a `... as const` definition is always marked as `Readonly`, so it shouldn't be accidentally changeable from typescript, without requiring the use of `freeze`. But either solution is fine.

src/connection.ts
arthurschreiber: 😬

arthurschreiber: Don't worry. I'll let you finish this up first before leaving any more comments. 🙇‍♂️

src/connection.ts
funkydev: Readonly variables are understandable only for TypeScript when someone will use the `tedious` package without typescript, it would be allowed to modify the code, but as you told: 
> If people want to shoot themselves in the foot, it's on them.  

I totally agree with you 😄 

Backing to the option between const and enum, as I said, we have to use enum here.
As I tried to change it to the const, I couldn't build the project, because the following error occurred: 
> TS2503: Cannot find namespace 'AuthenticationTypes'.

src/connection.ts
funkydev: Sorry! That is PR only for the refactor. You saw my PoC of the new authentication type, but let's keep the discussion about that for the next PR :D

funkydev:
@arthurschreiber ready for re-review :D

arthurschreiber:
@funkydev Here's a diff of the changes I'd propose:

```diff
diff --git a/src/connection.ts b/src/connection.ts
index 68eb02a6..61869a90 100644
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -196,33 +196,49 @@ const DEFAULT_LANGUAGE = 'us_english';
  */
 const DEFAULT_DATEFORMAT = 'mdy';
 
-export enum AuthenticationTypes {
-  AzureActiveDirectoryMsiAppService = 'azure-active-directory-msi-app-service',
-  AzureActiveDirectoryMsiVm = 'azure-active-directory-msi-vm',
-  AzureActiveDirectoryAccessToken = 'azure-active-directory-access-token',
-  AzureActiveDirectoryPassword = 'azure-active-directory-password',
-  AzureActiveDirectoryServicePrincipalSecret = 'azure-active-directory-service-principal-secret',
-  Ntlm = 'ntlm',
-  Default = 'default',
+export const AuthenticationTypes = Object.freeze({
+  AzureActiveDirectoryMsiAppService: 'azure-active-directory-msi-app-service',
+  AzureActiveDirectoryMsiVm: 'azure-active-directory-msi-vm',
+  AzureActiveDirectoryAccessToken: 'azure-active-directory-access-token',
+  AzureActiveDirectoryPassword: 'azure-active-directory-password',
+  AzureActiveDirectoryServicePrincipalSecret: 'azure-active-directory-service-principal-secret',
+  Ntlm: 'ntlm',
+  Default: 'default',
+} as const);
+
+type FedAuthAuthenticationType =
+  typeof AuthenticationTypes.AzureActiveDirectoryPassword |
+  typeof AuthenticationTypes.AzureActiveDirectoryMsiVm |
+  typeof AuthenticationTypes.AzureActiveDirectoryMsiAppService |
+  typeof AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret;
+
+type AADAuthenticationType = typeof AuthenticationTypes.AzureActiveDirectoryAccessToken | FedAuthAuthenticationType;
+
+function isFedAuthAuthenticationType(type: AuthenticationType): type is FedAuthAuthenticationType {
+  switch (type) {
+    case AuthenticationTypes.AzureActiveDirectoryAccessToken:
+    case AuthenticationTypes.AzureActiveDirectoryMsiVm:
+    case AuthenticationTypes.AzureActiveDirectoryMsiAppService:
+    case AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret:
+      return true;
+
+    default:
+      return false;
+  }
 }
 
-const AADAuthenticationTypes = [
-  AuthenticationTypes.AzureActiveDirectoryPassword,
-  AuthenticationTypes.AzureActiveDirectoryAccessToken,
-  AuthenticationTypes.AzureActiveDirectoryMsiVm,
-  AuthenticationTypes.AzureActiveDirectoryMsiAppService,
-  AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret,
-];
-
-const FedAuthAuthentications = [
-  AuthenticationTypes.AzureActiveDirectoryPassword,
-  AuthenticationTypes.AzureActiveDirectoryMsiVm,
-  AuthenticationTypes.AzureActiveDirectoryMsiAppService,
-  AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret,
-];
+function isAADAuthenticationType(type: AuthenticationType): type is AADAuthenticationType {
+  switch (type) {
+    case AuthenticationTypes.AzureActiveDirectoryAccessToken:
+      return true;
+
+    default:
+      return isFedAuthAuthenticationType(type);
+  }
+}
 
 export interface AzureActiveDirectoryMsiAppServiceAuthentication {
-  type: AuthenticationTypes.AzureActiveDirectoryMsiAppService;
+  type: typeof AuthenticationTypes.AzureActiveDirectoryMsiAppService;
   options: {
     /**
      * If you user want to connect to an Azure app service using a specific client account
@@ -243,7 +259,7 @@ export interface AzureActiveDirectoryMsiAppServiceAuthentication {
 }
 
 export interface AzureActiveDirectoryMsiVmAuthentication {
-  type: AuthenticationTypes.AzureActiveDirectoryMsiVm;
+  type: typeof AuthenticationTypes.AzureActiveDirectoryMsiVm;
   options: {
     /**
      * If you user want to connect to an Azure app service using a specific client account
@@ -260,7 +276,7 @@ export interface AzureActiveDirectoryMsiVmAuthentication {
 }
 
 export interface AzureActiveDirectoryAccessTokenAuthentication {
-  type: AuthenticationTypes.AzureActiveDirectoryAccessToken;
+  type: typeof AuthenticationTypes.AzureActiveDirectoryAccessToken;
   options: {
     /**
      * A user need to provide `token` which they retrived else where
@@ -271,7 +287,7 @@ export interface AzureActiveDirectoryAccessTokenAuthentication {
 }
 
 export interface AzureActiveDirectoryPasswordAuthentication {
-  type: AuthenticationTypes.AzureActiveDirectoryPassword;
+  type: typeof AuthenticationTypes.AzureActiveDirectoryPassword;
   options: {
     /**
      * A user need to provide `userName` asscoiate to their account.
@@ -290,7 +306,7 @@ export interface AzureActiveDirectoryPasswordAuthentication {
 }
 
 export interface AzureActiveDirectoryServicePrincipalSecret {
-  type: AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret;
+  type: typeof AuthenticationTypes.AzureActiveDirectoryServicePrincipalSecret;
   options: {
     /**
      * Application (`client`) ID from your registered Azure application
@@ -308,7 +324,7 @@ export interface AzureActiveDirectoryServicePrincipalSecret {
 }
 
 export interface NtlmAuthentication {
-  type: AuthenticationTypes.Ntlm;
+  type: typeof AuthenticationTypes.Ntlm;
   options: {
     /**
      * User name from your windows account.
@@ -328,7 +344,7 @@ export interface NtlmAuthentication {
 }
 
 export interface DefaultAuthentication {
-  type: AuthenticationTypes.Default;
+  type: typeof AuthenticationTypes.Default;
   options: {
     /**
      * User name to use for sql server login.
@@ -3365,7 +3381,7 @@ Connection.prototype.STATE = {
 
             const { authentication } = this.config;
 
-            if (FedAuthAuthentications.includes(authentication.type)) {
+            if (isFedAuthAuthenticationType(authentication.type)) {
               this.transitionTo(this.STATE.SENT_LOGIN7_WITH_FEDAUTH);
             } else if (authentication.type === AuthenticationTypes.Ntlm) {
               this.transitionTo(this.STATE.SENT_LOGIN7_WITH_NTLM);
@@ -3389,7 +3405,7 @@ Connection.prototype.STATE = {
       featureExtAck: function(token) {
         const { authentication } = this.config;
 
-        if (AADAuthenticationTypes.includes(authentication.type)) {
+        if (isAADAuthenticationType(authentication.type)) {
           if (token.fedAuth === undefined) {
             this.loginError = ConnectionError('Did not receive Active Directory authentication acknowledgement');
             this.loggedIn = false;
```

It changes `AuthenticationTypes` back to an object, and uses the `typeof` operator to get the correct type in the places where a property is used and we want to keep the type intact.

It also replaces the `AADAuthenticationTypes` and `FedAuthAuthentications` arrays I proposed with two functions that allow checking the same condition, but without having an array defined and with type propagation. What do you think?

arthurschreiber:
One other thing I'm wondering - do we even want to export `AuthenticationTypes`? 🤔

funkydev:
Let me check that. I saw some castings, and I need to check if your proposition covers all of them. 

> One other thing I'm wondering - do we even want to export AuthenticationTypes? 🤔

Because we export the interfaces of authentication types (and we should do that), so we should do the same with authentication types. That would allow developers to use tedious internal constants instead of hardcoding strings.

Additionally we could also move authentication types (constants and interfaces) to a separate folder to organize code and decrease lines in that huge file.

funkydev:
That would allow to make the following files per authentication type

funkydev:
Finally, it would allow keeping the validation methods of each type in dedicated files instead of one huge file.

funkydev:
@arthurschreiber I have finished.

Summarizing: 
1. There is a `SupportedAuthenticationTypes` that is used for storing supported authentication types. It is a constant, as you asked.
2. I created `FedAuthAuthenticationType`, `AADAuthenticationType`, `AuthenticationType` types;
3. I created `isSupportedAuthenticationType`, `isFedAuthAuthenticationType`, `isAADAuthenticationType` methods for validating authentication types
4. I extracted all authentications to their files, and I created methods for validating and parsing options depends on the authentication type
5. I added tests for authentications 😎

arthurschreiber:
Thanks for preparing this PR! I haven't gotten around to reviewing this again, but I'll try to find some time over the coming days.

rafagsiqueira:
How about "azure-active-directory-default"? It is listed on the docs:
```
azure-active-directory-default
This uses DefaultAzureCredential to try multiple methods of authentication. You can find more information on what values to provide and how to provide them from [this Microsoft document.](https://docs.microsoft.com/en-us/javascript/api/overview/azure/identity-readme?view=azure-node-latest#defaultazurecredential)
```

PR #1244, opened 2021-04-15
fix: procedure validate errors not being caught

dhensby:
node-mssql tests have failed when I bumped from v11.0.5 -> v11.0.7 and one of those failures was because the error being emitted here wasn't being passed to the callback.

This fixes the problem


IanChokS:
Thanks @dhensby!

arthurschreiber:
Thanks @dhensby! 🙇‍♂️  I pushed a few small changes to the PR to make it follow the error handling pattern of other similar methods.

PR #1253, opened 2021-04-29
chore: remove codecov from CI tests

IanChokS:
Removes codecov form CI tests as it doesn't seem to be accurate in its results.


arthurschreiber:
I think https://github.com/tediousjs/tedious/pull/1272 fixed the issues with codecov, so I'm going to close this PR out. 😅 Let's hope codecov reports are more useful now. 🤞

PR #1262, opened 2021-05-13
fix: clear connection timeout on connection error

### LINKED ISSUE #1261: Errors on connection creation don't cancel the connection timeout timer
Please see https://github.com/tediousjs/node-mssql/issues/1226#issuecomment-840519960

It looks like the error handler in [this function](https://github.com/tediousjs/tedious/blob/08ebc25c738c200654dd6a645cf1d36ddbac672c/src/connection.ts#L1881-L1904) should call `this.clearConnectTimer();`

dhensby:
This fixes #1261 (at least under my test conditions).

I'm not sure if this is the best way to handle errors at connection time, but it seems that the time is not being cleared correctly if/when a connection errors before being established.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1904,6 +1904,7 @@ class Connection extends EventEmitter {
           this.connectOnPort(port, this.config.options.multiSubnetFailover, signal);
         });
       }, (err) => {
+        this.clearConnectTimer();
         if (err.name === 'AbortError') {
           // Ignore the AbortError for now, this is still handled by the connectTimer firing
           return;
@@ -1991,6 +1992,7 @@ class Connection extends EventEmitter {
         this.transitionTo(this.STATE.SENT_PRELOGIN);
       });
     }, (err) => {
+      this.clearConnectTimer();
       if (err.name === 'AbortError') {
         return;
       }
--- a/test/integration/connection-test.js
+++ b/test/integration/connection-test.js
@@ -5,7 +5,6 @@ const fs = require('fs');
 const homedir = require('os').homedir();
 const assert = require('chai').assert;
 const os = require('os');
-const dns = require('dns');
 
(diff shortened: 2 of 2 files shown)


dhensby:
@IanChokS / @arthurschreiber - is there any interest in getting this in? I've gone through and written a test that covers off the issue.

The crux of the matter is that when a connection is initialised a timer is created using `setTimeout` - under some connection errors, the timer is not cleared and so the node process hangs until the time ends - for applications with a long timeout, this can cause applications to hang for a long time before ending.

This fix just ensures that the connection timer is cleaned up when the errors are handled allowing the node process to exit immediately.

dhensby:
The test is a bit gross (I'm having to monkey-patch `dns.lookup` to force an error to throw at the right point), the only other alternative is to either install another dev dependency (like [`rewire`](https://github.com/jhnns/rewire)) or add some other config that would allow consumers to provide their own DNS lookup function (which feels very out of scope).

Nonetheless, the tests do replicate/prove the problem if you comment out the `clearConnectTimer()` calls.

arthurschreiber:
> The test is a bit gross (I'm having to monkey-patch `dns.lookup` to force an error to throw at the right point), the only other alternative is to either install another dev dependency (like [`rewire`](https://github.com/jhnns/rewire)) or add some other config that would allow consumers to provide their own DNS lookup function (which feels very out of scope).

I agree. I pushed a change to use a domain name like `something.invalid`, which is guaranteed to make the domain lookup fail (`.invalid` is a reserved top level domain name, see https://www.rfc-editor.org/rfc/rfc2606).

arthurschreiber:
Will merge once CI is ready.

dhensby:
👍 I had not realised there were specific reserved TLDs that will never resolve

PR #1280, opened 2021-07-03
New BulkLoad streaming API

arthurschreiber:
This adds a new `BulkLoad` API that greatly simplifies streaming data for bulk loading.

Instead of having to access a `BulkLoad`'s via the `.getRowStream` method, we can now simply pass in an `Iterable` (like a regular `Array` or a `Generator`) or an `AsyncIterable` (like a `ReadableStream` or an `AsyncGenerator`) to the `.execBulkLoad` method on a `Connection`.

Here's a very basic example that streams data from a regular `Array`:

```js
const bulkLoad = connection.newBulkLoad('employees', (err, rowCount) => {
  // ...
});

bulkLoad.addColumn('first_name', TYPES.NVarchar, { nullable: false });
bulkLoad.addColumn('last_name', TYPES.NVarchar, { nullable: false });
bulkLoad.addColumn('date_of_birth', TYPES.Date, { nullable: false });

connection.execBulkLoad(bulkLoad, [
  { 'first_name': 'Steve', 'last_name': 'Jobs', 'day_of_birth': new Date('02-24-1955') },
  { 'first_name': 'Bill', 'last_name': 'Gates', 'day_of_birth': new Date('10-28-1955') }
]);
```

The main advantage over the old API is that the same API covers now streaming and non-streaming use cases, and the User does not have to decide between one or the other - they're both implemented and used in the same way.

### DIFF
--- a/benchmarks/bulk-load/iterable.js
+++ b/benchmarks/bulk-load/iterable.js
@@ -1,0 +1,61 @@
+// @ts-check
+
+const { createBenchmark, createConnection } = require('../common');
+
+const { Request, TYPES } = require('../../src/tedious');
+
+const bench = createBenchmark(main, {
+  n: [10, 100],
+  size: [
+    10,
+    100,
+    1000,
+    10000
+  ]
+});
+
+function main({ n, size }) {
+  createConnection((connection) => {
+    const request = new Request(`
+      CREATE TABLE "#tmpTestTable" (
+        "id" int NOT NULL
+      )
+    `, (err) => {
+      if (err) {
+        throw err;
+      }
+
+      let i = 0;
+
+      bench.start();
+
+      (function cb() {
+        const bulkLoad = connection.newBulkLoad('#tmpTestTable', (err) => {
+          if (err) {
+            throw err;
+          }
+
+          if (i++ === n) {
+            bench.end(n);
+
+            connection.close();
+
+            return;
+          }
+
+          cb();
+        });
+
+        bulkLoad.addColumn('id', TYPES.Int, { nullable: false });
+
+        const rows = [];
+        for (let j = 0; j < size; j++) {
+          rows.push([ j ]);
+        }
+        connection.execBulkLoad(bulkLoad, rows);
+      })();
+    });
+
+    connection.execSqlBatch(request);
+  });
+}
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -372,7 +372,7 @@ class BulkLoad extends EventEmitter {
    *
    * @param name The name of the column.
    * @param type One of the supported `data types`.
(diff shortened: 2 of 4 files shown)


arthurschreiber:
@chdh I know you've not been active lately around `tedious`, but I'd still love to hear your feedback on this if you can spare a bit of your time? 🙇‍♂️ 

@dl-eric What do you think about this API? would this come in handy for your use cases of streaming bulk loads?

chdh:
@arthurschreiber
Thanks. I took a look, but I didn't understand all the details, mainly because I haven't learned the new async stream features yet. 
I appreciate that the API is now becoming compatible with the new streaming technology.

Maybe it would be reasonable to add a test to verify that flow control is working correctly when importing a large amount of data?

dl-eric:
Looks great to me - pretty clean and I can refactor to fit the new API pretty easily. My use case mainly concerns with bulk inserting generated/random data into a database as fast as possible.

arthurschreiber:
> Maybe it would be reasonable to add a test to verify that flow control is working correctly when importing a large amount of data?

I tried added some tests, and flow control seems to be working correctly, depending on what one thinks to be "correct".

Because there's multiple streams involved in a bulk load, there's multiple places where buffering happens:

* The `Readable` created by wrapping the passed in `Iterable` or `AsyncIterable` will buffer up to 16 objects.
* The `RowTransform` will buffer up to 16 objects on its `Writable` side, and up to `16kb` on its `Readable` side.
* The `Message`, which is basically just a `PassThrough` will buffer 16 objects on its `Writable` and `Readable` side.
* The `OutgoingMessageStream` buffers up to `16kb` on its `Readable` side, while the `Writable` side will never have more than one `Message` being written to it.
* For encrypted connections, there is the `DuplexPair` object that I think again buffers up to `16kb` on both of its Duplex streams (both `Readable` and `Writable` sides I think?)
* At the end, there is the network socket, which also buffers up to `16kb` on the writable side.

Phew - that's a lot of streams. 😅 And because of that, there's a lot of buffering involved as well. It's probably not ideal, but I think it's "acceptable", as overall, there should be no excessive buffering, even if all intermediate streams need to make full use of their buffering capabilities. ​I think that long term, there's probably some improvements that could be made (e.g. use generators internally, and only wrap them in Node.js streams at the edge when interacting with Node.js APIs).

But if we focus just on the change proposed in this PR, it just adds the additional `16` object buffering for the `Readable` that wraps the `Iterable` / `AsyncIterable` - nothing else is being changed.

arthurschreiber:
> But if we focus just on the change proposed in this PR, it just adds the additional `16` object buffering for the `Readable` that wraps the `Iterable` / `AsyncIterable` - nothing else is being changed.

Oh, and the nice thing is that flow control in this part is always doing "the right thing" because it's using `Readable.from` and `.pipe` internally, so there's nothing that the user of this API has to think about.

PR #1282, opened 2021-07-06
feat: add `order` option to `BulkLoad`

arthurschreiber:
This adds a new `order` option to `Connection.prototype.newBulkLoad` that allows specifying the order of bulk loaded data.

This can increase the performance of the bulk load if the data is ordered in the same way as the target table of the bulk load.

Closes: https://github.com/tediousjs/tedious/issues/1197
Closes: https://github.com/tediousjs/tedious/pull/1198

### DIFF
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -48,6 +48,7 @@ interface InternalOptions {
   fireTriggers: boolean;
   keepNulls: boolean;
   lockTable: boolean;
+  order: { [columnName: string]: 'ASC' | 'DESC' };
 }
 
 export interface Options {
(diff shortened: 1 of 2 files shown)


arthurschreiber:
@dl-eric I'm sorry it has taken me so long to get back to you here!

I recently fixed another issue around bulk loads (https://github.com/tediousjs/tedious/pull/1281) and figured I should take another look at your PR!

I updated your PR with a similar fix as in #1281 for "exotic" column names, and added additional integration tests to cover different a few more edge cases. I squashed all your changes and my fixes into a single commit.

Does this look good to you? 🙇‍♂️

dl-eric:
Looks amazing, no worries about the delay. Thanks so much for this! lgtm :)

PR #1291, opened 2021-07-21
Allow type checking on JavaScript files

arthurschreiber:
This enables the `allowJs` option in the typescript configuration, which allows the use of the typescript compiler to type check `.js`.

Type checking will be opt-in initially, and can be enabled for individual files by adding a `// @ts-check` comment. Having all test files type checked will ensure that the types defined in the TypeScript code are "correct" and is a first step to including TypeScript definition files in our npm packages in the future.


arthurschreiber:
I had to fix some of the type information in the `Connection` class, as the tests were exposing some use cases that I didn't know about when some of the code was initially converted to TypeScript. I think we'll find a few more of these going forward, which will help us making our type information more accurate, and will allow us to actually include the type information in the released packages.

PR #1295, opened 2021-07-21
use nchar and nvarchar type IDs plus ucs2 encoding for char and varchar

### LINKED ISSUE #1294: two-byte characters in char and varchar are truncated to one byte
Duplicate of #723

Char and varchar are currently configured to read the values passed to them purely as ascii, but char and varchar columns in MSQL can support the entire BMP (i.e. one- and two-byte UTF-16 characters). On read this is handled by decoding records with iconv, but the current state is that storing a two-byte character such as `"\u2021"` (‡) in a char or varchar column will truncate that character to only the second byte, resulting in an incorrect character being stored, such as `"\u0021"` (!).

We could resolve this by encoding char and varchar data with nchar and nvarchar type IDs when serializing them into the RPC stream, as in the attached PR. Is there any reason not to?

Ceshion:
Closes #1294

### DIFF
--- a/src/data-types/char.ts
+++ b/src/data-types/char.ts
@@ -2,6 +2,11 @@ import { DataType } from '../data-type';
 
 const NULL_LENGTH = Buffer.from([0xFF, 0xFF]);
 
+
+// we are using this when sending type info in order to support
+// all BMP characters in char columns.
+const NCHAR_TYPE_ID = 0xEF;
+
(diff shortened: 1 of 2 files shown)


arthurschreiber:
@Ceshion I'm going to close this out, as I believe we "correctly" support `varchar`/`char`/`text` encodings now. Please let me know if my assumption is wrong! 🙇‍♂️

PR #1307, opened 2021-08-01
feat: drop deprecated bulk load APIs

arthurschreiber:
BREAKING CHANGE: This drops the `BulkLoad.addRow` and `BulkLoad.getRowStream` methods, in favor of passing an `Iterable` or `AsyncIterable` to the `Connection.execBulkLoad` method.


arthurschreiber:
This was merged in #1357.

PR #1314, opened 2021-08-06
Improve collation handling

arthurschreiber:
This refactors the collation parsing logic and fixes some issues around parsing the different collation flags.

It also adds tracking of collation information on the connection, and also ensures we write back the collation for `varchar` / `nvarchar` / `char` / `nchar` / `text` / `ntext` values.

### DIFF
--- a/src/collation.ts
+++ b/src/collation.ts
@@ -1,10 +1,14 @@
+type Encoding = 'utf-8' | 'CP437' | 'CP850' | 'CP874' | 'CP932' | 'CP936' | 'CP949' | 'CP950' | 'CP1250' | 'CP1251' | 'CP1252' | 'CP1253' | 'CP1254' | 'CP1255' | 'CP1256' | 'CP1257' | 'CP1258';
+
 // http://technet.microsoft.com/en-us/library/aa176553(v=sql.80).aspx
-export const codepageByLcid: { [key: number]: string | undefined } = {
+export const codepageByLanguageId: { [key: number]: Encoding } = {
   // Arabic_*
   [0x0401]: 'CP1256',
 
   // Chinese_Taiwan_Stroke_*
   // Chinese_Traditional_Stroke_Count_*
+  // Chinese_Taiwan_Bopomofo_*
+  // Chinese_Traditional_Bopomofo_*
   [0x0404]: 'CP950',
 
   // Czech_*
@@ -33,6 +37,7 @@ export const codepageByLcid: { [key: number]: string | undefined } = {
   [0x040D]: 'CP1255',
 
   // Hungarian_*
+  // Hungarian_Technical_*
   [0x040E]: 'CP1250',
 
   // Icelandic_*
@@ -40,6 +45,8 @@ export const codepageByLcid: { [key: number]: string | undefined } = {
 
   // Japanese_*
   // Japanese_XJIS_*
+  // Japanese_Unicode_*
+  // Japanese_Bushu_Kakusu_*
   [0x0411]: 'CP932',
 
   // Korean_*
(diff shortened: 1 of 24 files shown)


arthurschreiber:
@jeremy-w Mind taking a look? 👀

src/token/token.ts
```diff
@@ -281,10 +282,10 @@ export class CollationChangeToken extends Token {
   declare event: 'sqlCollationChange';
 
   type: 'SQL_COLLATION';
-  oldValue: Buffer;
-  newValue: Buffer;
+  oldValue: Collation | undefined;
+  newValue: Collation | undefined;
```
jeremy-w, line 286: Would a collation envchange ever be sent without a new collation?

src/connection.ts
```diff
@@ -1063,6 +1066,7 @@ class Connection extends EventEmitter {
 
     this.fedAuthRequired = false;
     this.fedAuthInfoToken = undefined;
+    this.collation = undefined;
```
jeremy-w, line 1069: Would it be worth naming more narrowly as databaseCollation? I wonder if we’ll wind up needing to snag serverCollation manually at some point in order to work with tempdb parameters that show up in executed sql while working with a database whose collation is not the server collation.

jeremy-w:
I looked over the diff. I might not have a chance to run it or do any further investigation till Monday, but it looks like we arrived at the same nibbles. https://github.com/jeremy-w/tedious/commit/dca84f0c51209648a9f1ecb09d84c4c59eb65e21

I hadn’t got to using the collation yet, though. Since you’ve picked up using the collation, I can drop that commit.

Do you want me to PR the utf8 login7 & featureextack commits? They would get client and server opted into utf8. (I still don’t know what happens if the db collation is utf8 and that’s not enabled. Disconnect?)

src/collation.ts
```diff
@@ -182,3 +182,70 @@ export const codepageBySortId: { [key: number]: string | undefined } = {
...
+
+  private buffer: Buffer | undefined;
+
+  static fromBuffer(buffer: Buffer, offset = 0) {
+    let lcid = (buffer[offset + 2] & 0x0F) << 16;
+    lcid |= buffer[offset + 1] << 8;
+    lcid |= buffer[offset + 0];
+
+    let flags = (buffer[offset + 3] & 0x0F) << 4;
+    flags |= (buffer[offset + 2] & 0xF0) >> 4;
```
jeremy-w, line 211: I suspect all our right shifts should either be unsigned (using `>>>` instead of `>>`) or careful to mask to the expected bitwidth to avoid sign extension turning on bits that shouldn’t be on.

src/connection.ts
arthurschreiber: Sounds good to me! This property is intended to be "private" for now.

Regarding the tempdb, I don't think that's on `tedious` really? I think if your application creates tables in the tempdb, it should make sure those use the same collation as your database - that will ensure that `varchar` data send to tempdb tables will have the correct encoding.

Even if we would track the collation of the server (and thus the collation used for tempdb), SQL Server would still expect `varchar` parameters as encoded with the current database's encoding.

src/token/token.ts
arthurschreiber: Practically, probably not, but technically the server can send back a zero length `BVARBYTE` value.

src/collation.ts
arthurschreiber: Done! Thanks for the suggestion. 🙇‍♂️

arthurschreiber:
@jeremy-w I just pushed a few more changes that should make sending `varchar`, `char` and `text` values as expected.

I think the only place where using these data types does not work quite right yet is bulk loads, but hopefully that's not too complicated to fix either.

@heroboy Do you mind taking a look as well? 🙇‍♂️

arthurschreiber:
@Ceshion Would you also mind taking a look? I think this should also fix the issues you had around using these data types.

arthurschreiber:
> Do you want me to PR the utf8 login7 & featureextack commits?

Yes, I think it makes sense if you open a separate PR for those changes.

jeremy-w:
👍 PR opened: #1317

Issue #1143, opened 2020-08-11
Support for VarChar UTF-8


aklietz:
SQL Server 2019 now supports UTF-8 natively, stored internally as `VarChar(n)` using a collation such as `Latin1_General_100_CI_AS_SC_UTF8`.  (CI = case insensitive, AS = accent sensitive, SC = include supplementary UTF8 characters).  A native UTF-8 database column saves 50% space over `NVarChar` for databases that contain mostly Western European content. It also reduces the TDS network traffic by a similar amount. It's a big win.

I've been trying to coax Tedious to accept VarChar bindings containing UTF-8 strings, but no joy. I found in bug in `mssql` where it was wrongly passing in UTF-8 character lengths using `string.length` when it needed to send byte lengths using `Buffer.byteLength(string, 'utf8')` for `VarChar(n)`, but the UPDATEs were still getting munged at the Tedious level.  Packet tracing shows the UTF-8 2-byte values are getting garbled in weird ways.  For example the copyright symbol © should be encoded as 0xa9 0x0a but the TDS packet trace shows it being sent as 0xa9 0x0e (huh?).  

Furthermore, SQL Server chops out any incoming multibyte UTF-8 characters on UPDATE or INSERT (another bug?).  I'm also wondering if there is a magic 'Use UTF-8' flag that needs to be enabled in the TDS packet protocol for UPDATE and INSERT.  SELECT seems to work okay.

IanChokS:
Hi @aklietz 

Thanks for raising this issue. We'll get started to work on UTF-8 support!

jeremy-w:
This is a blocker for using tedious (via knex 0.95.0+) for me as well. UTF-8 is as much a default choice for encoding for me as UTC is for timezone.

@aklietz Re: the magic 'Use UTF-8' flag, I don't know about update/insert in particular, but I had a peek at the TDS v30 2020-10-01 doc, and at the session-establishment level, the LOGIN7 stream indeed has a UTF8_SUPPORT FeatureEx introduced in TDS 7.4. The collation rules used in COLMETADATA also have an fUTF8 flag.

jeremy-w:
@IanChokS I'm not clear on how Tedious tracks the column collation for writing data even absent UTF-8 support. Does it currently not, and that's worked because most folks are using NVARCHAR exclusively?

I see when parsing values, the Metadata is used to drive codepage selection, with a default encoding of utf8 when codepage is not available.

[...]

PR #1317, opened 2021-08-07
feat: log in with UTF-8 option [#1143]

jeremy-w:
Before, we did not send the UTF-8 option during login, nor did we parse the corresponding FeatureExtAck from the server.

Now, we do.

(Though TBH, I'm not clear on how this actually changes the server behavior; it seemed perfectly happy to send UTF-8 encoded data on down to versions of the library that did not express UTF-8 support. So we could maybe drop the last commit, which remembers the status, since we probably don't need to actually check it ever.)

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3614,10 +3614,10 @@ Connection.prototype.STATE = {
             this.loginError = ConnectionError(`Active Directory authentication acknowledgment for ${authentication.type} authentication method includes extra data`);
             this.loggedIn = false;
           }
-        } else if (token.fedAuth === undefined) {
+        } else if (token.fedAuth === undefined && token.utf8Support === undefined) {
           this.loginError = ConnectionError('Received acknowledgement for unknown feature');
           this.loggedIn = false;
-        } else {
+        } else if (token.fedAuth) {
           this.loginError = ConnectionError('Did not request Active Directory authentication, but received the acknowledgment');
           this.loggedIn = false;
         }
--- a/src/login7-payload.ts
+++ b/src/login7-payload.ts
@@ -1,4 +1,5 @@
 import { sprintf } from 'sprintf-js';
+import { versions } from './tds-versions';
 
 const FLAGS_1 = {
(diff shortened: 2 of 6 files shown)


src/connection.ts
```diff
@@ -891,6 +891,12 @@ class Connection extends EventEmitter {
    * @private
    */
   fedAuthInfoToken: undefined | FedAuthInfoToken;
+  /**
+   * @private
+   * Set when FeatureExtAck for server UTF-8 support is received.
+   * Applicable only when tdsVersion >= 7.4.
+   */
+  utf8Support?: boolean;
```
arthurschreiber, line 899: Would it make sense to default this to `false`? 🤔

jeremy-w: i was making a three-way distinction of n/a, true, false. but i'm tempted to not save it at all; i don't know that we'll ever actually need to refer to it once we turn it on.

jeremy-w: further argument in favor of dropping the commit that saves that in the connection: mssql-jdbc doesn't save it, either.

https://github.com/microsoft/mssql-jdbc/blob/0c007b5f678f26f00b84e28895ee449dafd753a9/src/main/java/com/microsoft/sqlserver/jdbc/SQLServerConnection.java#L5041-L5050

jeremy-w:
I dropped the commit remembering the state after login and rebased atop master. I believe this is ready for merge.

arthurschreiber:
> I believe this is ready for merge.

Thanks for preparing this! ❤️ I'm planning to do a few last cleanups on #1314, and will have these two merged soon! 👍

arthurschreiber:
> it seemed perfectly happy to send UTF-8 encoded data on down to versions of the library that did not express UTF-8 support

It will change behaviour, I think. Without this option, `varchar`/`char`/`text` columns that use an UTF-8 encoding will be sent back as `nvarchar` data. Also, data send to the server is still expected to be encoded using the regular codepages, not UTF-8. I think this is done for backwards compatibility purposes.

Issue #1324, opened 2021-08-19
CVE-2021-32796 due to dependency on deprecated adal-node package which in turn uses xmldom@0.6.0


prueker:
Tedious needs to upgrade to @azure/msal-node to mitigate CVE-2021-32796 which is caused by xmldom as a dependency on deprecated adal-node package as per https://github.com/AzureAD/azure-activedirectory-library-for-nodejs.

https://nvd.nist.gov/vuln/detail/CVE-2021-32796

```
└─┬ tedious@11.4.0
    └─┬ adal-node@0.2.2
      └── xmldom@0.6.0
```

wickedest:
Note, due to inability take ownership of xmldom, it is now published as [@xmldom/xmldom@0.7.0](https://www.npmjs.com/package/@xmldom/xmldom) by the new maintainers ([ref](https://github.com/xmldom/xmldom/discussions/270)).

rossgp:
Also worth noting there's a dependency via @azure/ms-rest-nodeauth as well

```
└─┬ tedious@11.4.0
  ├─┬ @azure/ms-rest-nodeauth@3.0.10
  │ └── adal-node@0.2.2  deduped
  └── adal-node@0.2.2
```

And that's not going to be addressed according to #https://github.com/Azure/ms-rest-nodeauth/issues/128
and are recommending moving to @azure/identity instead

MichaelSun90:
Thank you guys for the suggestions. We are currently looking into migrating to using @azure/identity.  Like I replied in issue #1238, we will give you guys an update.

jasonnutter:
`adal-node@0.2.3` has been published with `@xmldom/xmldom`.

wickedest:
if anyone else is wondering, the reply from @jasonnutter effectively says that this issue is resolved and this issue can be closed.

Issue #1238, opened 2021-03-15
Deprecated dependency libraries: request, adal, ms-rest-nodeauth


mcgrewsm:
Tedious is dependent on these libraries that are no longer being maintained:

"@azure/ms-rest-nodeauth": "^3.0.6",
"adal-node": "^0.2.1"

ms-rest-node-auth is deprecated and uses an old version of adal-node, 0.1.28. That 0.1.28 version of adal-node uses the deprecated request library, which is a flagged security vulnerability.

The 0.2.x versions of adal-node use axios instead of request, which is fine, however, I don't believe ms-rest-nodeauth will ever be able to pick up that update because of the jump from 0.1.x to 0.2.x. In addition, adal-node itself is deprecated, in favor of the new library MSAL.

Ideal fix for this issue would be if Tedious was able to switch to MSAL, but I am not sure of the scope of that.

The big issue is really the deprecated/vulnerable **request** library down the dependency chain, but unfortunately I cannot see how this can be solved lower down the chain since the in-between projects themselves are deprecated.

adal page saying to switch to MSAL: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs#readme
Migration page / MSAL library for JS: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/migration.md
"ms-rest-nodeauth is in maintenance mode": https://github.com/Azure/ms-rest-nodeauth/issues/84
"adal is in maintenance mode": https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/issues/229#issuecomment-590518066

IanChokS:
Thanks for pointing this out! We were actually looking into migrating to [@azure/identity](https://www.npmjs.com/package/@azure/identity) in replace of [@azure/ms-rest-nodeauth ](https://www.npmjs.com/package/@azure/ms-rest-nodeauth), which fortunately uses the new [@azure/msal-node](https://www.npmjs.com/package/@azure/identity) dependency.

mcgrewsm:
Hi, it is good news that there seems to be another solution to this problem. I am not partial to one library or the other, as long as there are no vulnerabilities in the tree.

opravil-jan:
Hi, 
when this will be done please?
Thanks

MichaelSun90:
Hi @opravil-jan We just start looking into this migration and plan to work on this in the near future, but we do not have a concrete estimate yet. We will update you when we finish it or we come up with a solid estimate.

PR #1328, opened 2021-09-05
feat: Migrate to azure identity

akshayganeshen:
**Before submitting a PR :**

- [x] Ensure your fork is created from `master` branch of [the repository](https://github.com/tediousjs/tedious).
- [x] Run `npm install` in the root folder.
- [x] After bug fix/code change, ensure all the existing tests and new tests (if any) pass (`npm run-script test-all`). During development, to run individual test use `node_modules/nodeunit test/<test_file.js> -t <test_name>`.
- [x] Build the driver (`npm run build`).
- [x] Run eslint and flow typechecker (`npm run lint`).
- [x] Run commitlint (`node_modules/.bin/commitlint --from origin/master --to HEAD`). Refer [commit conventions](https://commitlint.js.org/#/concepts-commit-conventions) and [commit rules](https://commitlint.js.org/#/reference-rules).

**Description of changes**

Migrate usages of `@azure/ms-rest-nodeauth` and `adal-node` to use `@azure/identity` instead. This removes the transitive dependency on `xmldom@0.6.0` (#1238, #1324).

**Testing**

Needs testing for the following authentication types:

- [ ] `azure-active-directory-password`
- [ ] `azure-active-directory-msi-vm`
- [ ] `azure-active-directory-msi-app-service`
- [ ] `azure-active-directory-service-principal-secret`

All other authentication types have been left untouched.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -8,13 +8,11 @@ import { createSecureContext, SecureContext, SecureContextOptions } from 'tls';
 import { Readable } from 'stream';
 
 import {
-  loginWithVmMSI,
-  loginWithAppServiceMSI,
-  UserTokenCredentials,
-  MSIVmTokenCredentials,
-  MSIAppServiceTokenCredentials,
-  ApplicationTokenCredentials
-} from '@azure/ms-rest-nodeauth';
+  ClientSecretCredential,
+  ManagedIdentityCredential,
+  TokenCredential,
+  UsernamePasswordCredential,
+} from '@azure/identity';
 
 import BulkLoad, { Options as BulkLoadOptions, Callback as BulkLoadCallback } from './bulk-load';
 import Debug from './debug';
@@ -42,7 +40,6 @@ import { createNTLMRequest } from './ntlm';
 import { ColumnMetadata } from './token/colmetadata-token-parser';
 import { ColumnEncryptionAzureKeyVaultProvider } from './always-encrypted/keystore-provider-azure-key-vault';
 import depd from 'depd';
-import { MemoryCache } from 'adal-node';
 
 import { AbortController, AbortSignal } from 'node-abort-controller';
 import { Parameter, TYPES } from './data-type';
@@ -50,6 +47,7 @@ import { BulkLoadPayload } from './bulk-load-payload';
 import { Collation } from './collation';
 
 import { version } from '../package.json';
+import { URL } from 'url';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const deprecate = depd('tedious');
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -3718,54 +3713,46 @@ Connection.prototype.STATE = {
...
-                  resource: fedAuthInfoToken.spn,
-                  clientId: authentication.options.clientId
-                }, getTokenFromCredentials);
+                getTokenFromCredentials(credentials);
+              } else if (authentication.type === 'azure-active-directory-msi-vm' || authentication.type === 'azure-active-directory-msi-app-service') {
+                const msiArgs = authentication.options.clientId ? [ authentication.options.clientId, {} ] : [ {} ];
+                const credentials = new ManagedIdentityCredential(...msiArgs);
+
+                if (authentication.options.msiEndpoint) {
+                  process.env.IDENTITY_ENDPOINT = authentication.options.msiEndpoint;
```
akshayganeshen, line 3739: Note to reviewers: I don't like that this assigns values to `process.env`, but I couldn't find any alternatives in the `@azure/identity` options...

https://github.com/Azure/azure-sdk-for-js/blob/0596e3a57f84cd3289c4bdad1271da7aa1f8eb37/sdk/identity/identity/src/credentials/managedIdentityCredential/arcMsi.ts#L86

https://github.com/Azure/azure-sdk-for-js/blob/0596e3a57f84cd3289c4bdad1271da7aa1f8eb37/sdk/identity/identity/src/credentials/managedIdentityCredential/appServiceMsi2017.ts#L51-L52

arthurschreiber:
This is great, thank you! ❤️ I'll take a closer look in the next few days!

src/connection.ts
```diff
@@ -3718,54 +3713,46 @@ Connection.prototype.STATE = {
...
-
-                credentials!.getToken().then((tokenResponse: { accessToken: string | undefined }) => {
-                  callback(null, tokenResponse.accessToken);
-                }, callback);
-              };
+              const getTokenFromCredentials = (credentials: TokenCredential) =>
+                credentials.getToken([ tokenScope ]).then(
+                  (tokenResponse) => callback(null, tokenResponse?.token),
+                  callback
+                );
```
arthurschreiber, line 3723: Would you mind adding braces here for readability?

```suggestion
              const getTokenFromCredentials = (credentials: TokenCredential) => {
                credentials.getToken([ tokenScope ]).then(
                  (tokenResponse) => callback(null, tokenResponse?.token),
                  callback
                );
              };
```

src/connection.ts
arthurschreiber: Should we maybe just deprecate and remove the `msiEndpoint` and  `msiSecret` options? What do you think?

akshayganeshen: Yeah I was thinking the same.

Would you want to keep the changes and just add some `depd` logs + documentation updates? I can take a look at adding those updates if so.

arthurschreiber: I think we should prepare a PR that will deprecate explicitly setting these options (and make them fallback to the environment variables if not set explicitly), and release that first.

And then mark these changes here as breaking and release them with a new major version bump.

arthurschreiber:
I'm going to close this one in favor of https://github.com/tediousjs/tedious/pull/1335. For security reasons, pull requests originating from other repositories than `tediousjs/tedious` do not execute any of the Azure CI tests, so I cleaned up the changes, fixed a few issues and opened a new PR.

@akshayganeshen Thank you so much for this contribution! 🙇‍♂️

PR #1355, opened 2021-10-06
feat: allow specifying the `clientId` when using `azure-active-directory-password` auth

arthurschreiber:


mShan0:
Did we want the warning to emit even if the client ID was specified?

arthurschreiber:
> Did we want the warning to emit even if the client ID was specified?

I'm a bit confused. I don't think we want that, and I don't think the current code does that - it only emits the warning if the client id is not specified. Or am I misunderstanding your question?

arthurschreiber:
Ah, now I see! 💡 You're right, we should only emit the warning if the value is `undefined`. Would you mind fixing this up?

mShan0:
will do!

src/connection.ts
```diff
@@ -1085,12 +1091,19 @@ class Connection extends EventEmitter {
           throw new TypeError('The "config.authentication.options.password" property must be of type string.');
         }
 
+        if (options.clientId !== undefined && typeof options.clientId !== 'string') {
+          throw new TypeError('The "config.authentication.options.clientId" property must be of type string.');
+        } else {
+          emitAzureADPasswordClientIdDeprecationWarning();
```
arthurschreiber, line 1097: We're emitting a deprecation warning unless the user specified a value for `clientId`, because we want to require the user to pass in a `clientId` value in the future. 👍

PR #1386, opened 2022-01-21
feat: rename `domain` to `tenantId` when using `azure-active-director…

mShan0:
Renames `config.authenticaiton.options.domain` to `config.authenticaiton.options.tenantId`
when using the `azure-active-directory-password` authentication type.

`domain` can still currently be used but will emit a deprecation warning message if used.

Documentation changes: #1389

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -56,6 +56,16 @@ const emitTrustServerCertificateWarning = () => {
   }
 };
 
+let domainRenameToTenantIdWarningEmitted = false;
+const emitDomainRenameToTenantIdWarning = () => {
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -3432,8 +3452,8 @@ Connection.prototype.STATE = {
 
               if (authentication.type === 'azure-active-directory-password') {
                 const credentials = new UsernamePasswordCredential(
-                  authentication.options.domain,
                   authentication.options.clientId,
+                  authentication.options.tenantId ?? 'common',
```
mShan0, line 3456: I need to swap these arguments

PR #1399, opened 2022-02-09
docs: update connection and faq with nltm and node 17 info

mShan0:
Document changes on explaining how to use NTLM with node 17

### DIFF
--- a/frequently-encountered-problems.html
+++ b/frequently-encountered-problems.html
@@ -166,6 +166,41 @@ title: F.E.Q.
     </dd>
     <br />
     <dt>
+        <h2>Using NTLM authentication with Node.js 17</h2>
+    </dt>
+    <dd>
+        <p>
+            Node 17 has upgraded to OpenSSL 3.0, which considers md4 encryption a legacy type due to md4 being insecure. 
+            Since NTLM uses md4 encryption, you must enable the `--openssl-legacy-provider` command line option in order to use
+            NTLM authentication with Node 17. If you attempt to use NTLM without the option enabled, 
+            it will throw an error and not allow you to connect.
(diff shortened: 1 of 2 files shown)


frequently-encountered-problems.html
```diff
@@ -165,6 +165,40 @@ <h2>The Behavior Change of connect() function</h2>
...
     </dd>
     <br />
+    <dt>
+        <h2>Using NTLM authentication with Node.js 17</h2>
+    </dt>
+    <dd>
+        <p>
+            Node 17 has upgraded to OpenSSL 3.0, which considers md4 encryption a legacy type. 
+            Since NTLM uses md4 encryption, you must enable the `--openssl-legacy-provider` command line option in order to use
+            NTLM authentication with Node 17.
```
MichaelSun90, line 175: Do we want to mention, if the user does not enable this, we will throw an error? Just give users a heads up, there may be an error coming if they did not do this. But I am also concerned that mentioning the error message here and the error message itself should be also self-explanatory as well. Will this be a kind of repeated information?

mShan0: Good point. I'll add that in

mShan0: I can take out the "check the FAQ" part in the actual error message in the code to avoid repetition. But the FAQ includes an explanation on how to enable it if the user doesn't know how to

MichaelSun90: I think we can leave the check the FAQ in the error message, just encase the user can reach this piece of info from anywhere.  Or, do you think we can directly mention that legacy option in the error message instead of pointing the user to FAQ? Not sure if that will be too complicated for a error message?

PR #1400, opened 2022-02-09
feat: fail ntlm auth gracefully when md4 hashing is not available

mShan0:
Emits an error explaining that the user needs to enable `--openssl-legacy-provider`. Also adds tests to test the flag's functionality. 

Document changes: #1399 (already merged)
CI Config testing node 17: #1396

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3404,27 +3404,39 @@ Connection.prototype.STATE = {
               this.transitionTo(this.STATE.LOGGED_IN_SENDING_INITIAL_SQL);
             }
           } else if (this.ntlmpacket) {
-            const authentication = this.config.authentication as NtlmAuthentication;
+            try {
+              const authentication = this.config.authentication as NtlmAuthentication;
 
-            const payload = new NTLMResponsePayload({
-              domain: authentication.options.domain,
-              userName: authentication.options.userName,
-              password: authentication.options.password,
-              ntlmpacket: this.ntlmpacket
-            });
+              const payload = new NTLMResponsePayload({
+                domain: authentication.options.domain,
+                userName: authentication.options.userName,
+                password: authentication.options.password,
+                ntlmpacket: this.ntlmpacket
+              });
 
-            this.messageIo.sendMessage(TYPE.NTLMAUTH_PKT, payload.data);
-            this.debug.payload(function() {
-              return payload.toString('  ');
-            });
+              this.messageIo.sendMessage(TYPE.NTLMAUTH_PKT, payload.data);
+              this.debug.payload(function() {
+                return payload.toString('  ');
+              });
 
-            this.ntlmpacket = undefined;
+              this.ntlmpacket = undefined;
 
-            this.messageIo.readMessage().then((message) => {
-              this.dispatchEvent('message', message);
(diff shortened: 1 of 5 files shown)


src/connection.ts
```diff
@@ -3368,27 +3368,38 @@ Connection.prototype.STATE = {
...
+
+              this.messageIo.sendMessage(TYPE.NTLMAUTH_PKT, payload.data);
+              this.debug.payload(function() {
+                return payload.toString('  ');
+              });
+
+              this.ntlmpacket = undefined;
+
+              this.messageIo.readMessage().then((message) => {
+                this.dispatchEvent('message', message);
```
MichaelSun90, line 3389: Not sure if you need to try to catch the whole block. I have not tried it myself, but you could try to check if the income message or the err object has a code match ERR_OSSL_EVP_UNSUPPORTED, and emit the error.

test/integration/connection-test.js
```diff
@@ -455,17 +456,39 @@ describe('Ntlm Test', function() {
...
+      runNtlmTest.call(this, done, DomainCaseEnum.Upper);
+    });
+
+  } else {
+
+    it('should throw a connection error with node 17', () => {
+      const child = childProcess.spawnSync(process.execPath,
+                                           ['./test/integration/child-processes/ntlm-connect-node17.js'],
+                                           { encoding: 'utf8' });
+      assert.strictEqual(child.status, 1);
```
MichaelSun90, line 479: Maybe we can also do an assert check for the error messages?

test/unit/ntlm-payload-test.js
```diff
@@ -1,48 +1,66 @@
...
+      assert.strictEqual(targetData, 'aaaaaaaa');
+
+      assert.strictEqual(expectedLength, response.data.length);
+    });
+  });
+} else {
+  describe('ntlm payload test node 17 and newer', function() {
+
+    it('should throw error when `--openssl-legacy-provider` is not enabled', function() {
+      assert.throws(() => new NTLMPayload(challenge));
```
MichaelSun90, line 57: Same as the integrate test, maybe we can do an assert test for the error message just to make sure.

src/connection.ts
```diff
@@ -3368,27 +3368,38 @@ Connection.prototype.STATE = {
...
+              }, (err) => {
+                this.socketError(err);
+              });
+            } catch (error: any) {
+              if (error.code === 'ERR_OSSL_EVP_UNSUPPORTED') {
+                this.emit('connect', new ConnectionError(`Node 17 now uses OpenSSL 3, which considers md4 encryption a legacy type.
+                In order to use NTLM with Node 17, enable the '--openssl-legacy-provider' command line flag.
+                Check the Tedious FAQ for more information.`, 'ELOGIN'));
+              } else {
+                this.emit('connect', new ConnectionError('Login failed.' + error.message, 'ELOGIN'));
```
arthurschreiber, line 3399: I think it would be best to just re-throw any "unknown" errors we encounter here. Otherwise other errors like syntax errors or trying to call methods on an `undefined` object and so on will get turned into `ConnectionErrors` here.

What do you think?

PR #1409, opened 2022-02-23
feat: better error handling and test fixes

MichaelSun90:
Added better error handling for request errors. The aggregate error will catch all the errors emitted along the process, so the user will get the whole picture.

Also, added an aggregate error for catching AAD token retrieve error, so not only the generic error message will be emitted but also the original server-side error.

Fix some tests that related to this behavior change


src/connection.ts
```diff
@@ -3510,7 +3520,9 @@ Connection.prototype.STATE = {
 
             getToken((err, token) => {
               if (err) {
-                this.loginError = new ConnectionError('Security token could not be authenticated or authorized.', 'EFEDAUTH');
+                this.debug.log(err.name + ' : ' + err.message);
```
arthurschreiber, line 3523:
```suggestion
```

PR #1410, opened 2022-03-01
docs: add faq for encryption options

mShan0:
Also updates connection.options.cryptoCredentialsDetails.

### DIFF
--- a/frequently-encountered-problems.html
+++ b/frequently-encountered-problems.html
@@ -201,6 +201,54 @@ C:\Tedious> node --openssl-legacy-provider example.js
     </dd>
     <br />
     <dt>
+        <h2>The relations between <code>options.trustServerCertificate, options.encrypt,</code> and <code>options.cryptoCredentialDetails</code></h2>
+    </dt>
+    <dd>
+        <ul>
+            <li>
+                If your server has encryption set to 'ON' or 'REQ', and:
+                <ul>
+                    <li>
+                        <code>encrypt = true</code> in your config, the driver will start TLS
+                        with the corresponding <code>trustServerCertificate</code> settings.
+                    </li>
+                    <li>
+                        <code>encrypt = false</code>, the driver will throw an error 
+                        stating that <code>Server requires encryption, set 'encrypt' option to true.</code> then close the connection.
+                    </li>
+                </ul>
+            </li>
+            <li>
+                If <code>encrypt = true</code> and <code>trustServerCertificate = false</code>, 
+                then you may be required to provide your certificate details using <code>cryptoCredentialsDetails</code>. 
+                Details on how to provide the <code>cryptoCredentialsDetails</code> object can be found in the
+                 <a href="/api-connection.html">connection</a> page of the documentation.
+            </li>
+            <li>
+                If <code>encrypt = false</code>, then <code>trustServerCertificate</code> is ignored.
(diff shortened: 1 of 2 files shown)


frequently-encountered-problems.html
```diff
@@ -200,6 +200,39 @@ <h2>Using NTLM authentication with Node.js 17</h2>
...
+
+    </dd>
+    <br />
+    <dt>
+        <h2>TLS 1.0 compatibility with Node 12 and newer</code></h2>
+    </dt>
+    <dd>
+        <p>
+            The default TLS settings were tightened in Node 12 and newer, requiring TLS 1.2 as the minimum version. You can find documentation
+            on how to upgrade your server to TLS 1.2
```
MichaelSun90, line 225: just a reminder in case we forget: we can say 1.2 or above, just to be safe.

frequently-encountered-problems.html
```diff
@@ -200,6 +200,39 @@ <h2>Using NTLM authentication with Node.js 17</h2>
         </p>
     </dd>
     <br />
+    <dt>
+        <h2>The relations between <code>options.trustServerCertificate, options.encrypt,</code> and <code>options.cryptoCredentialDetails</code></h2>
+    </dt>
+    <dd>
+        <ul>
+            <li>
+                If <code>encrypt = true</code> and <code>trustServerCertificate = false</code>, then <code>cryptoCredentialsDetails</code> must be provided.
```
MichaelSun90, line 209: I am wondering will there be some kind of error if the user does not provide "cryptoCredentialsDetails" here? if cryptoCredentialsDetails is empty, does this meaning tls.createSecureContext() will loading a empty option when try to create the secure context?

frequently-encountered-problems.html
```diff
@@ -200,6 +200,39 @@ <h2>Using NTLM authentication with Node.js 17</h2>
...
+        <h2>The relations between <code>options.trustServerCertificate, options.encrypt,</code> and <code>options.cryptoCredentialDetails</code></h2>
+    </dt>
+    <dd>
+        <ul>
+            <li>
+                If <code>encrypt = true</code> and <code>trustServerCertificate = false</code>, then <code>cryptoCredentialsDetails</code> must be provided. 
+                Details on how to provide the <code>cryptoCredentialsDetails</code> object can be found in the <a href="/api-connection.html">connection</a> page of the documentation.
+            </li>
+            <li>
+                If <code>encrypt = false</code>, then <code>trustServerCertificate</code> is ignored.
```
MichaelSun90, line 213: This is what I found when we has the discussion with David, seems the current logic will attempt to check the certificate only based on the server-side settings. Not 100% sure, but you can double-check the logic, see if I got anything wrong. 

If server-side has set to encrypt to 'ON' or 'REQ', encrypt is set to true.  The driver will start TLS  with corresponding trustservercertificate settings.
If server-side has set to encrypt to 'ON' or 'REQ', and encrypt is set to false. The driver will throw an error with the message "Server requires encryption, set 'encrypt' config option to true." and close the connection.

mShan0: Added to the FAQ

frequently-encountered-problems.html
mShan0: Good point. Added it

PR #1411, opened 2022-03-01
docs: add clientid field to azure ad password auth

mShan0:


api-connection.html
```diff
@@ -112,15 +112,20 @@ <h2 id="function_newConnection">new Connection(config)</h2>
...
 {% endhighlight %}
           <ul>
             <li>username (string): A user needs to provide the `userName` associated to their account.</li>
             <li>password (string): A user needs to provide the `password` associated to their account.</li>
+            <li>clientId (string): Application (`client`) ID from your registered Azure application. This currently is optional
+              and uses a default clientId, but a user-provided clientId will be required in the near future.
+            </li>
             <li>tenantId (string): Optional parameter for specific Azure tenant ID. Note: this used to be called 
-              <code>domain</code>. <code>domain</code> will proceed to work but will be removed in the near future.</li>
+              <code>domain</code>. <code>domain</code> will proceed to work but will be removed in the near future. (default: <code>true</code>)
```
MichaelSun90, line 127: For the optional part, I think we could also add an explanation of why it is optional, but also mostly always needed for most cases. 
if tenantId does not provide during the connection, then MSAL will use common or consumer as the tenant for this connection. If these two tenants are not supported on the server-side setup, then the server will return the error: "The grant type is not supported over the /common or /consumers endpoints. Please use the /organizations or tenant-specific endpoint." Therefore, even though, tenantId is optional here, users should choose to provide this piece of information based on their server setting.

MichaelSun90 (commented):
only on comment for the tenant id part.

MichaelSun90 (approved):
only on comment for the tenant id part.

api-connection.html
mShan0: Updated in the latest commit

PR #1430, opened 2022-04-22
docs: add faq section for unicode conversions

mShan0:
Adds FAQ section explaining how Tedious handles Unicode conversions

### DIFF
--- a/frequently-encountered-problems.html
+++ b/frequently-encountered-problems.html
@@ -249,7 +249,7 @@ C:\Tedious> node --openssl-legacy-provider example.js
     </dd>
     <br />
     <dt>
-        <h2>How Tedious handles Unicode datatype conversions when using parameters</code></h2>
+        <h2>How Tedious handles Unicode datatype conversions when using parameters</h2>
     </dt>
     <dd>
         <p>
@@ -259,7 +259,7 @@ C:\Tedious> node --openssl-legacy-provider example.js
             (like characters that can't be translated to the target encoding).
         </p>
         <p>
-            There are some alternatives to this. You can use a Unicode datatype as the parameter type. Tedious will send the string as Unicode, 
+            There are some alternatives to this. You can use a Unicode datatype as the parameter type. Tedious will send the string in the SQL server's default encoding, 
             then the SQL server will do the conversion, which may cause a performance hit in certain situations.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


frequently-encountered-problems.html
```diff
@@ -248,6 +248,25 @@ <h2>TLS 1.0 compatibility with Node 12 and newer</code></h2>
...
     </dd>
     <br />
+    <dt>
+        <h2>How Tedious handles Unicode datatype conversions when using parameters</code></h2>
+    </dt>
+    <dd>
+        <p>
+            Since all JavaScript strings are Unicode, when using a non-Unicode datatype (such as VarChar), then Tedious will do the conversion 
+            from a Unicode string to a non-Unicode string then send the data to the server. 
+            The conversion may not be what the SQL server expects and cause inaccurate results.
```
arthurschreiber, line 258: Maybe we should describe that we're using `iconv-lite` to perform this conversion? The conversion might not match the way SQL Server would do the conversion in some edge cases (like characters that can't be translated to the target encoding).

frequently-encountered-problems.html
```diff
@@ -248,6 +248,25 @@ <h2>TLS 1.0 compatibility with Node 12 and newer</code></h2>
...
+        </p>
+        <p>
+            There are some alternatives to this. You can use a Unicode datatype as the parameter type. Tedious will send the string as Unicode, 
+            then the SQL server will do the conversion. This will cause a performance hit.
```
arthurschreiber, line 262: > This will cause a performance hit.

I don't think that's always going to be true - we probably should reword this. 😅

frequently-encountered-problems.html
mShan0: Mentioned iconv-lite and added the additional info on the conversion

frequently-encountered-problems.html
mShan0: Good point. I reworded it

frequently-encountered-problems.html
```diff
@@ -248,6 +248,26 @@ <h2>TLS 1.0 compatibility with Node 12 and newer</code></h2>
         </p>
     </dd>
     <br />
+    <dt>
+        <h2>How Tedious handles Unicode datatype conversions when using parameters</code></h2>
```
arthurschreiber, line 252:
```suggestion
        <h2>How Tedious handles Unicode datatype conversions when using parameters</h2>
```

arthurschreiber (approved):

PR #1435, opened 2022-05-09
refactor: cleanup `Connection` internals

arthurschreiber:
This refactors the `Connection` internals, so far focusing on the functionality around connection establishing.

This replaces the state machine, state switching and event based logic. Instead, we now use `async` methods.

This has the following advantages:

* The flow of execution is easier to understand. E.g. `.connect` calls `.establishConnection`, which calls `.handlePreloginResponse`, which calls one of the `LOGIN7` response handling methods like `.handleLogin7WithStandardLoginResponse`.
* Error handling is much improved. Previously, errors could happen _anywhere_ in the flow. Either low level socket errors, or higher level "login failure" errors. They were passed to `.socketError`, which in turn needed to have state specific logic for error handling and cleanup. Now, errors just get `throw`n and handled inside of the `.connect` method in a consistent fashion.
* Cancellation and connect timeout handling are also now explicitly handled via `AbortSignal`s. Previously, e.g. connection timeout was handled by having a timer emit the `connectTimeout` state event, which each state then tried to handle correctly (but mostly failed). Now, each `async` method takes an `signal` argument that is used to detect whether the current operation has been cancelled or not, and can act accordingly.

There's still some state handling code, for the `SENT_ATTENTION` and `SENT_CLIENT_REQUEST` states, and this probably should be refactored as well.

I left the state transition calls in place (`.transitionTo`), mostly for debug output purposes, but I think this can be completely removed at some later time.

This is a very large set of changes, and I feel like it's a bit risky. It probably warrants adding more extensive test cases (especially for failure conditions like timeouts or invalid messages being sent from the server).

@mShan0 @MichaelSun90 What do you think?

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -17,7 +17,7 @@ import {
 
 import BulkLoad, { Options as BulkLoadOptions, Callback as BulkLoadCallback } from './bulk-load';
 import Debug from './debug';
-import { EventEmitter } from 'events';
+import { EventEmitter, once } from 'events';
 import { InstanceLookup } from './instance-lookup';
 import { TransientErrorLookup } from './transient-error-lookup';
 import { TYPE } from './packet';
@@ -48,6 +48,9 @@ import AggregateError from 'es-aggregate-error';
 import { version } from '../package.json';
 import { URL } from 'url';
 import { AttentionTokenHandler, InitialSqlTokenHandler, Login7TokenHandler, RequestTokenHandler, TokenHandler } from './token/handler';
+import { promisify } from 'util';
+import { AbortError } from '@azure/abort-controller';
+import { addAbortSignal } from './util/add-abort-signal';
 
 let trustServerWarningEmitted = false;
 
@@ -343,9 +346,9 @@ interface DefaultAuthentication {
   };
 }
 
-interface ErrorWithCode extends Error {
-  code?: string;
-}
+// interface ErrorWithCode extends Error {
+//   code?: string;
+// }
 
 interface InternalConnectionConfig {
   server: string;
@@ -1027,6 +1030,11 @@ class Connection extends EventEmitter {
    * @private
    */
   databaseCollation: Collation | undefined;
+
+  /**
+   * @private
+   */
+  closeController: AbortController;
 
   /**
    * Note: be aware of the different options field:
@@ -1744,6 +1752,8 @@ class Connection extends EventEmitter {
       this.messageIo.sendMessage(TYPE.ATTENTION);
       this.createCancelTimer();
     };
+
+    this.closeController = new AbortController();
   }
 
   connect(connectListener?: (err?: Error) => void) {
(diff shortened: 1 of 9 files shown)


src/connection.ts
```diff
@@ -1774,6 +1777,442 @@ class Connection extends EventEmitter {
...
+  async lookupInstancePort(signal: AbortSignal) {
+    const lookup = new InstanceLookup();
+
+    return (await promisify(lookup.instanceLookup.bind(lookup))({
+      server: this.config.server,
+      instanceName: this.config.options.instanceName!,
+      timeout: this.config.options.connectTimeout,
+      signal: signal
+    }))!;
+  }
```
arthurschreiber, line 1823: The instance lookup code itself should be refactored to use promises. See https://github.com/tediousjs/tedious/pull/1211 for some previous work that was never finished.

src/connection.ts
```diff
@@ -1774,6 +1777,442 @@ class Connection extends EventEmitter {
...
+
+    this.messageIo = new MessageIO(socket, this.config.options.packetSize, this.debug);
+    this.messageIo.on('secure', (cleartext) => { this.emit('secure', cleartext); });
+
+    this.socket = socket;
+
+    this.closed = false;
+    this.debug.log('connected to ' + this.config.server + ':' + this.config.options.port);
+
+    this.sendPreLogin();
```
arthurschreiber, line 1867: Sending of messages currently does not take an `AbortSignal`. It would be important to change this for scenarios where sending data to the server is extremly slow and reaches the connection timeout. Without signal support, the connection would just end up being stuck.

Also, test cases for this would be good.

src/connection.ts
```diff
@@ -1774,6 +1777,442 @@ class Connection extends EventEmitter {
...
+
+        const message = `Connection lost - ${err.message}`;
+        this.debug.log(message);
+        throw new ConnectionError(message, 'ESOCKET');
+      }
+
+      const handler = new Login7TokenHandler(this);
+      const tokenStreamParser = this.createTokenStreamParser(message, handler);
+
+      await once(tokenStreamParser, 'end');
```
arthurschreiber, line 1968: The `tokenStreamParser` does not honor the `AbortSignal` yet. Potentially `addAbortSignal(message, signal)` before the `this.createTokenStreamParser` might be enough, but I haven't tried that yet.

src/connection.ts
```diff
@@ -1991,38 +2401,21 @@ class Connection extends EventEmitter {
...
+        host: this.routingData ? this.routingData.server : this.config.server,
+        port: this.routingData ? this.routingData.port : port,
+        localAddress: this.config.options.localAddress
+      };
 
-    new Connector(connectOpts, signal, multiSubnetFailover).execute((err, socket) => {
-      if (err) {
-        if (err.name === 'AbortError') {
-          return;
+      new Connector(connectOpts, signal, multiSubnetFailover).execute((err, socket) => {
```
arthurschreiber, line 2412: The `Connector` class should be changed to provide an `async` API. And maybe change this from a class to just a single method?

src/message-io.ts
```diff
@@ -133,8 +135,26 @@ class MessageIO extends EventEmitter {
...
+
+      new Promise((resolve, reject) => {
+        onAbort = () => { reject(new AbortError()); };
+        signal.addEventListener('abort', onAbort);
+      }),
+
+      new Promise((resolve, reject) => {
+        onSocketError = reject;
+        this.socket.addListener('error', onSocketError);
+      })
```
arthurschreiber, line 153: We have logic here to ensure that an `error` event while waiting for a new message is handled correctly. But I'm wondering what happens if a socket error happens before the call to `readMessage`? 🤔 and unhandled rejection maybe?

PR #1438, opened 2022-05-12
refactor: further `Connection` state implementation refactor

mShan0:
Continuation of #1437. 

This refactors some of the state handling in the Connection class to make use of async/await.

The goal here is to gradually move us to a more readable, maintainable and overall more robust connection handling logic.

This PR changes the following states:

- `SENT_LOGIN7_WITH_FEDAUTH`
- `SENT_ATTENTION`
- `SENT_CLIENT_REQUEST`

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3691,12 +3691,40 @@ Connection.prototype.STATE = {
   SENT_ATTENTION: {
     name: 'SentAttention',
     enter: function() {
-      this.emptyMessageBuffer();
+      (async () => {
+        this.emptyMessageBuffer();
 
-      this.messageIo.readMessage().then((message) => {
-        this.dispatchEvent('message', message);
-      }, (err) => {
-        this.socketError(err);
+        let message;
+        try {
+          message = await this.messageIo.readMessage();
+        } catch (err: any) {
+          return this.socketError(err);
+        }
+
+        const handler = new AttentionTokenHandler(this, this.request!);
+        const tokenStreamParser = this.createTokenStreamParser(message, handler);
+
+        await once(tokenStreamParser, 'end');
+        // 3.2.5.7 Sent Attention State
+        // Discard any data contained in the response, until we receive the attention response
+        if (handler.attentionReceived) {
+          this.clearCancelTimer();
+
+          const sqlRequest = this.request!;
+          this.request = undefined;
+          this.transitionTo(this.STATE.LOGGED_IN);
+
+          if (sqlRequest.error && sqlRequest.error instanceof RequestError && sqlRequest.error.code === 'ETIMEOUT') {
+            sqlRequest.callback(sqlRequest.error);
+          } else {
+            sqlRequest.callback(new RequestError('Canceled.', 'ECANCEL'));
+          }
+        }
+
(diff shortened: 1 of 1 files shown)


src/connection.ts
```diff
@@ -3456,102 +3456,108 @@ Connection.prototype.STATE = {
...
+            case 'azure-active-directory-service-principal-secret':
+              credentials = new ClientSecretCredential(
+                authentication.options.tenantId,
+                authentication.options.clientId,
+                authentication.options.clientSecret
+              );
+              break;
+            default:
+              throw new Error();
+          }
```
mShan0, line 3516: What do you think of using a switch statement instead of if-else clauses here?

arthurschreiber: I think using a switch statement is a good idea. 👍

Thanks to the switch statement, I think you can drop the `default` case as well, as TypeScript can understand that the cases in the switch statement itself cover all potential values of `authentication.type`.

src/connection.ts
```diff
@@ -3456,102 +3456,108 @@ Connection.prototype.STATE = {
...
-            if (isTransientError(this.loginError)) {
-              this.debug.log('Initiating retry on transient error');
-              this.transitionTo(this.STATE.TRANSIENT_FAILURE_RETRY);
-            } else {
-              this.emit('connect', this.loginError);
-              this.transitionTo(this.STATE.FINAL);
-            }
+          if (!tokenResponse) {
+            throw new Error('no valid token');
+          }
```
arthurschreiber, line 3531: I think `tokenResponse` is always set to an actual value here, so this whole `if` condition can safely be removed. 👍

src/connection.ts
```diff
@@ -3456,102 +3456,108 @@ Connection.prototype.STATE = {
...
+
+          const token = tokenResponse.token;
+          this.sendFedAuthTokenMessage(token!);
```
arthurschreiber, line 3534: I think `token` is neither `null` nor `undefined` here, so you can change this from `token!` to `token`.

src/connection.ts
```diff
@@ -3456,102 +3456,108 @@ Connection.prototype.STATE = {
...
+          return;
+        }
 
+        const fedAuthInfoToken = handler.fedAuthInfoToken;
+
+        if (fedAuthInfoToken && fedAuthInfoToken.stsurl && fedAuthInfoToken.spn) {
+          const authentication = this.config.authentication as AzureActiveDirectoryPasswordAuthentication | AzureActiveDirectoryMsiVmAuthentication | AzureActiveDirectoryMsiAppServiceAuthentication | AzureActiveDirectoryServicePrincipalSecret | AzureActiveDirectoryDefaultAuthentication;
+          const tokenScope = new URL('/.default', fedAuthInfoToken.spn).toString();
+
+          let credentials: TokenCredential;
```
arthurschreiber, line 3487: because the switch statement below covers all potential values that `authentication.type` can have, and each switch statement assigns a value to `credentials`, there is no need to specify the type of this variable here.

arthurschreiber (commented):
Looking great! 👍 I left just a few smaller comments.

PR #1444, opened 2022-06-10
refactor: remove callbacks from `metadata parser`

mShan0:
Just an idea, converts callbacks to async/await for the column metadata parser-related classes. Would need to do more work to make all of the other tests pass if we want to continue with this approach.

### DIFF
--- a/src/metadata-parser.ts
+++ b/src/metadata-parser.ts
@@ -327,8 +327,159 @@ function metadataParse(parser: Parser, options: InternalConnectionOptions, callb
   });
 }
 
+// ------------- Remove callbacks ----------------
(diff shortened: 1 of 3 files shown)


arthurschreiber:
@mShan0 Thanks for opening this PR!

While this makes the code a lot easier to read and understand, performance might become a problem here. Each call to `await` requires to go through at least one iteration of the event loop. When parsing a packet, this can easily add up to thousands of iterations through the event loop and a severe degradation of the data parsing performance.

I think eventually, we will have to bite the bullet and refactor all of the parsing code to be based on something that neither uses promises nor callbacks. I'll try to come up with an example that should allow you to take over and apply it to all of the parsing code.

PR #1451, opened 2022-06-23
feat: support NTLM authentication on Node.js 17+ out of the box

mShan0:
NTML authentication requires the use of the MD4 hash algorithm, which is disabled by default in Node.js 17 and newer. It can be re-enabled by specifying the `--openssl-legacy-provider` to the `node` executable, but that's a suboptimal user experience.

Instead, we switch to using `js-md4` instead, which is a pure JavaScript implementation of the MD4 hashing algorithm.

MD4 is disabled by default in Node.js 17+ because of security concerns, but we want to support NTLM authentication as long as it's supported by Windows / SQL Server as well.


arthurschreiber:
Is there one dependency (`js-md4` vs `hash-wasm`) that you prefer?

mShan0:
I think i prefer `js-md4` if the TS definitions aren't too much of a hassle

src/ntlm-payload.ts
```diff
@@ -143,7 +144,7 @@ class NTLMResponsePayload {
 
   ntHash(text: string) {
     const unicodeString = Buffer.from(text, 'ucs2');
-    return crypto.createHash('md4').update(unicodeString).digest();
+    return Buffer.from(md4.hex(unicodeString), 'hex');
```
arthurschreiber, line 147:
```suggestion
    return Buffer.from(md4.arrayBuffer(unicodeString));
```

types/js-md4.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'js-md4' {
+  declare const md4: {
+    arrayBuffer(message: string | ArrayBuffer): string;
```
arthurschreiber, line 3: Isn't this returning an `ArrayBuffer`? 🤔

PR #1470, opened 2022-07-15
refactor: remove callbacks from metadataParse

mShan0:
Would it be a good idea to create a new stream parser-style class to handle these parsers?

Planning to move the parser position reset upon retry into metadataParse.


arthurschreiber:
Parser architecture was replaced by #1580, so I'm closing this.

PR #1471, opened 2022-07-20
fix: set maximum supported TLS version to 1.2

mShan0:
Modifies tests to pass on Node 18.6.0+

### DIFF
--- a/test/integration/connection-test.js
+++ b/test/integration/connection-test.js
@@ -289,7 +289,6 @@ describe('Initiate Connect Test', function() {
 
     // Specify a cipher that should never be supported by SQL Server
     config.options.cryptoCredentialsDetails = {
-      maxVersion: 'TLSv1.2',
       ciphers: 'NULL'
     };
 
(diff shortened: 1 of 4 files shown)


test/integration/connection-test.js
```diff
@@ -289,6 +289,7 @@ describe('Initiate Connect Test', function() {
 
     // Specify a cipher that should never be supported by SQL Server
     config.options.cryptoCredentialsDetails = {
+      maxVersion: 'TLSv1.2',
```
arthurschreiber, line 292: I think we should move this out of the tests somewhere else. This is going to fix the test failures, but I think we should make sure that we only ever use `TLSv1.2` and never `TLSv1.3` inside tedious when TDS < 8.0 is used - and only support `TLSv1.3` once we add TDS 8.0 support later.

mShan0: Just committed some changes. Was this closer to what you had in mind?

Issue #1468, opened 2022-07-13
Replace `JSBI` with native `BigInt`


arthurschreiber:
We used JSBI as a polyfill for `BigInt` on older Node.js versions. All versions of Node.js that we currently support also support `BigInt` out-of-the box, so we can drop the dependency on that.

MichaelSun90:
Hi @arthurschreiber , I am current working on this, and had already replace the places that uses JSBI with BigInt. Found something with in  writable-tracking-buffer.ts:
```js
 writeBigInt64LE(value: number) {
    this.writeBigU_Int64LE(value);
  }

  writeInt64LE(value: number) {
    this.writeBigInt64LE(value);
  }

  writeUInt64LE(value: number) {
    this.writeBigUInt64LE(value);
  }

  writeBigUInt64LE(value: number) {
    this.writeBigU_Int64LE(value);
  }

  private writeBigU_Int64LE(value: number) {
```
I move the functions a bit, so this group function can be compared together. For writeInt64LE, it calls writeBigInt64LE which calls writeBigU_Int64LE. Is there a specific reason that we cannot directly call writeBigU_Int64LE which holds the actual logic. Also, I found that  writeBigUInt64LE is not called anywhere else. Not sure why we need to keep this function. If there is no specific reason for this, I can refactor this part as well.

arthurschreiber:
I think you can remove those / clean them up. The Buffer class in Nodejs provides int64 read and write methods now.

PR #1472, opened 2022-07-25
feat: replace `JSBI` with native `BigInt`

MichaelSun90:
We used JSBI as a polyfill for BigInt on older Node.js versions. All versions of Node.js that we currently support also support BigInt out-of-the box, so we can drop the dependency on that. Resolving issue #1468

### DIFF
--- a/src/data-types/bigint.ts
+++ b/src/data-types/bigint.ts
@@ -30,9 +30,7 @@ const BigInt: DataType = {
     if (parameter.value == null) {
       return;
     }
-    // const buffer = Buffer.alloc(8);
-    // buffer.writeBigInt64LE(parameter.value);
-    // buffer.
+
     const buffer = new WritableTrackingBuffer(8);
     buffer.writeInt64LE(Number(parameter.value));
     yield buffer.data;
--- a/src/data-types/bigint.ts
+++ b/src/data-types/bigint.ts
@@ -30,7 +30,9 @@ const BigInt: DataType = {
     if (parameter.value == null) {
       return;
     }
-
+    // const buffer = Buffer.alloc(8);
+    // buffer.writeBigInt64LE(parameter.value);
+    // buffer.
     const buffer = new WritableTrackingBuffer(8);
     buffer.writeInt64LE(Number(parameter.value));
     yield buffer.data;
--- a/src/token/stream-parser.ts
+++ b/src/token/stream-parser.ts
@@ -255,7 +255,7 @@ class Parser {
     });
   }
 
-  readBigInt64LE(callback: (data: BigInt) => void) {
+  readBigInt64LE(callback: (data: bigint) => void) {
     this.awaitData(8, () => {
       const data = this.buffer.readBigInt64LE(this.position);
       this.position += 8;
@@ -279,7 +279,7 @@ class Parser {
     });
   }
 
-  readBigUInt64LE(callback: (data: BigInt) => void) {
+  readBigUInt64LE(callback: (data: bigint) => void) {
     this.awaitData(8, () => {
       const data = this.buffer.readBigUInt64LE(this.position);
       this.position += 8;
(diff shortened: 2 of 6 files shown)


src/token/stream-parser.ts
```diff
@@ -298,14 +293,14 @@ class Parser {
     });
   }
 
-  readBigUInt64LE(callback: (data: JSBI) => void) {
+  readBigUInt64LE(callback: (data: BigInt) => void) {
     this.awaitData(8, () => {
-      const low = JSBI.BigInt(this.buffer.readUInt32LE(this.position));
-      const high = JSBI.BigInt(this.buffer.readUInt32LE(this.position + 4));
+      const low = BigInt(this.buffer.readUInt32LE(this.position));
+      const high = BigInt(this.buffer.readUInt32LE(this.position + 4));
```
arthurschreiber, line 299: Instead of doing the calculations ourselves here, I think we can just call out to [`readBigUInt64LE`](https://nodejs.org/docs/latest-v14.x/api/buffer.html#buffer_buf_readbiguint64le_offset).

src/token/stream-parser.ts
```diff
@@ -256,26 +255,22 @@ class Parser {
...
+        ) +
+        BigInt(
           this.buffer[this.position] +
           this.buffer[this.position + 1] * 2 ** 8 +
           this.buffer[this.position + 2] * 2 ** 16 +
           this.buffer[this.position + 3] * 2 ** 24
-        )
-      );
-
+        );
```
arthurschreiber, line 273: Instead of doing the calculations ourselves here, I think we can just call out to [`readBigInt64LE`](https://nodejs.org/docs/latest-v14.x/api/buffer.html#buffer_buf_readbigint64le_offset).

src/data-types/bigint.ts
```diff
@@ -30,7 +30,9 @@ const BigInt: DataType = {
     if (parameter.value == null) {
       return;
     }
-
+    // const buffer = Buffer.alloc(8);
+    // buffer.writeBigInt64LE(parameter.value);
+    // buffer.
```
MichaelSun90, line 35: Forget to remove this comment, will remove it. Is this what we want for cleaning up the writing functions from writable buffer and use the native buffer writing functions instead?

arthurschreiber, line 35: I think the comment here can be removed. 🙇‍♂️

src/token/stream-parser.ts
MichaelSun90: got it, I will do a fallowing up commit for this. Thanks for clear this up!

test/unit/tracking-buffer/writable-tracking-buffer-test.js
```diff
@@ -116,18 +115,18 @@ describe('Wrtiable Tracking Buffer', () => {
     assertBuffer(buffer, [0x03, 0x00, 0x61, 0x00, 0x62, 0x00, 0x63, 0x00]);
   });
 
-  it('should write 64-bit signed JSBIs', () => {
+  it('should write 64-bit signed BIGINTs', () => {
```
arthurschreiber, line 118:
```suggestion
  it('should write 64-bit signed `BigInt`s', () => {
```

test/unit/tracking-buffer/writable-tracking-buffer-test.js
```diff
@@ -116,18 +115,18 @@ describe('Wrtiable Tracking Buffer', () => {
...
     const buffer = new TrackingBuffer(8);
 
-    buffer.writeBigInt64LE(JSBI.BigInt('0x0807060504030201'));
+    buffer.writeBigInt64LE(BigInt('0x0807060504030201'));
 
     assertBuffer(buffer, [0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08]);
   });
 
-  it('should write 64-bit unsigned JSBIs', () => {
+  it('should write 64-bit unsigned BIGINTs', () => {
```
arthurschreiber, line 126:
```suggestion
  it('should write 64-bit unsigned `BigInt`s', () => {
```

PR #1478, opened 2022-08-08
fix: export AggregateError

dhensby:
`AggregateError`s are being used to wrap multiple errors in v15, but the `AggregateError` class is not being exported, meaning it's not possible to perform `instanceof` checks on errors. This class should be exported so that these types of checks can be performed reliably.

### DIFF
--- a/src/tedious.ts
+++ b/src/tedious.ts
@@ -8,6 +8,7 @@ import { ConnectionError, RequestError } from './errors';
 import { TYPES } from './data-type';
 import { ISOLATION_LEVEL } from './transaction';
 import { versions as TDS_VERSION } from './tds-versions';
+import AggregateError from 'es-aggregate-error';
 
 const library = { name: name };
 
@@ -18,6 +19,7 @@ export function connect(config: ConnectionConfiguration, connectListener?: (err?
 }
 
 export {
+  AggregateError,
   BulkLoad,
   Connection,
   Request,


MichaelSun90:
Hi @dhensby,  I brought this PR to @arthurschreiber the other day, and seems he has some concern about expose the aggregate error from the tedious side. If we expose it from tedious side, then tedious side will kind be responsible for it, and this seems not really ideal. Also, When node 14 can be drop from the tedious support, aggregate error will not longer need to be import specifically since it will be natively support from node side. This problem should be resolved by then.

dhensby:
OK - it's fairly standard practice for a library to export it's dependencies like this otherwise it's not possible for consumers to perform `instanceof` checks or similar.

If a consumer includes the dependency in their own `package.json` and imports it then attempts to do an `instanceof` check, it can be very fragile as if the versions diverge at all then npm will install both and the `instanceof` will fail.

I'm not sure exactly what you mean by "will be responsible for it". It's a straight export of a dependency and it allows consumers to make concrete assertions about the errors that are emitted by the library. At the moment it has to be duck-typed instead, which is somewhat acceptable, but just means the API is less clearly defined and it's harder for consumers to be clear on the APIs being used.

arthurschreiber:
> OK - it's fairly standard practice for a library to export it's dependencies like this otherwise it's not possible for consumers to perform `instanceof` checks or similar.

I don't think exposing dependencies is a particularly great idea, because that can lead to changes in dependencies breaking consumers of a library.

I don't think you should rely on `instanceof` checks for errors at all. Either check for the `name` property (`err.name === 'AggregateError'`), or just check if the returned error has a `.errors` that is an array (`Array.isArray(err.errors)`). The latter is exactly what Node.js is doing:

https://github.com/nodejs/node/blob/eb298dff8d70d5049b8d7787c0418da146e69cb6/lib/internal/util/inspect.js#LL1373-L1376C4

dhensby:
Ok. That just means this library is recommending duck typing instead of concrete typing.

I'm happy not to argue with that and will work on a duck typed implementation.

There are many libraries that export the concrete types of dependencies (like the Microsoft/azure storage libraries) but that doesn't mean we all have to.

PR #1484, opened 2022-10-24
test: clean up typecast through type assertions

arthurschreiber:
This removes the `@types/chai` package and instead places a copy of the type definitions to `types/chai.d.ts`.

I made a few changes to the type definitions for `chai` and opened a pull request over at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60464. Unfortunately, the changes haven't been merged yet, and I didn't have time to fix up the PR.

To clean up the tests in `tedious`, I decided that it'll be easier to vendor the type definitions for `chai` for the time being, and then re-open that pull request at some later time.

One of the changes I made to the `chai` definitions is to change `assert.instanceOf` to perform a type assertion. This helps TypeScript's understanding of what types a variable can have after the call to the assertion method. A similar change was also done to `assert.exists`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15474 @@
 {
   "name": "tedious",
   "version": "0.0.0-dev",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "tedious",
+      "version": "0.0.0-dev",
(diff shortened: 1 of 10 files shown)


arthurschreiber:
Upstream `@types/chai@4.3.20` now ships the `asserts value is T` narrowing; the test cleanup itself landed via #1713 / #1719.

PR #1494, opened 2022-11-05
ci: update github actions

Fdawgs:
This PR bumps GitHub Actions to their latest major versions, which now use Node 16 internally instead of Node 12 (EOL as of end of April 2022, no longer receives security updates):

- Releases for `actions/cache` can be [found here](https://github.com/actions/cache/releases)
- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)
-   Changelog for `github/codeql-action` actions can be [found here](https://github.com/github/codeql-action/blob/main/CHANGELOG.md)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
     - name: Use Node.js 14
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -15,9 +15,9 @@ jobs:
     name: Linting
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/nodejs.yml
```diff
@@ -15,9 +15,9 @@ jobs:
     name: Linting
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
```
arthurschreiber, line 18: Looks like this change breaks the commit lint job because newer versions of the checkout action only fetch one commit, not the full history. Could you change this to fetch the whole history for the lint job?

Fdawgs: Thanks @arthurschreiber, https://github.com/tediousjs/tedious/pull/1494/commits/76b371be4adb106b3c0b6802758a412cb90a4380 should fix this

arthurschreiber:
Thanks! ❤️

PR #1496, opened 2022-11-09
refactor: easier commitlint command

dhensby:
This creates a slightly easier to use commitlint command for contributors to use.

Instead of `node_modules/.bin/commitlint --from origin/master --to HEAD` contributors can now use `npm run lint:commit`. PR template updated as well


dhensby:
ping @arthurschreiber 😇

PR #1499, opened 2022-11-18
Non-buffering stream parsing

arthurschreiber:
/cc @mShan0


arthurschreiber:
@mShan0 @MichaelSun90 Here's what I put together so far. I'll stop continue working on this now, so you can pick this up. 😬 

I'd suggest working on token parsers in the following order:

* `src/token/returnstatus-token-parser.ts`: This should be quite easy to convert.
* `src/token/loginack-token-parser.ts`: This should be fairly straightforward to convert, see `src/token/infoerror-token-parser.ts` for inspiration.
* `src/token/fedauth-info-parser.ts`: This one might be a bit more involved. First step is to just get the full buffer for the token (using `LVarbyte`). And then use `Map` to convert the read buffer into a `FedAuthInfoToken` object using the same logic as is already used in that file.
* Similar to `src/token/infoerror-token-parser.ts`, see `src/parser/tokens/env-change-token.ts` for a partial implementation of this. Move the contents of `src/parser/tokens/env-change-token.ts` over into `src/token/infoerror-token-parser.ts`.

Once you have those tackled, we can discuss how to get the more complicated token parsing pieces converted as well.

arthurschreiber:
I had some time in the last 2 weeks to think about this more, and I think this is overcomplicating the whole parsing code without bringing as much benefit as I hoped.

When I played around with the idea of this parsing code, I was under the impression it would be best if the parsing code can be paused and resumed at any point when we need to wait for more data. unfortunately, this comes at quite some high cost in terms of code complexity, and eventually performance.

What I missed is that for most tokens, we don't need the ability to resume parsing at any point. There's three types of tokens in TDS:

* Fixed length tokens (like `RETURNSTATUS`, `DONE`, `DONEINPROC`, `DONEPROC`): These tokens have the same length always (with slight variations depending on the TDS protocol version). We can always check upfront whether the tokens can be parsed using the data we have received so far.
* Length-prefixed tokens (like `ENVCHANGE`, `FEDAUTHINFO`, `INFO`, `ERROR`): These tokens are prefixed with the length of their data. In these cases, we only need to check if we can read the length prefix, and then we can check if there's enough data buffered to read the whole content of the token. Only if either of these two steps fail because of not enough data buffered, we need to wait for the next chunks of the incoming message to arrive.
* Dynamic length tokens (like `ROW`, `COLMETADATA`, `FEATUREEXTACK`): These tokens have a dynamic length that is not necessarily known up front. These are the tokens where hitting the end of the buffered data can be potentially expensive, if we start the parsing of the token from the beginning whenever we run out of data.

Another thing I missed is the fact that the messages send by the server consist of packets, and we always read a full packet. A packet can be between 512 and 32767 bytes (with the default being 4096 bytes, and the max on TLS connections being 16383). This means even at the lowest setting, data will always come into the parser chunks of 512 bytes in the worst case.

So the number of times where we will actually hit a packet boundary and will need to wait for more data will be "relatively" low - low enough in fact that it's not really justified to be able to suspend the parsing at any time. It's much cheaper to just try parsing the few bits we need to figure out the total length of the token, and fail with an exception that specifies how much data was missing if we don't have enough data, and then retry once the next chunk / enough data to parse the packet has come in.

arthurschreiber:
The only more complex cases become the dynamic length tokens, but they are low enough in their number that I believe we can handle them as well.

For `COLMETADATA` tokens, instead of retrying to parse the whole token whenever we run out of data, we need to have that parsing logic be able to resume between the individual column metadata entries. Same for `ROW` tokens, where we need to parse the data column by column (instead of retrying the whole token). If we can't parse a column or value, it's fine to fail and retry when the next chunk comes in, because in all likelihood the next chunk will have enough data to finish the current column or value.

PR #1517, opened 2023-01-31
fix: fix type check for `options.encrypt`

mShan0:


src/connection.ts
```diff
@@ -1555,12 +1555,9 @@ class Connection extends EventEmitter {
             }
           }
         }
-        else if(typeof config.options.encrypt == 'boolean' && !config.options.encrypt)
-        {
-          this.config.options.encrypt = 'optional';
-        }
-        else
-        {
```
MichaelSun90, line 1563: Hi Mark, I think what I did here is actually somehow similar to yours. I should add a comment or something to explain this. Now I did feel this part is a bit unclear. The options.encrypt supposes to be default to 'Mandatory 'here. That is why I only check if encrypt is set to false, then set the options.encrypt to 'optional'.

MichaelSun90: Actually Mark is right. This will throw when config.options.encrypt = true. The "!config.options.encrypt" should happen within the else if. Then it will be correct and clear.  The if, else if, and else handles the type check for sting , boolean and throws an error if the type are not matching. Inside them, we do the value checks.

PR #1520, opened 2023-02-06
docs: update readme to include reference to newer SQLServer versions

arthurschreiber:


MichaelSun90:
Thanks Arthur for the PR! I only check out and reply the user who has the question for SQL server support for TDS 7_4 yesterday. Update the readme definitely is more clear for users to check out this. 

The linting tests are failing, we may need to force merge this.

PR #1522, opened 2023-02-08
feat: add TDS8.0 Support for tedious

MichaelSun90:
This PR added logics for supporting TDS8.0. 
Co-authored with @arthurschreiber 
Co-authored with @mShan0


.github/workflows/nodejs.yml
```diff
@@ -193,8 +193,7 @@ jobs:
         Push-Location C:\temp
 
         Invoke-WebRequest -Uri https://download.microsoft.com/download/c/c/9/cc9c6797-383c-4b24-8920-dc057c1de9d3/SQL2022-SSEI-Dev.exe -OutFile SQL2022-SSEI-Dev.exe
-        ./SQL2022-SSEI-Dev.exe /ACTION=Download /MEDIAPATH=C:\temp /MEDIATYPE=CAB /QUIET
-
+        Start-Process -Wait -FilePath ./SQL2022-SSEI-Dev.exe -ArgumentList /ACTION=Download, /MEDIAPATH=C:\temp, /MEDIATYPE=CAB, /QUIET
         Start-Process -Wait -FilePath ./SQLServer2022-DEV-x64-ENU.exe -ArgumentList /qs, /x:setup
 
         .\setup\setup.exe /q /ACTION=Install /INSTANCENAME=MSSQLSERVER /FEATURES=SQLEngine /UPDATEENABLED=0 /SQLSVCACCOUNT='NT SERVICE\MSSQLSERVER' /SQLSYSADMINACCOUNTS='BUILTIN\ADMINISTRATORS' /TCPENABLED=1 /NPENABLED=0 /IACCEPTSQLSERVERLICENSETERMS /SQLCOLLATION=SQL_Latin1_General_CP1_CI_AS /USESQLRECOMMENDEDMEMORYLIMITS
```
arthurschreiber: TODO: We can use `/SECURITYMODE='SQL'` to enable mixed authentication on installation, and use `/SAPWD='...'` to set the password. That would allow us to skip changing these settings (and the server restart).

src/connection.ts
```diff
@@ -1967,21 +1981,52 @@ class Connection extends EventEmitter {
...
+        const secureContext = tls.createSecureContext(this.secureContextOptions);
+        if (this.config.options.encrypt === 'strict') {
+          const encryptOptions = {
+            host: this.config.server,
+            socket: socket,
+            ALPNProtocols: ['tds/8.0'],
+            secureContext: secureContext,
+            servername: this.config.options.serverName ? this.config.options.serverName : this.config.server,
+          };
+          const encryptsocket = tls.connect(encryptOptions, () => {
```
arthurschreiber, line 1993: What will happen if someone connects using `strict` encryption to a SQL Server instance that does not support it?
How do we need to handle that, and how do we signal to the user that something is wrong? How do we help them understand _what_ is wrong?

src/connection.ts
```diff
@@ -1967,21 +1981,52 @@ class Connection extends EventEmitter {
...
+            this.socket = socket;
+
+            this.closed = false;
+            this.debug.log('connected to ' + this.config.server + ':' + this.config.options.port);
+
+            this.sendPreLogin();
+            this.transitionTo(this.STATE.SENT_PRELOGIN);
+          });
+
+          console.log('using TDS 8.0 strict TLS encryption');
```
arthurschreiber, line 2012: I believe this might be some leftover log output.

src/connection.ts
```diff
@@ -1967,21 +1981,52 @@ class Connection extends EventEmitter {
...
 
-        this.closed = false;
-        this.debug.log('connected to ' + this.config.server + ':' + this.config.options.port);
+          this.closed = false;
+          this.debug.log('connected to ' + this.config.server + ':' + this.config.options.port);
 
-        this.sendPreLogin();
-        this.transitionTo(this.STATE.SENT_PRELOGIN);
+          this.sendPreLogin();
+          this.transitionTo(this.STATE.SENT_PRELOGIN);
```
arthurschreiber, line 2028: Can we remove this duplication?

src/connection.ts
MichaelSun90: If user uses a strict to connect to a SQL server instance that does not support it, a socket error will be returned as : " Client network socket disconnected before secure TLS connection was established" 
Server side message: 
"The login packet used to open the connection is structurally invalid; the connection has been closed. Please contact the vendor of the client library. [CLIENT: 172.28.192.1]
"
 what do you think that we consume this message internally, and throw a custom error with a bit more information, so we can guide user through this

PR #1540, opened 2023-04-20
feat: add connector factory method

ruyadorno:
This changeset adds a new `connector` config option that may be used to define a custom socket factory method. Providing a much more flexible control of the socket connection creation.

Defining a custom `connector` config value allows **Tedious** to support a larger variety of environments/setups such as proxy servers using secure socket connections that are used by cloud providers such as GCP.

Linked below, the `pg` driver for PostgreSQL and the `mysql2` driver for MySQL are prior art example of this pattern. Also linked below is the Cloud SQL Node.js Connector, which demonstrates how third-party libraries can leverage the custom socket factory method.

**Refs:** https://github.com/sidorares/node-mysql2/blob/ba15fe25703665e516ab0a23af8d828d1473b8c3/lib/connection.js#L63-L65
**Refs:** https://github.com/brianc/node-postgres/blob/b357e1884ad25b23a4ab034b443ddfc8c8261951/packages/pg/lib/connection.js#L20
**Refs:** https://github.com/GoogleCloudPlatform/cloud-sql-nodejs-connector
**Signed-off-by:** Ruy Adorno <ruyadorno@google.com>

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -343,6 +343,7 @@ export interface InternalConnectionOptions {
   columnEncryptionSetting: boolean;
   columnNameReplacer: undefined | ((colName: string, index: number, metadata: Metadata) => string);
   connectionRetryInterval: number;
+  connector: undefined | (() => Promise<Socket>);
   connectTimeout: number;
   connectionIsolationLevel: typeof ISOLATION_LEVEL[keyof typeof ISOLATION_LEVEL];
   cryptoCredentialsDetails: SecureContextOptions;
@@ -534,6 +535,13 @@ export interface ConnectionOptions {
    * (default:`500`)
    */
   connectionRetryInterval?: number;
+
+  /**
+   * Custom connector factory method.
+   *
+   * (default: `undefined`)
+   */
+  connector?: () => Promise<Socket>;
 
   /**
    * The number of milliseconds before the attempt to connect is considered failed
@@ -1222,6 +1230,7 @@ class Connection extends EventEmitter {
         columnNameReplacer: undefined,
         connectionRetryInterval: DEFAULT_CONNECT_RETRY_INTERVAL,
         connectTimeout: DEFAULT_CONNECT_TIMEOUT,
+        connector: undefined,
         connectionIsolationLevel: ISOLATION_LEVEL.READ_COMMITTED,
         cryptoCredentialsDetails: {},
         database: undefined,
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -535,6 +536,13 @@ export interface ConnectionOptions {
    */
   connectionRetryInterval?: number;
 
+  /**
+   * Custom connector factory method.
+   *
+   * (default: `undefined`)
+   */
+  connector?: () => Promise<import('net').Socket>;
```
arthurschreiber, line 544: I think we import `Socket` from `net` somewhere further up, so you can just change the type to `Promise<Socket>`. What do you think?

arthurschreiber:
The changes look great, just a tiny nitpick. ❤️ 

How does this interact with encrypted connections? I assume the "Cloud SQL Node.js Connector" proxies via a TLS connection, and then tedious would open another TLS connection over that proxy connection?

Do you know what the perf impact of that setup would be? This doesn't sound _great_, but as it's only specific to GCP cloud connections, it's definitely no blocker. Just trying to understand if that's something we should document to make people aware of.

Thanks!

src/connection.ts
ruyadorno: right! that makes sense! let me change it 😊

ruyadorno:
> How does this interact with encrypted connections? I assume the "Cloud SQL Node.js Connector" proxies via a TLS connection, and then tedious would open another TLS connection over that proxy connection?

For GCP users using the **Cloud SQL Node.js Connector** the idea is to also set `options.ecrypt=false` (along with the connection factory method) by default in order to skip the driver encryption, so at the end of the day GCP users are not going to be running a doubly-encrypted communication system.

Every connection open by the factory method provided by the **Cloud SQL Node.js Connector** will use a TLS socket that uses an ephemeral certificate and keys that are rotated continuously (every hour or so). Essentially moving the encryption layer up a level so that our custom proxy handles the rotating certs logic.

arthurschreiber:
Oh, I just thought of one more thing. The custom connector should eventually be able to take an AbortSignal so that we can properly forward hitting the connection timeout.

I think the PR can properly be changed as-is for now, but eventually that should be added. What do you think?

arthurschreiber (approved):

arthurschreiber:
Thanks for your contribution! ❤️

PR #1542, opened 2023-05-30
feat: skip config validation when using connector

### LINKED ISSUE #1541: `server` and `port` properties are required when using the custom `connector` factory method
**Software versions**
* Tedious: 16.1.0
* Node.js: 16+

**Additional Libraries Used and Versions**

Particularly noticeable when using alongside the `@google-cloud/cloud-sql-connector` library that uses the `connector` factory method.

**Connection configuration**

Notice the highlighted properties that should not be required, since a custom connector function is being used:

```patch
-  server: '0.0.0.0',
  authentication: {
    type: 'default',
    options: {
      userName: process.env.SQLSERVER_USER,
      password: process.env.SQLSERVER_PASS,
    },
  },
  options: {
    connector: async () => net.connect({
      host: '192.168.1.212',
      port: 1433,
    }),
-    port: 9999,
    database: process.env.SQLSERVER_DB,
  },
})
```

**Problem description**

When using the new custom `connector` factory method, the `server` and `port` properties should not be required.

**Expected behavior**

Config options defining only a `connector` factory method should be valid.

**Actual behavior**

**Error message/stack trace**

```
The "config.server" property is required and must be of type string.

  test: open connection and run basic sqlserver commands
  stack: |
    new Connection (node_modules/tedious/src/connection.ts:1056:13)
    Test.<anonymous> (system-test/tedious-connect.cjs:30:24)
  at:
    line: 1056
    column: 13
    file: node_modules/tedious/src/connection.ts
    function: Connection
```

**Any other details that can be helpful**

Please note that I'm the original contributor of that feature and will be happy to follow up with a fix for this problem 😊

ruyadorno:
This changeset fixes a problem in the original custom `connector` factory method implementation that required the `server` config property to be defined even though its value is ignored.

Removing the validation when `config.options.connector` is defined fixes the problem.

Ref: https://github.com/tediousjs/tedious/pull/1540
Fixes: https://github.com/tediousjs/tedious/issues/1541

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -331,7 +331,7 @@ interface ErrorWithCode extends Error {
 }
 
 interface InternalConnectionConfig {
-  server: string;
+  server: undefined | string;
   authentication: DefaultAuthentication | NtlmAuthentication | AzureActiveDirectoryPasswordAuthentication | AzureActiveDirectoryMsiAppServiceAuthentication | AzureActiveDirectoryMsiVmAuthentication | AzureActiveDirectoryAccessTokenAuthentication | AzureActiveDirectoryServicePrincipalSecret | AzureActiveDirectoryDefaultAuthentication;
   options: InternalConnectionOptions;
 }
@@ -1061,7 +1061,7 @@ class Connection extends EventEmitter {
       throw new TypeError('The "config" argument is required and must be of type Object.');
     }
 
-    if (typeof config.server !== 'string') {
+    if (typeof config.server !== 'string' && !config.options!.connector) {
       throw new TypeError('The "config.server" property is required and must be of type string.');
     }
 
@@ -1352,8 +1352,15 @@ class Connection extends EventEmitter {
         if (typeof config.options.connector !== 'function') {
           throw new TypeError('The "config.options.connector" property must be a function.');
         }
+        if (config.server) {
+          throw new Error('Server and connector are mutually exclusive, but ' + config.server + ' and a connector function were provided');
+        }
+        if (config.options.port) {
+          throw new Error('Port and connector are mutually exclusive, but ' + config.options.port + ' and a connector function were provided');
+        }
 
         this.config.options.connector = config.options.connector;
+        this.config.options.port = undefined;
       }
 
       if (config.options.cryptoCredentialsDetails !== undefined) {
@@ -1917,7 +1924,10 @@ class Connection extends EventEmitter {
   initialiseConnection() {
     const signal = this.createConnectTimer();
 
-    if (this.config.options.port) {
+    if (this.config.options.connector) {
+      // port and multiSubnetFailover are not used when using a custom connector
+      return this.connectOnPort(0, false, signal, this.config.options.connector);
+    } else if (this.config.options.port) {
       return this.connectOnPort(this.config.options.port, this.config.options.multiSubnetFailover, signal, this.config.options.connector);
     } else {
       return instanceLookup({
@@ -1990,7 +2000,7 @@ class Connection extends EventEmitter {
     return new TokenStreamParser(message, this.debug, handler, this.config.options);
   }
 
-  socketHandlingForSendPreLogin(socket: net.Socket) {
+  socketHandlingForSendPreLogin(socket: net.Socket, customConnector: boolean) {
     socket.on('error', (error) => { this.socketError(error); });
     socket.on('close', () => { this.socketClose(); });
     socket.on('end', () => { this.socketEnd(); });
@@ -2002,7 +2012,11 @@ class Connection extends EventEmitter {
     this.socket = socket;
 
     this.closed = false;
-    this.debug.log('connected to ' + this.config.server + ':' + this.config.options.port);
+    const message =
+      'connected to ' + this.config.server + ':' + this.config.options.port;
+    const customConnectorMessage =
+      'connected via custom connector';
+    this.debug.log(customConnector ? customConnectorMessage : message);
 
     this.sendPreLogin();
     this.transitionTo(this.STATE.SENT_PRELOGIN);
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -1052,7 +1052,7 @@ class Connection extends EventEmitter {
       throw new TypeError('The "config" argument is required and must be of type Object.');
     }
 
-    if (typeof config.server !== 'string') {
+    if (typeof config.server !== 'string' && !config.options!.connector) {
       throw new TypeError('The "config.server" property is required and must be of type string.');
```
MichaelSun90, line 1056: We should change the error message since this is not just verify server property and it is no long required.
perhaps we can change it to something looks like : "The "config.server" property is required and must be of type string if "config.options.connector" is not used "

@arthurschreiber Do you think this message looks ok?

src/connection.ts
```diff
@@ -1902,6 +1902,9 @@ class Connection extends EventEmitter {
 
     if (this.config.options.port) {
       return this.connectOnPort(this.config.options.port, this.config.options.multiSubnetFailover, signal, this.config.options.connector);
+    } else if (this.config.options.connector) {
+      // port and multiSubnetFailover are not used when using a custom connector
+      return this.connectOnPort(0, false, signal, this.config.options.connector);
```
MichaelSun90, line 1907: There is a possibility that user proved both port and connector from config, the logic will go into the if statement instead of the else if and still work properly since the connector is passed into connectOnPort. Another case is, if use only pass in connector and logic will got into the else if and still call the connectOnPort. 
Under both cases, with or without port passed into connectOnPort function, the connect socket here will be provided by the passed in customConnector, and the passed in port number should be ignored within connectOnPort.

arthurschreiber:
@ruyadorno I'm wondering how this is going to interact with all the code that expects `config.server` (and also `config.port`) to be set? There's the type checking that's supposed to make sure that `config.server` is set to an actual value, and here's a whole bunch of logging code that expects `config.server` to be set to a string (and `config.port` to be set as well as).

It would be nice if that all could be cleaned up to either be consistent, or at least not to lead to log messages containing `undefined` as the server name and a bogus port. 🤔

src/connection.ts
ruyadorno: thanks @MichaelSun90! these are also great examples of the kind of tests I can add to make codecov happy 😊

ruyadorno:
@arthurschreiber that's a good point! let me take a look at these `config.server` logging LOC  👀  

I might just try to clean them up as part of this PR then in case it looks actionable 😊

src/connection.ts
ruyadorno: I'm in favor of leaving the error message as is. My line of thought is that bringing up the `config.options.connector` property to users not using it might end up being more confusing.

src/connection.ts
ruyadorno: I added the connector check to take precedence just to err on the safe side but at the same time I made sure to set `port` to undefined during the validation phase of the constructor along with adding tests for the combinations of providing port/server/connector. Let me know if it looks better now 😊

ruyadorno:
@arthurschreiber I updated the PR with a pass at updating occurrences of many logging message that were using the `server:port` combo!

src/connection.ts
MichaelSun90: Agreed! Chatted with Arthur, and he thinks that we should keep the old message as well.

ruyadorno: awesome! 👍

ruyadorno:
Went through a few more iterations to improve code coverage and cleaned a few things up, should be good to go now 😊 lmk!

ruyadorno:
hi @MichaelSun90! Added your suggestion to update the type definition for the `server` config (from c72786cbb19034545a475ae80af6346e99ec5852) along with a few type conversions for TS lint and rebased on top of the latest commits from **master**.

Let me know if you think this is in a good shape to land now 😊 I'm looking forward to finally have the `connector` config usage cleaned up!

src/instance-lookup.ts
```diff
@@ -14,7 +14,7 @@ const MYSTERY_HEADER_LENGTH = 3;
 type LookupFunction = (hostname: string, options: dns.LookupAllOptions, callback: (err: NodeJS.ErrnoException | null, addresses: dns.LookupAddress[]) => void) => void;
 
 // Most of the functionality has been determined from from jTDS's MSSqlServerInfo class.
-export async function instanceLookup(options: { server: string, instanceName: string, timeout?: number, retries?: number, port?: number, lookup?: LookupFunction, signal: AbortSignal }) {
+export async function instanceLookup(options: { server: undefined | string, instanceName: string, timeout?: number, retries?: number, port?: number, lookup?: LookupFunction, signal: AbortSignal }) {
```
arthurschreiber, line 17: @ruyadorno I think this change is not correct. Logically, the `instanceLookup` can not work without a hostname, so changing the type to allow `undefined` is not correct. Also, casting `options.server` to a string using `String` further down is also not right - if someone would call this method with `undefined`, it would try to send udp messages to the hostname `"undefined"`. 😅 

I think it's better to change the location where this method is called instead (will leave a separate comment for that).

src/connection.ts
```diff
@@ -1917,7 +1924,10 @@ class Connection extends EventEmitter {
...
     const signal = this.createConnectTimer();
 
-    if (this.config.options.port) {
+    if (this.config.options.connector) {
+      // port and multiSubnetFailover are not used when using a custom connector
+      return this.connectOnPort(0, false, signal, this.config.options.connector);
+    } else if (this.config.options.port) {
       return this.connectOnPort(this.config.options.port, this.config.options.multiSubnetFailover, signal, this.config.options.connector);
     } else {
       return instanceLookup({
```
arthurschreiber, line 1933: I think this is the place you need to change. We know here that `this.config.server` is not `undefined` in this `else` block (even if the type is declared as `string | undefined`), so we can tell TypeScript that this is the case by changing `this.config.server` to `this.config.server!`.

PR #1559, opened 2023-07-24
docs: update `clientId` explanation

mShan0:


api-connection.html
```diff
@@ -138,8 +138,10 @@ <h2 id="function_newConnection">new Connection(config)</h2>
           <ul>
             <li>username (string): A user needs to provide the `userName` associated to their account.</li>
             <li>password (string): A user needs to provide the `password` associated to their account.</li>
-            <li>clientId (string): Application (`client`) ID from your registered Azure application. This currently is optional
-              and uses a default clientId, but a user-provided clientId will be required in the near future.
+            <li>clientId (string): Application (`client`) ID from your registered Azure application. 
+              The application has to have public client flows enabled and has to have `user_impersonation` 
+              permissions for Azure SQL Database. The user account that is being used also has to be assigned
+              to the application, and has to grant consent (either explicitly or via admin consent).
             </li>
```
MichaelSun90, line 145: I also remember David mentioned that our pervious hard coded clientID will be expired, so if there are users saved and keep using that client id, it won't work in the future. They have to grad their own clientID and use that for connections. We should also mention this as well.

frequently-encountered-problems.html
```diff
@@ -75,6 +75,17 @@ <h3>Tedious Configurations</h3>
 
     </dd>
     <br />
+    <dt>
+        <h2>Providing a client ID when using <code>azure-active-directory-password</code></h2>
+    </dt>
+    <dd>
+        <p>The clientId field is now mandatory when using the <code>azure-active-directory-password</code> 
+            authentication type since the provided default client ID has been removed from Tedious. 
+            That default client ID will expire in the near future and stop working.</p>
```
MichaelSun90, line 84: can we add a quick explaination why this is removed? something like "for other Microsoft owned first party driver, a hard coded client id is provided within the driver. Since tedious is not a Microsoft owner first party driver, we have to remove the hard coded client id and let user provide their own. "

MichaelSun90 (approved):

Issue #1558, opened 2023-07-22
Network socket not correctly cleaned up when connection timeout fires during TDS8.0 TLS negotiation


arthurschreiber:
The recently added `Connection#wrapWithTls` method does not behave correctly if the connection timeout fires during the TLS negotiation.

When the connect timeout fires, we dispatch a [`ConnectionError` to the `connect` event](https://github.com/tediousjs/tedious/blob/b78df149fd55d11125a35af7e2366c0a36ed4c74/src/connection.ts#L2132-L2143) and move the internal state over to the `FINAL` state. Because we haven't fully connected yet, the network socket does not get cleaned up when moving to the `FINAL` state.

When the connection timeout fires between the call to [`connect`](https://github.com/tediousjs/tedious/blob/b78df149fd55d11125a35af7e2366c0a36ed4c74/src/connection.ts#L2045C7-L2045C67) and before [`wrapWithTls`](https://github.com/tediousjs/tedious/blob/b78df149fd55d11125a35af7e2366c0a36ed4c74/src/connection.ts#L2050C11-L2050C51) returns, we never clean up the socket correctly, because `wrapWithTls` doesn't return an error if the connection timeout fires.

I think this can be fixed by making `wrapWithTls` take the `signal` passed to `connectOnPort`, and handle it correctly (i.e. call `signal.throwIfError` at the beginning of the function, and set the correct event handlers to reject the promise if the signal is aborted).

MichaelSun90:
Hi @arthurschreiberm, I will look into this 👍

PR #1564, opened 2023-07-28
fix: handle timeouts during tls negotiation for `strict` encryption

MichaelSun90:
Network socket not correctly cleaned up when connection timeout fires during TDS8.0 TLS negotiation #1558

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2089,7 +2089,7 @@ class Connection extends EventEmitter {
       this.socketHandlingForSendPreLogin(socket);
     })().catch((err) => {
       this.clearConnectTimer();
-      if (err.name === 'AbortError' || err.message === 'AbortError') {
+      if (err === signal.reason) {
         return;
       }
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1931,7 +1931,8 @@ class Connection extends EventEmitter {
         });
       }, (err) => {
         this.clearConnectTimer();
-        if (err.name === 'AbortError') {
+
+        if (signal.aborted) {
           // Ignore the AbortError for now, this is still handled by the connectTimer firing
           return;
         }
@@ -2089,7 +2090,8 @@ class Connection extends EventEmitter {
       this.socketHandlingForSendPreLogin(socket);
     })().catch((err) => {
       this.clearConnectTimer();
-      if (err === signal.reason) {
+
+      if (signal.aborted) {
         return;
       }
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -5,7 +5,7 @@ import * as net from 'net';
 import dns from 'dns';
 
 import constants from 'constants';
-import { SecureContextOptions } from 'tls';
+import { type SecureContextOptions } from 'tls';
 
 import { Readable } from 'stream';
 
@@ -16,7 +16,7 @@ import {
   UsernamePasswordCredential,
 } from '@azure/identity';
 
-import BulkLoad, { Options as BulkLoadOptions, Callback as BulkLoadCallback } from './bulk-load';
+import BulkLoad, { type Options as BulkLoadOptions, type Callback as BulkLoadCallback } from './bulk-load';
 import Debug from './debug';
 import { EventEmitter, once } from 'events';
 import { instanceLookup } from './instance-lookup';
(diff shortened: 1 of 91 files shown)


MichaelSun90:
@arthurschreiber remove the TDS 7_1 checks against SQL server 2022 since 7_1 suppose not work with SQL server 2022 and it did failing the encrypt tests. We did no run into this because the added encrypt tests are not actually running due to a error within our test. If we still want to keep the 7_1 test run against SQL server 2022 for certain reason, we can also add some logic to skip the tests like that we did in 'should reset Connection'.

src/connection.ts
```diff
@@ -2008,7 +2009,11 @@ class Connection extends EventEmitter {
     this.transitionTo(this.STATE.SENT_PRELOGIN);
   }
 
-  wrapWithTls(socket: net.Socket): Promise<tls.TLSSocket> {
+  wrapWithTls(socket: net.Socket, signal: AbortSignal): Promise<tls.TLSSocket> {
+    if (signal.aborted) {
+      throw new AbortError();
+    }
```
arthurschreiber, line 2015:
```
signal.throwIfAborted();
```

src/connection.ts
```diff
@@ -2023,12 +2028,41 @@ class Connection extends EventEmitter {
...
-        encryptsocket.removeListener('error', reject);
+      const encryptsocket = tls.connect(encryptOptions);
+
+      const onAbort = () => {
+        encryptsocket.removeListener('error', onError);
+        encryptsocket.removeListener('connect', onConnect);
+
+        encryptsocket.destroy();
+
+        reject(new AbortError());
```
arthurschreiber, line 2039:
```
reject(signal.reason);
```

src/connection.ts
```diff
@@ -41,6 +41,7 @@ import { createNTLMRequest } from './ntlm';
 import { ColumnEncryptionAzureKeyVaultProvider } from './always-encrypted/keystore-provider-azure-key-vault';
 
 import { AbortController, AbortSignal } from 'node-abort-controller';
+import AbortError from './errors/abort-error';
```
arthurschreiber, line 44: You might be able to remove this import if you remove the cases where you construct `AbortError` instances.

MichaelSun90:
Figured out, that signal.reason can be undefined which causing an additional error if we do tostring and compare it with the error that has been caught see if it include the abort title. However, just err === signal.reason still does not handles the case that :
signal.reason :
```
Error: AbortError
    at AbortController.abort (C:\Users\msun\Documents\GitHub\tedious\node_modules\node-abort-controller\index.js:56:31)
    at Timeout._onTimeout (C:\Users\msun\Documents\GitHub\tedious\src\/connection.ts:2116:18)
    at listOnTimeout (node:internal/timers:569:17)
    at processTimers (node:internal/timers:512:7)
```
err is:
```
AbortError: The operation was aborted
    at EventEmitter.onAbort (C:\Users\msun\Documents\GitHub\tedious\src\/connector.ts:108:18)
    at EventEmitter.emit (node:events:514:28)
    at AbortSignal.dispatchEvent (C:\Users\msun\Documents\GitHub\tedious\node_modules\node-abort-controller\index.js:28:23)
    at AbortController.abort (C:\Users\msun\Documents\GitHub\tedious\node_modules\node-abort-controller\index.js:58:17)
    at Timeout._onTimeout (C:\Users\msun\Documents\GitHub\tedious\src\/connection.ts:2116:18)
    at listOnTimeout (node:internal/timers:569:17)
    at processTimers (node:internal/timers:512:7) {
  code: 'ABORT_ERR'
}
```

```
if (err === signal.reason || err.name === 'AbortError'){
    return;
}
```
 ```
if(signal.reason && signal.reason.tostring().include('AbortError')) {
      return;
 }
```
both of these will handle the errors correctly, but not exactly a clean solution in my opinion.
Not sure if we should go with either one or @arthurschreiber  you have suggestions for handling this.

arthurschreiber:
@MichaelSun90 I changed the code to check for `signal.aborted` instead, and I think this now has the correct behaviour.

If the `signal` (which is coming from the `connectTimeout`) was aborted, that means that the connect timeout has fired and we need to ignore that error (because a different error will be generated by the connect timeout code.

Anyway, hopefully this can be cleaned up via the changes in https://github.com/tediousjs/tedious/pull/1562.

MichaelSun90:
> @MichaelSun90 I changed the code to check for `signal.aborted` instead, and I think this now has the correct behaviour.
> 
> If the `signal` (which is coming from the `connectTimeout`) was aborted, that means that the connect timeout has fired and we need to ignore that error (because a different error will be generated by the connect timeout code.
> 
> Anyway, hopefully this can be cleaned up via the changes in #1562.

Thanks for help on this change, and release it ! :bowing_man:

PR #1566, opened 2023-08-04
feat: add new error type for param validation

### LINKED ISSUE #1557: Minor addition to exception information on failed bulk-load data due to type validation: Mention the failed column and errorneous value
**Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it**
If I do a bulk-load of many columns and many values, if any of the values fail validation, the corresponding exception does not mention the concerned column name and neither the erroneous value. This makes it unnecessarily difficult to narrow down the wrong values.

**Describe the preferred solution**
In https://github.com/tediousjs/tedious/blob/master/src/bulk-load.ts#L188 please add the `c.name` and also the `value` to the exception object that is returned. For example by simply adding more properties:

> error.column = c.name
> error.wrongValue = value

Other properties could be added instead as well.

**Describe alternatives you've considered**
hm... the exception here is pretty much exactly the normal error handling. It is simply missing the information about the place where the exception occurred.

MichaelSun90:
This should resolve #1557
create a new Error subclass (maybe called ParameterValidationError), have it inherit from whatever error class we currently use (that might be TypeError, but I'm not sure), and define the additional properties on that.

User will have access to the param name and param value information via both message and error properties.

### DIFF
--- a/src/bulk-load.ts
+++ b/src/bulk-load.ts
@@ -1,6 +1,7 @@
 import { EventEmitter } from 'events';
 import WritableTrackingBuffer from './tracking-buffer/writable-tracking-buffer';
 import Connection, { InternalConnectionOptions } from './connection';
+import { ParameterValidationError } from './errors';
 
 import { Transform } from 'stream';
 import { TYPE as TOKEN_TYPE } from './token/token';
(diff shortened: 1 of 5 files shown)


MichaelSun90:
@arthurschreiber , do we need to worry about this new parameter validation error will review some sensitive information that user may insert within their column, since we are expose the parameter values via this error?

MichaelSun90:
Create a new PR for test these error related, then use the test test against this current changes within this PR.

PR #1573, opened 2023-09-14
ci: improve github actions

dhensby:
1. The setup-node action automatically stores caches for NPM and uses the same logic as the current action implementation, so there's no need to duplicate it in the action definition.
2. Remove redundant NPM upgrade step - v6.x of Node is no longer tested/supported.
3. Linting step does not need complete repository history.
4. Use [setup-sqlserver](https://github.com/tediousjs/setup-sqlserver) action to install SQL Server

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -126,59 +126,10 @@ jobs:
 
     steps:
     - name: Install SQL Server ${{ matrix.mssql-version }}
-      shell: pwsh
-      run: |
-        Push-Location C:\temp
-
-        $exe_link, $box_link, $update_link, $update_download_link = '', '', '', ''
-
-        switch (${{ matrix.mssql-version }}) {
-          '2022' {
-            $exe_link = 'https://download.microsoft.com/download/3/8/d/38de7036-2433-4207-8eae-06e247e17b25/SQLServer2022-DEV-x64-ENU.exe'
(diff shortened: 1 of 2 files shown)


arthurschreiber:
@dhensby Would you mind resolving the conflicts with the latest changes that happened on `master`?

dhensby:
@arthurschreiber done.

I had also wanted to replace lots of the repeated SQL Server installer code with the [setup-sqlserver](https://github.com/tediousjs/setup-sqlserver) action, but I'm not sure it's an entirely 1-to-1 replacement

arthurschreiber:
@dhensby It'd be great if `setup-sqlserver` would also fetch the latest updates - otherwise sql server 2022 is broken (TDS protocol version 8 does not work correctly).

dhensby:
@arthurschreiber - working on it: https://github.com/tediousjs/setup-sqlserver/pull/45

arthurschreiber:
I never quite understood why this didn't work out of the box. The installer has an option (`UpdateEnabled`) which defaults to true, and the default update source is `MU` (= Microsoft Update). 🤷

dhensby:
I stopped trying to understand the SQL installer a long time ago! 😬

dhensby:
@arthurschreiber looks like it's working, one of the tests hit a timeout and needs restarting

dhensby:
🎉 we are green

PR #1580, opened 2023-10-05
feat: improve performance of response data parsing

arthurschreiber:
This replaces the current token parsing code (including value and metadata parsers) to use the "try parsing a single token, fail if not enough data is buffered, retry once more data comes in" approach described over in https://github.com/tediousjs/tedious/pull/1499#issuecomment-1742813365.

### TODOs

* [x] Fix CI failures 😬 Seems like there's some tokens that are currently not parsed correctly?
* [x] Remove remaining references to `parser.streamBuffer.waitForData`
* [x] Implement `NBCROW` token parsing.
* [x] Change `buf: Buffer | BufferList` type to only `buf: Buffer`, I don't think we ever pass a `BufferList` object here.
* [x] Change `Result` to be a proper class? So we can do `return Result(value, offset)`. Probably should be benchmarked to see if it makes any difference.

### Results

On benchmarks that measure the full request-response cycle (like the `select-many-rows.js` benchmark), we can see an improvement ranging between a few percent up to 50% increase in iterations per second. Actual numbers will vary depending on various variables.

Here are example benchmark results from my machine:

<details>

<summary>Node.js v16.20.0 <code>master</code> branch</summary>

```
query/select-many-rows.js size=10 n=10: 732.5808750971586 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=10: 236.7257682532375 (minor: 1 - 2.320625999942422ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=10: 108.46899089030772 (minor: 3 - 6.047306000255048ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=10: 23.63362124251892 (minor: 28 - 10.39379300083965ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=100: 943.124882268542 (minor: 1 - 0.5346679999493062ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 674.9636528698111 (minor: 7 - 3.505343999247998ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=100: 196.5643086662155 (minor: 30 - 17.99759300146252ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=100: 27.498503692295312 (minor: 271 - 105.15239100111648ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=1000: 1,498.2127961884485 (minor: 10 - 11.73080400051549ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 970.4751161722522 (minor: 42 - 28.6563880010508ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=1000: 240.3745889329216 (minor: 297 - 111.55920999590307ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=1000: 28.68140546790126 (minor: 2713 - 854.066908005625ms, major: 3 - 19.69343699933961ms, incremental: 6 - 6.423661998938769ms)
```

</details>

<details>

<summary>Node.js v16.20.0 <code>arthur/rewrite-parser</code> branch</summary>

```
query/select-many-rows.js size=10 n=10: 851.4735516136872 (minor: 0 - 0ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=10: 350.286941050631 (minor: 1 - 2.049265999812633ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=10: 151.6876111358244 (minor: 2 - 4.578842999879271ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=10: 27.913029431453573 (minor: 16 - 8.864512000698596ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=100: 1,015.8547807959602 (minor: 1 - 1.1216849996708333ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=100: 830.9837475616963 (minor: 5 - 3.184196999296546ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=100: 279.6579015364296 (minor: 18 - 12.758315999992192ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=100: 39.79706595757135 (minor: 156 - 54.824087999761105ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10 n=1000: 1,543.7394101309656 (minor: 8 - 12.882265999447554ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=100 n=1000: 1,115.126761878539 (minor: 27 - 22.832545001525432ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=1000 n=1000: 329.5490264037709 (minor: 172 - 69.64875099854544ms, major: 0 - 0ms, incremental: 0 - 0ms)
query/select-many-rows.js size=10000 n=1000: 42.634000067428566 (minor: 1555 - 534.4554190100171ms, major: 2 - 10.8917609998025ms, incremental: 4 - 0.6896780002862215ms)
```

</details>


arthurschreiber:
@mShan0 @MichaelSun90 As discussed, here's the changes I have so far. Would you mind taking this over?

src/token/helpers.ts
```diff
@@ -0,0 +1,233 @@
+import type { BufferList } from 'bl/BufferList';
+
+export type Result<T> = {
+  value: T;
+  offset: number;
+};
+
+export class NotEnoughDataError extends Error {
```
arthurschreiber, line 8: I don't think this actually needs to inherit from `Error`, and I can imagine this being expensive when there's a lot of these errors (e.g. when packetsize is low). We can change this later to remove the inheritance, because we don't actually rely on things like the location of where this error is created being recorded.

test/integration/datatypes-in-results-test.ts
arthurschreiber, line 1: The changes in this file need to be reverted.

src/value-parser.ts
arthurschreiber, line 1: Here we need to do a bunch of cleanups, e.g. remove the old-style value reading functions, and rename the new style ones to to start with a `_`.

src/token/infoerror-token-parser.ts
```diff
@@ -12,43 +14,59 @@ interface TokenData {
...
-            });
-          });
-        });
-      });
-    });
-  });
+function readToken(buf: Buffer | BufferList, offset: number, options: ParserOptions): Result<TokenData> {
+  let tokenLength;
+  ({ offset, value: tokenLength } = readUInt16LE(buf, offset));
+
```
MichaelSun90, line 20: is the token length that we got here represent the size for entire info error token? 
if true, if we compare the buf.length with tokenLength +offset, will this always fail the if statement? 
under the connection retry test, we have a mock token size of 31 byte, and here when we call to readtoken, the offset is 3, the buffer length is 31, the logic never pass through here with the pervious compare if(bur.length()< tokenlength +offset)
I change it to (buf.length()+offset<tokenLength) means that if incoming buffer size plus the current start offset is less then the tokenlength that we expect, then we throw and wait for more. 🤔

src/value-parser.ts
```diff
@@ -5,711 +5,757 @@ import { TYPE } from './data-type';
...
-        });
+
+      let dataLength;
+      ({ offset, value: dataLength } = readUInt16LE(buf, offset));
+
+      if (dataLength === NULL) {
+        return new Result(null, offset);
       }
 
+      console.log(metadata, dataLength);
```
arthurschreiber, line 194: I think this log output line here is causing the tests to time out 😬

src/token/loginack-token-parser.ts
```diff
@@ -1,45 +1,61 @@
...
-            });
-          });
-        });
-      });
-    });
-  });
+  let tokenLength;
+  ({ offset, value: tokenLength } = readUInt16LE(buf, offset));
+
+  // mocha --inspect test/integration/connection-test.js
```
arthurschreiber, line 18:
```suggestion
```

src/token/stream-parser.ts
```diff
@@ -16,153 +16,392 @@ import returnValueParser from './returnvalue-token-parser';
...
-  set position(value) {
-    this.streamBuffer.position = value;
+    this.position = 0;
   }
 
   suspend(next: () => void) {
-    this.suspended = true;
-    this.next = next;
+
   }
```
arthurschreiber, line 405: @MichaelSun90 @mShan0 Most of the code from here up to the end of the class is unused and can be deleted now. 🙇‍♂️

test/unit/token/env-change-token-parser-test.js
```diff
@@ -52,7 +52,7 @@ describe('Env Change Token Parser', () => {
     assert.strictEqual(token.newValue, 2048);
   });
 
-  it('should be of bad type', async () => {
+  it.only('should be of bad type', async () => {
```
arthurschreiber, line 55:
```suggestion
  it('should be of bad type', async () => {
```

test/unit/token/row-token-parser-test.js
```diff
@@ -425,42 +425,6 @@ describe('Row Token Parser', () => {
...
-      await parser.next();
-    } catch (err) {
-      error = err;
-    }
-
-    assert.instanceOf(error, Error);
-    assert.strictEqual(error.message, 'Partially Length-prefixed Bytes unmatched lengths : expected 7, but got 6 bytes');
-    assert.isTrue((await parser.next()).done);
-  });
-
```
arthurschreiber, line 463: Can we re-implement the logic to handle this test again? I think I skipped this when I re-implemented the PLP stream parsing, but i think it should be fairly straightforward to bring back.

src/value-parser.ts
```diff
@@ -14,702 +15,764 @@ const PLP_NULL = Buffer.from([0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF]);
...
+
+  const type = parser.buffer.slice(parser.position, parser.position + 8);
+  parser.position += 8;
+
+  if (type.equals(PLP_NULL)) {
+    return null;
+  }
 
+  const chunks: Buffer[] = [];
   let offset = 0;
```
MichaelSun90, line 600: added  a offset var here for recording the read token length. We may able to handling it with some calculation with the paser.position but I think this maybe a bit more straight forward.

arthurschreiber: Adding this variable to keep track how much we've read makes sense, but maybe we can call it something different than `offset`?

src/value-parser.ts
```diff
@@ -14,702 +15,764 @@ const PLP_NULL = Buffer.from([0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF]);
...
-        chunks.push(chunk);
-        length += chunkLength;
+    if (high >= (2 << (53 - 32))) {
+      console.warn('Read UInt64LE > 53 bits : high=' + high + ', low=' + low);
+    }
 
-        next(done);
-      });
-    });
+    const expectedLength = low + (0x100000000 * high);
```
arthurschreiber, line 631: I know you copied this from the old version of this code, but I'd love if we could improve this.

This code is reading the lower 4 bytes and the higher 4 bytes of an 8 byte integer, to then try and put it back together as a 8 byte integer. I think instead of doing that, we can just read this as a bigint and then convert it back to a `Number`?

src/value-parser.ts
MichaelSun90: how about call it readLength wich match the expectedLength

MichaelSun90: changed it to currentLength instead

src/value-parser.ts
MichaelSun90: The part for checking length looks like this now. Not sure if we still want the expectedLength var since 
`if (!type.equals(UNKNOWN_PLP_LEN)) {
    const expectedLength = Number(readBigInt(type, 0).value);
    if (currentLength !== expectedLength) {
      throw new Error('Partially Length-prefixed Bytes unmatched lengths : expected ' + expectedLength + ', but got ' + currentLength + ' bytes');
    }
  }`

src/value-parser.ts
```diff
@@ -14,702 +15,757 @@ const PLP_NULL = Buffer.from([0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF]);
...
+    chunks.push(parser.buffer.slice(parser.position, parser.position + chunkLength));
+    parser.position += chunkLength;
+    currentLength += chunkLength;
   }
 
-  next(() => {
-    if (offset !== totalLength) {
-      throw new Error('Partially Length-prefixed Bytes unmatched lengths : expected ' + totalLength + ', but got ' + offset + ' bytes');
+  if (!type.equals(UNKNOWN_PLP_LEN)) {
+    const expectedLength = Number(readBigInt(type, 0).value);
```
arthurschreiber, line 624: The `type` is guaranteed to be a 8 byte `Buffer` here - I think we can call `type.readBigUInt64LE` directly on it, without having to go through any of the helper functions.

PR #1596, opened 2024-01-17
fix: use userland module for `punycode`

EliTheCoder:
The slash at the end is necessary to use the userland module. The module built-in to node is deprecated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "@types/lru-cache": "^5.1.1",
         "@types/mocha": "^10.0.1",
         "@types/node": "^16.18.38",
+        "@types/punycode": "^2.1.3",
         "@types/sprintf-js": "^1.1.2",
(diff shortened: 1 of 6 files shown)


MichaelSun90:
Hi @EliTheCoder, Thanks for raising this PR. I also tried to use the url.domainToASCII function to replace the one that used to be provided by punnyCode. Seems the related tests still behave the same way. Not sure whether we want still keep the punnyCode or just use url instead. 
Hi @arthurschreiber, seems url.domainToASCII can handle what punycode.toASCII used to do. Do we want still keep punnyCode or just go with the url lib?

MichaelSun90:
Just double checked that Node 14 seems do not support node:url but we should be safe since the minimum node version maintained on tedious side is node 16 :thinking: .

arthurschreiber:
Thanks for your contribution! ❤️

PR #1597, opened 2024-02-02
refactor: replace `punycode` with `node:url`

MichaelSun90:
Seems that native node:url supports the same functionality that tedious required. Replace punnyCode side function with url.domainToASCII.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11612,6 +11612,14 @@
       "resolved": "https://registry.npmjs.org/psl/-/psl-1.8.0.tgz",
       "integrity": "sha512-RIdOzyoavK+hA18OGGWDqUTsCLhtA7IcZ/6NCs4fFJaHBDab+pDDmDIByWFRQJq2Cd7r1OoQxBGKOaztq+hjIQ=="
     },
+    "node_modules/punycode": {
+      "version": "2.3.1",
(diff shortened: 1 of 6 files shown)


arthurschreiber:
Can you update the PR to also drop the dependency on the punycode module? 🙇‍♂️

WikiRik:
Hi there! Coming over here from sequelize, a database ORM which is using this package as the adapter for mssql.
This PR broke one of our tests. We were using 'http://wowow.example' as the host, which punycode kept intact (since there all characters were already ASCII). Node however returned an empty string since it is not a valid domain (due to the usage of 'http://'). Luckily that's an easy fix for us, but others might stumble over that. 

We did also get this DeprecationWarning from Node with that, because the hostname that it tried to lookup was empty. It might be nice to catch that in advance and throw an error that the host given is not a valid domain.
https://github.com/tediousjs/tedious/blob/ddfa397bf8ed4a4d2ded46d98c4a7bed2e37dd72/src/connector.ts#L170

MichaelSun90:
Hi @WikiRik , Thanks for sharing your feedback. For the DeprecationWarning, I did a bit investigation. Seems this is returned by the lookup function relate to this: [DEP0118: dns.lookup() support for a falsy host name](https://nodejs.org/api/deprecations.html#DEP0118). The description mentions this will become an error in the future. We will discuses this, make sure proper error is thrown and see if we need to do any additional handling on this.

PR #1598, opened 2024-02-06
chore!: drop node 16 and 19 support

MichaelSun90:
BREAKING CHANGE: Node.js 16.x and 19.x are no longer supported by `tedious`.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -19,10 +19,10 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
-    - name: Use Node.js 16
+    - name: Use Node.js 18
       uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version: 18
 
     - name: Determine npm cache directory
       id: npm-cache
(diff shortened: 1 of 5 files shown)


.github/workflows/nodejs.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 19.x]
+        node-version: [18.x, 19.x, 20.x, 21.x]
```
mShan0, line 49: We can drop 19 support since it is no longer supported.

appveyor.yml
```diff
@@ -2,7 +2,6 @@ version: "{build}"
 
 environment:
   matrix:
-    - nodejs_version: "16"
     - nodejs_version: "18"
     - nodejs_version: "19"
```
mShan0, line 6: We can drop 19 support since it is no longer supported. Should we add 20 and 21 here?

arthurschreiber: Yes, let's drop support for 19 and add 20 and 21.

MichaelSun90: Just push a commit for adding 20, 21 and remove 19 👍 @arthurschreiber , the CI still expecting check for node 16 and 19, is this configurable from one of the yml files or it is a ci setting? 🤔

PR #1608, opened 2024-02-28
fix: error handling for domainToASCII

MichaelSun90:
switch from punyCode to node:url  introduce a behavior difference: https://github.com/tediousjs/tedious/pull/1597#issuecomment-1953027053

This fix the behavior.


MichaelSun90:
Hi @arthurschreiber, last time when we check this PR, you mentioned that description for this PR mentioned a test but it is missing from the changed files. I have went over the past commits, and I think we decided to remove the test since it seems the test is not needed. If you can give this PR a other look when you got time, see if we can merge this.

PR #1609, opened 2024-03-01
feat(types): export `ConnectionConfiguration` type

WikiRik:
Thanks for adding first-party typings! I've noticed that ConnectionConfiguration wasn't exported, while we are using that for sequelize here; https://github.com/sequelize/sequelize/blob/main/packages/core/src/dialects/mssql/connection-manager.ts (@types/tedious calls it ConnectionConfig but the object is basically the same as ConnectionConfiguration)

I went ahead and just exported the type here, but I'm open to hear if you have other suggestions on what type to use.

### DIFF
--- a/src/tedious.ts
+++ b/src/tedious.ts
@@ -20,7 +20,6 @@ export function connect(config: ConnectionConfiguration, connectListener?: (err?
 export {
   BulkLoad,
   Connection,
-  ConnectionConfiguration,
   Request,
   library,
   ConnectionError,
@@ -28,4 +27,8 @@ export {
   TYPES,
   ISOLATION_LEVEL,
   TDS_VERSION
+};
+
+export type {
+  ConnectionConfiguration
 };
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Cool, looks good to me! Will get it merged once I'm close to a computer.

WikiRik:
I should have ran tests locally first, but exporting the type should also work. Let's hope CI is green now, otherwise I will run stuff locally

WikiRik:
CI is mostly green, just codecov is more limited. Feel free to let me know if you want me to add a test (and mention where)

arthurschreiber:
> CI is mostly green, just codecov is mentioning that coverage has gone down. Feel free to let me know if you want me to add a test (and mention where)

Coverage is very finicky - in this case we're adding a line of code that can't really be tested in any sensible way (remember type definitions are completely stripped at runtime which is when coverage is collected).

Coverage going down is not a problem. Thank you for preparing the change (and fixing the typo in the title lol 🤣).

Issue #1610, opened 2024-03-08
Getting RangeError and Nodejs throws error for "bigint" type column if float values are provided in bulk load while using bulk insert 


shivamrai45:
**Software versions**
* Tedious: 17.0.0
* SQL Server: MsSQL Server 2017 - 14.0.3381.3
* Node.js: 18.12.1

**Additional Libraries Used and Versions**
NA

**Table schema**
Columns - MyBigInt (bigint 8)

**Connection configuration**
`
{ authentication: { type: 'default', options: { userName: '***', password: '***' } }, server: 'ms-sql-sql-server-2017.***.amazonaws.com', options: { database: 'MsSQL_Restart', encrypt: true, port: 1433, requestTimeout: 100000, trustServerCertificate: true } }
`

**Problem description**
I am trying to bulk insert data in MsSQL using bulkLoad. The destination column contains bigint datatype column with size 8, If we provide it with a float/decimal value it crases the server while inserting in buffer [here](https://github.com/tediousjs/tedious/blob/d26df9b232fbbdf4d32a8f4611a4672728b9f879/src/data-types/bigint.ts#L35)
I've attached a sample script to replicate the behaviour
[tediousTest.txt](https://github.com/tediousjs/tedious/files/14535459/tediousTest.txt)

**Expected behavior**
Ideally we should validate the data and return a error if invalid datatype is present in data or in case of float, round off to greatest minimum integer as we do in case of query run's for mssql.

**Actual behavior**
Node internals throw error, thus crashing the server

**Error message/stack trace**
RangeError: The number 0.5 cannot be converted to a BigInt because it is not an integer
    at BigInt (<anonymous>)
    at WritableTrackingBuffer.writeInt64LE (node_modules/tedious/lib/tracking-buffer/writable-tracking-buffer.js:103:26)
    at Object.generateParameterData (node_modules/tedious/lib/data-types/bigint.js:33:12)
    at generateParameterData.next (<anonymous>)
    at RowTransform._transform (node_modules/tedious/lib/bulk-load.js:126:18)
    at Transform._write (node:internal/streams/transform:175:8)
    at doWrite (node:internal/streams/writable:411:12)
    at clearBuffer (node:internal/streams/writable:572:7)
    at onwrite (node:internal/streams/writable:464:7)
    at node:internal/streams/transform:190:7
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)

**Any other details that can be helpful**

[...]

PR #1620, opened 2024-04-17
fix: fix `bigint` validation

MichaelSun90:
This PR is for fixing issue #1610 
Now the validate function will throw an error if a invalid value is passed into BigInt type :
"The number <value> cannot be converted to a BigInt because it is not an integer"


dhensby:
Hi team, it seems that this change has caused a regression that means a test in the `mssql` lib is breaking.

We're making an assertion that a bigint with value `0` === `0` but since 18.2.0 it is now `null` (not entirely sure how that's happening, but it's what I've observed) - we can see it here: https://github.com/tediousjs/node-mssql/pull/1664

dhensby:
OK - so this was a problem in our library because the `cast` function we have doesn't support native bigints. I would suggest that changing the return type of BigInt to use native bigints is a breaking change, so probably shouldn't have been released in a minor release.

Issue #1623, opened 2024-04-30
[FEATURE REQUEST] Custom `ChainedTokenCredential`


elliot-huffman:
## Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it

The default azure credential isn't always the best way to authenticate, sometimes people need to authenticate using a custom chain, e.g. by using the VS Code plugin, which isn't available in the default chain or to do some fancy environmental variable stuff using azure app config and or azure key vault.
These people would create their own credential chain by using the `@azure/identitiy` package's `ChainedTokenCredential` class.

## Describe the preferred solution

Add an additional authentication type which accepts a class that implements the chained token credential class from `@azure/identity`.

Additionally, this type of support would allow for the code configurations for `azure-active-directory-default`, `azure-active-directory-password`, `azure-active-directory-msi-vm`, `azure-active-directory-msi-app-service`, and `azure-active-directory-service-principal-secret` to be removed from the maintenance cycle to reduce codebase complexity as they would be able to be offloaded to the chained credential class implementation in the `@azure/identity` package.

Tl;Dr:
simplify the auth types in Tedious.JS by offloading the majority of auth to the provided `ChainedTokenCredential` instance.

## Describe alternatives you've considered

Default Azure credential (about to migrate to this), custom coded error handler to re-initialize the connection (current method) when the token expires.

## Additional context

#1144

## Reference Documentations/Specifications
https://tediousjs.github.io/tedious/api-connection.html#function_newConnection
https://www.npmjs.com/package/@azure/identity
https://learn.microsoft.com/en-us/javascript/api/%40azure/identity/?view=azure-node-latest

elliot-huffman:
I just created a pull request with the required code to enable this functionality.
I did not remove the existing auth methods so that they could be removed at a future date with plenty of communication as it would be a breaking API change (removing functionality is a semver major change).

Issue #1144, opened 2020-08-12
azure-active-directory-access-token REFRESH conn  automatically on expiring


panbhatt:
I am working on application, where we are connecting to the SQL Server database and trying to create the connection pool (with TYPEORM and Sequelize). As per the org policies our token expires every 24 hours, that means all the connections in the pool are getting expired, and we have to restart the app to make sure every object /pool is created again. 

Do we have an internal functionality to refresh the token automatically when it expires. It seems so obvious that the solution to restart the application wont' fly.  Is there any way in the app we can do it, so that we can save restarts. 
Any help would be appreciated.

IanChokS:
Hi @panbhatt I'll take a look and see if there's a way for tedious to do that!

panbhatt:
Thanks @IanChokS  : What i think can be done is that besides taking the token details, we can ask for a function (or a object having a function) that can be called once the duration of the token got expired. However, i would leave it up to the community to decide what best can be done. 
The other way to achieve this to to hack the ORM solutions and call their connection manager (which is bypass) and i dont think its a good way to go.

cameronstubber:
I think a good solution could be that we have a boolean flag in the configuration object used at the start to signal if we want to auto refresh or not.

I'm sure there is somebody out there who is happy with the session expiring.

panbhatt:
Any update on this, team.

arthurschreiber:
@panbhatt I think that's not really an issue in `tedious`, but an issue in the connection pool library you're using? If the connection pool library would allow specifying a factory function to "build" the connection (or connection configuration), you could use that to refresh the token ever so often.

panbhatt:
the only caveat with this issue is all the ORM's has to change their structure just for this. i.e. they have to implement this functionality, which i think its not required, as the problem is present at TEDIOUS JS level, if in the dialect options they take a function (since they are the only who is providing the functionality for token validation etc) and refresh their connection details, so that whenever a new connection is required, it always used the fresh token. 
i do not understand why we need to change 10's of other libraries, when this can be done at one central place.

arthurschreiber:
@panbhatt I understand. How do you feel about allowing to pass in an async function as `token` for the `azure-active-directory-access-token`, and we then call this function to get the token when establishing a connection?

Would you be willing to open a PR that implements this?

panbhatt:
[...]

PR #1624, opened 2024-05-01
feat: Add Token Credential Support

elliot-huffman:
## Overview

Adds support for the base token credential authentication that is implemented by [`@Azure/Identity`](https://www.npmjs.com/package/@azure/identity).
This PR does not remove support for the other [`@Azure/Identity`](https://www.npmjs.com/package/@azure/identity) methods.

The other auth methods are now able to be removed as the credential chain type supports whatever the end user throws at it. Default, service principal, and any combination of crazy auth methods and all future ones that MSFT adds to the MSAL system.
All [`@Azure/Identity`](https://www.npmjs.com/package/@azure/identity) auth classes derive from the credential chain class so all the other existing authentication methods are supported through this specific option as well as 100% of their configurations with no additional code that is required from us.

See #1623 for more info.
This would be the core auth system work required to solve issues like #1144 as they would be able to use auth that makes sense for them.

## Sample Documentation

- `microsoft-credential-chain`
```TypeScript
const msAuth = new DefaultAzureCredential()

"authentication": {
  "type":'microsoft-credential-chain',
  ...,
  "options": {
    "credential": msAuth
  }
}
```
- credential (TokenCredential): Instance of a token credential from a package like [`@Azure/Identity`](https://www.npmjs.com/package/@azure/identity). This could be any of the built-in credential types such as `DefaultAzureCredential`, `EnvironmentCredential`, `ChainedTokenCredentia`, `VisualStudioCodeCredential`, etc. All the MS supported classes derive from the `TokenCredential` interface so all the pre-built auth methods are supported by inheritance. See the @Azure/Identity package's documentation (https://www.npmjs.com/package/@azure/identity) for more ways to customize your authentication experience for your environment's specifics.

(optional section)
**Best Practices**
Migrate from other `azure-active-directory` prefixed auth config types to this one as they will be removed in the future to simplify the code base.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -10,11 +10,12 @@ import { type SecureContextOptions } from 'tls';
 import { Readable } from 'stream';
 
 import {
-  DefaultAzureCredential,
-  ClientSecretCredential,
-  ManagedIdentityCredential,
-  UsernamePasswordCredential,
-} from '@azure/identity';
+    ChainedTokenCredential,
+    ClientSecretCredential,
+    DefaultAzureCredential,
+    ManagedIdentityCredential,
+    UsernamePasswordCredential,
+  } from '@azure/identity';
 
 import BulkLoad, { type Options as BulkLoadOptions, type Callback as BulkLoadCallback } from './bulk-load';
 import Debug from './debug';
@@ -324,11 +325,22 @@ interface DefaultAuthentication {
   };
 }
 
+/** Structure that defines the options that are necessary to authenticate the tedious instance with an `@azure/identity` chained token credential derived class. */
+interface CredentialChainAuthentication {
+    /** Unique designator for the type of authentication to be used. */
+    type: 'microsoft-credential-chain';
+    /** Set of configurations that are required or allowed with this authentication type. */
+    options: {
+        /** Credential object used to authenticate to the resource. */
+        credential: ChainedTokenCredential;
+    };
+}
+
 interface ErrorWithCode extends Error {
   code?: string;
 }
 
-export type ConnectionAuthentication = DefaultAuthentication | NtlmAuthentication | AzureActiveDirectoryPasswordAuthentication | AzureActiveDirectoryMsiAppServiceAuthentication | AzureActiveDirectoryMsiVmAuthentication | AzureActiveDirectoryAccessTokenAuthentication | AzureActiveDirectoryServicePrincipalSecret | AzureActiveDirectoryDefaultAuthentication;
+export type ConnectionAuthentication = DefaultAuthentication  | NtlmAuthentication | CredentialChainAuthentication | AzureActiveDirectoryPasswordAuthentication | AzureActiveDirectoryMsiAppServiceAuthentication | AzureActiveDirectoryMsiVmAuthentication | AzureActiveDirectoryAccessTokenAuthentication | AzureActiveDirectoryServicePrincipalSecret | AzureActiveDirectoryDefaultAuthentication;
 
 interface InternalConnectionConfig {
   server: string;
@@ -417,6 +429,7 @@ interface State {
 
 type Authentication = DefaultAuthentication |
   NtlmAuthentication |
+  CredentialChainAuthentication |
   AzureActiveDirectoryPasswordAuthentication |
   AzureActiveDirectoryMsiAppServiceAuthentication |
   AzureActiveDirectoryMsiVmAuthentication |
@@ -482,6 +495,7 @@ interface AuthenticationOptions {
    *
    * * `default`: [[DefaultAuthentication.options]]
    * * `ntlm` :[[NtlmAuthentication]]
+   * * `microsoft-credential-chain`: [[CredentialChainAuthentication.options]]
    * * `azure-active-directory-password` : [[AzureActiveDirectoryPasswordAuthentication.options]]
    * * `azure-active-directory-access-token` : [[AzureActiveDirectoryAccessTokenAuthentication.options]]
    * * `azure-active-directory-msi-vm` : [[AzureActiveDirectoryMsiVmAuthentication.options]]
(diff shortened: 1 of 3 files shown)


arthurschreiber (commented):
Sweet, this looks great! Can you fix the formatting to use 2 spaces instead of tabs? 🙇

elliot-huffman:
> Sweet, this looks great! Can you fix the formatting to use 2 spaces instead of tabs? 🙇

Fixed!

src/connection.ts
```diff
@@ -290,6 +291,17 @@ interface AzureActiveDirectoryServicePrincipalSecret {
   };
 }
 
+/** Structure that defines the options that are necessary to authenticate the tedious instance with an `@azure/identity` chained token credential derived class. */
+interface CredentialChainAuthentication {
+    /** Unique designator for the type of authentication to be used. */
```
arthurschreiber, line 296: I think this is still using tabs instead of spaces. 😬

arthurschreiber:
Would you mind updating the lock file as well?

src/connection.ts
elliot-huffman: Sorry about that, I tried to regex search `/t` in GitHub but I found nothing. Should be fixed now.

elliot-huffman:
> Would you mind updating the lock file as well?

That would be important. One sec.

elliot-huffman:
I believe it is fixed now.
I found an unrelated bug when I added some types to variables as part of this change. It was a bug where the access token could be null, so I added a type guard that triggered the correct error if this case is ever reached.

src/connection.ts
elliot-huffman: @arthurschreiber, I found one other place where the formatting might not be correct. Fixed that too.

elliot-huffman: In the future, I will 100% not be free handing this stuff in the GitHub file editor. I will follow the PR instructions and do it locally :-P

arthurschreiber:
I had some more time to look at this and think about this. 😅 

How do you feel about renaming the credential method to `azure-identity-token-credential`, and change this to only use the `TokenCredential` interface instead? `TokenCredential` is the _actual_ thing we're interested in - we don't actually care whether it's a chained token credential or not (and apparently there's various token credentials that don't inherit from `ChaintedTokenCredential`: https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-js%20%22implements%20TokenCredential%22&type=code)

arthurschreiber:
Also, in general I'd be in favor of deprecating (and removing) the other azure identity related methods - it would allow us to completely drop the runtime dependency on `@azure/identity`, resolving some long standing issues around startup performance.

src/connection.ts
```diff
@@ -290,6 +293,17 @@ interface AzureActiveDirectoryServicePrincipalSecret {
...
 }
 
+/** Structure that defines the options that are necessary to authenticate the tedious instance with an `@azure/identity` chained token credential derived class. */
+interface CredentialChainAuthentication {
+  /** Unique designator for the type of authentication to be used. */
+  type: 'microsoft-credential-chain';
+  /** Set of configurations that are required or allowed with this authentication type. */
+  options: {
+    /** Credential object used to authenticate to the resource. */
+    credential: ChainedTokenCredential;
```
arthurschreiber, line 303: I think we should change the type here to `credential: TokenCredential | (): TokenCredential | (): Promise<TokenCredential>`

This would allow clients to re-use the connection configuration and re-generate tokens whenever a connection is opened with the same configuration automatically.

elliot-huffman:
> Also, in general I'd be in favor of deprecating (and removing) the other azure identity related authentication methods - it would allow us to completely drop the runtime dependency on `@azure/identity`, resolving some long standing issues around startup performance.

I believe the actual constructor validation that I wrote is for token credential instead of change credential, as I was testing with some non-chained credentials it failed and I had to update the validation to support it properly.
The documentation I wrote was definitely for change credentials but I will update that wording even though the back end is looking for the token credential structure. 
The token credential class is not an actual class, it's just an interface, the constructor type guard I wrote is for validating that interface. I found out the hard way when I tried to do an instanceof keyword against it, LOL.

elliot-huffman:
> I had some more time to look at this and think about this. 😅 
> 
> How do you feel about renaming the credential method to `azure-identity-token-credential`, and change this to only use the `TokenCredential` interface instead?
> 
> `TokenCredential` is the _actual_ thing we're interested in - we don't actually care whether it's a chained token credential or not (and apparently there's various token credentials that don't inherit from `ChainedTokenCredential`: https://github.com/search?q=repo%3AAzure%2Fazure-sdk-for-js%20%22implements%20TokenCredential%22&type=code)

I can definitely rename it :-)

src/connection.ts
elliot-huffman: Also, since `Azure AD` is the old name of Microsoft's directory service. I think `microsoft-token-credential` would be a better name since it would be future proofed for when MSFT eventually renames the or changes the terminology.

What do you think?

elliot-huffman: > I think we should change the type here to `credential: TokenCredential | () => TokenCredential | () => Promise<TokenCredential>`
> 
> This would allow clients to re-use the connection configuration and re-generate tokens whenever a connection is opened with the same configuration automatically.

Also, did I read this right? do you want me to create some sort of function invoker thingy?

arthurschreiber: Yeah, but we can also just skip this for now and add it later. Sorry for the confusion 🙇‍♂️

elliot-huffman: I will make a second Issue for this for this behavior.

elliot-huffman:
Just read through the lint report, this one is not my fault:
image

I am planning to overhaul the linting system with a future PR and this would be fixed at that time.

If you are ok with it, would you be able to override this one?

src/connection.ts
```diff
@@ -3094,10 +3120,7 @@ class Connection extends EventEmitter {
     }
   }
 
-  /**
-   * @private
-   */
-  makeRequest(request: Request | BulkLoad, packetType: number, payload: (Iterable<Buffer> | AsyncIterable<Buffer>) & { toString: (indent?: string) => string }) {
+  private makeRequest(request: Request | BulkLoad, packetType: number, payload: (Iterable<Buffer> | AsyncIterable<Buffer>) & { toString: (indent?: string) => string }) {
```
arthurschreiber, line 3123: Do you mind switching this back to not be `private`? This is breaking some unrelated code.

elliot-huffman: interesting, it fails a compile if I keep it as a JS Doc though, specifically 'type annotaiton not expected'

src/connection.ts
```diff
@@ -290,6 +291,17 @@ interface AzureActiveDirectoryServicePrincipalSecret {
   };
 }
 
+/** Structure that defines the options that are necessary to authenticate the Tedious.JS instance with an `@azure/identity` token credential. */
+interface TokenCredentialAuthentication {
+  /** Unique designator for the type of authentication to be used. */
+  type: 'microsoft-token-credential';
```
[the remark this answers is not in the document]
elliot-huffman: I just want to keep the namespace from anything else that could be a token credential in the future that isn't from MSFT but I do also see that this is a low possibility, I personally sorta lean towards keeping it prefixed with `microsoft` but I will remove it if you still lean the other way. I am not that strongly opiniated on it too.

elliot-huffman:
It looks like it ran successfully, it is just failing to connect to codecov? I could be wrong; I don't use that service yet. @arthurschreiber, any todo items for me?

arthurschreiber:
I'm trying to figure out what's going on with codecov, seems like it needs to be updated to the latest version of the action (see https://github.com/tediousjs/tedious/pull/1627).

arthurschreiber:
Would you be okay with me taking the changes done so far in here and opening a new PR? I'd like to make sure we have tests running for this new functionality, but azure tests don't run on pull requests originating from forks because of security restrictions.

elliot-huffman:
> Would you be okay with me taking the changes done so far in here and opening a new PR? I'd like to make sure we have tests running for this new functionality, but azure tests don't run on pull requests originating from forks because of security restrictions.

On the condition that I get recognition, my ego demands it :-P

maybe being added to the contributors section of the `package.json` as "Elliot Huffman <ehuffman@elliot-labs.com> (https://elliot-labs.com)" or something like that.

arthurschreiber:
The commit(s) will stay attributed to you. 😬

elliot-huffman:
> The commit(s) will stay attributed to you. 😬

my bad, works for me!

PR #1630, opened 2024-05-03
feat: IDE Integration

elliot-huffman:
Changes:
- Updated Dev Container config to latest spec
  - The VS Code configs are now nested options, not root level configs
- Updated dev container Node.JS version from Node 16 LTS to Node 20 LTS
- Migrate to VS Code Workspace Config (off the `settings.json` file)
- Add recommended extensions to the dev container and VS Code workspace for better Node.JS dev integration (Docker, Dev Containers, Socket.dev, spell checker, MS SQL, and a markdown linter)
- Create build task for VS Code
- Create launch config for VS Code (so that native VS Code debugging is possible)
- Randomized the SA Password to push for better credential habits (no functional changes)
- Merge recommended `.gitignore` from GitHub for `Sublime Text` and `Node.JS`.
  - This should prevent additional tyles of artifacts from accidentally getting committed.


.vscode/settings.json
```diff
@@ -1,19 +0,0 @@
...
-  "[json]": {
-    "editor.insertSpaces": true,
-    "editor.tabSize": 2
-  },
-  "javascript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
-  "typescript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
-  "files.trimTrailingWhitespace": true,
-  "files.insertFinalNewline": true,
-  "files.trimFinalNewlines": true
-}
```
MichaelSun90, line 19: I saw these configuration are moved into Tedious.code-workspace.  Is settings.json becoming a outdated way to handle this or it is just simpler to do this within the Tedious.code-workspace?

.devcontainer/docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
 
     environment:
       - "ACCEPT_EULA=Y"
-      - "SA_PASSWORD=yourStrong(!)Password"
+      - "MSSQL_SA_PASSWORD=QhNemHjCWHui9vNnRC.TMRAhusvM3n9_"
```
MichaelSun90, line 24: May I ask is this password here generated in someway during the run time?

MichaelSun90 (commented):
Thank you for the PR. I just added few question based on my limited knowledge on IDE setup.  It will still be better to get some feed back from @arthurschreiber to finalize the changes within the PR.

.vscode/settings.json
elliot-huffman: settings.json is fully supported and not going away to the best of my knowledge. 

I believe it is simpler as you can centralize more configurations in the workspace config. It technically gives you more capabilities as settings.json relates only to a single set of configurations for the folder, the configuration of the workspace allows for more options and integrations and I believe it also has codespace support.
Some examples are folder exclusions on the workspace to hide stuff that is not necessary, this is not supported on settings.json

elliot-huffman: You can also include extension recommendations in the same file. 
It is also better for monolithic projects.

.devcontainer/docker-compose.yml
elliot-huffman: It is not generated, I just regenerated the password to reduce reliance on hardcoded non-random passwords. 
It doesn't change any behavior.

I did update the environmental variable name as the one that was used is no longer supported.

PR #1631, opened 2024-05-03
chore: Convert ES Lint to Flat Config

elliot-huffman:
Changes:
- Migrate typescript-eslint to the next gen package (https://typescript-eslint.io/blog/announcing-typescript-eslint-v7#switching-to-typescript-eslint).
- Convert config to Flat Config
- Remove deprecated rules.
- Remove duplicate rules for rulesets. (to be added in additional PRs)
- Remove removed rules.
- Update package config lint commands to the modern standard.
- Split `tsc` command in lint to a `postlint` command so that it simplifies and modularizes the lint command set.


elliot-huffman:
Once the #1630 PR is merged, then I can add the IDE configs required for the new flat files into this PR

arthurschreiber:
This is great, thank you! ❤️ 

Would you mind changing this pull request to only focus on switching to the flat config, without changing the rules that are applied? I'm open to changing the linting rules, but I'd prefer a separate PR per changed lint rule so that we can have all config change plus all the required code changes per rule change together.

Otherwise, this PR will just become too unwieldy and impossible to review. 🙇‍♂️

elliot-huffman:
Works for me, changes incoming on Monday!

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

elliot-huffman:
@arthurschreiber, the linting rulesets have been disabled. I will enable one ruleset at a time and have the contents of the project fixed and make that a PR.

elliot-huffman:
[`no-loss-of-precision`](https://eslint.org/docs/latest/rules/no-loss-of-precision) is a new rule that was just introduced to check for number casting. It was not previously listed in the rule config and has found a bunch of potential cast errors (that look accurate to me).
This commit will fail linting because of this new rule that is enabled by default.

arthurschreiber:
The reason for the new lint failures for files in the `test` folder is because there exists a separate `test/.eslintrc` file that defines the linter configuration for these files. Can you convert that into the new config format as well? That should remove all the test failures.

We can later focus on aligning the configuration between the test folder and the rest of the source.

elliot-huffman:
> The reason for the new lint failures for files in the `test` folder is because there exists a separate `test/.eslintrc` file that defines the linter configuration for these files. Can you convert that into the new config format as well? That should remove all the test failures.
> 
> We can later focus on aligning the configuration between the test folder and the rest of the source to use the same linter settings.

Oh, I missed this, can do!

PR #1634, opened 2024-06-13
test: use a JavaScript based config file format

arthurschreiber:
This replaces the previously used JSON based configuration file with a JavaScript based configuration file.

The new config format allows to represent configurations that wouldn't have been possible with the JSON format, like the [proposed token credential](https://github.com/tediousjs/tedious/pull/1624) support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "tag": "next"
   },
   "dependencies": {
-    "@azure/identity": "^3.4.1",
+    "@azure/identity": "^4.2.1",
     "@azure/keyvault-keys": "^4.4.0",
     "@js-joda/core": "^5.6.1",
     "@types/node": ">=18",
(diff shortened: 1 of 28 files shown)


package.json
```diff
@@ -85,7 +85,7 @@
     "docs": "typedoc",
     "lint": "eslint src test --ext .js,.ts && tsc",
     "test": "mocha --forbid-only test/unit test/unit/token test/unit/tracking-buffer",
-    "test-integration": "mocha --forbid-only test/integration/",
+    "test-integration": "mocha test/integration/",
```
MichaelSun90, line 88: Hi @arthurschreiber , is this forbid-only removed by an accident or it will inference with the new configuration handle for integration test? Or just for easy debugging?

arthurschreiber: Ah, no, that's a mistake and needs to be reverted before this can be merged. Good catch!

PR #1642, opened 2024-07-03
fix: don't convert BigInt type to Number before writing to buffer to prevent RangeError

paulish:
Passing big integer values as bigint parameters caused a RangeError with the following message:
`RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= -(2n ** 63n) and < 2n ** 63n. Received 9_223_372_036_854_775_808n`.

The patch removes this unneeded conversion and instead convert other types to BigInt.


arthurschreiber:
Thanks! ❤️

Issue #1550, opened 2023-07-05
[FEATURE REQUEST] AsyncIterator for Request class (promise-based streaming)


chdh:
I suggest to implement an `AsyncIterator` for the `Request` class. 

This would allow promise-based streaming of query results.

Example of how to use it:

```
const request = new Request(sql);
connection.execSql(request);
for await (const row of request) {
   console.log(row);
}
```

arthurschreiber:
This is a great idea, and I've thought about this a bit before. The main problem is that a single `Request` does not mean you get a single "type" of rows - there's multiple rowstreams which each will come with their own column definitions.

Do you have a suggestion how this could be modeled?

chdh:
> there's multiple rowstreams which each will come with their own column definitions.

A simple solution would be to use the following structure for the items returned by the iterator:

```
interface RequestIteratorItem {
   row: ColumnValue[] | Record<string, ColumnValue>;
   resultSetNo: number;
   columnMetadata: ColumnMetaData[] | Record<string, ColumnMetaData>;
}
```

The user would have to detect a change in `resultSetNo` to process the start of a new result set.

chdh:
@arthurschreiber I already have a working private implementation of an `AsyncIterator` for the Tedious `Request` class. I needed the functionality for a client project. I could use it as the basis for a PR.

Would you be interested in a PR from me, or have you already started development yourself?

arthurschreiber:
I'd be happy if you open a PR. 🥳

PR #1644, opened 2024-08-01
feat: implement an async iterable request class

chdh:
This PR implements an async iterable `request` class.

The new class `IterableRequest` is implemented as a super class of the normal `Request` class. It is an `AsyncIterable` providing an `AsyncIterator` and can be used with [`for await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of).

Usage:

```js
const request = new IterableRequest("select 42, 'hello world'");
connection.execSql(request);
for await (const item of request) {
  console.log(item.row);
}
```

A unit test is included. The API documentation has yet to be written.

Implements #1550

### DIFF
--- a/src/iterable-request.ts
+++ b/src/iterable-request.ts
@@ -1,0 +1,254 @@
+// This module implements an iterable `Request` class.
+
+import Request, { type RequestOptions } from './request';
+import { type ColumnMetadata } from './token/colmetadata-token-parser';
+
+export interface ColumnValue {
+  metadata: ColumnMetadata;
+  value: any;
+}
+
+type RowData = ColumnValue[] | Record<string, ColumnValue>;                    // type variant depending on config.options.useColumnNames
+type ColumnMetadataDef = ColumnMetadata[] | Record<string, ColumnMetadata>;    // type variant depending on config.options.useColumnNames
+
+export interface IterableRequestOptions extends RequestOptions {
+  iteratorFifoSize: number;
+}
+
+/**
+* The item object of the request iterator.
+*/
+export interface IterableRequestItem {
+
+  /**
+  * Row data.
+  */
+  row: RowData;
+
+  /**
+  * Result set number, 1..n.
+  */
+  resultSetNo: number;
+
+  /**
+  * Metadata of all columns.
+  */
+  columnMetadata: ColumnMetadataDef;
+}
+
+type iteratorPromiseResolveFunction = (value: IteratorResult<IterableRequestItem>) => void;
+type iteratorPromiseRejectFunction = (error: Error) => void;
+
+// Internal class for the state controller logic of the iterator.
+class IterableRequestController {
+
+  private request: Request;
+  private requestCompleted: boolean;
+  private requestPaused: boolean;
+  private error: Error | undefined;
+  private terminating: boolean;
+
+  private resultSetNo: number;
+  private columnMetadata: ColumnMetadataDef | undefined;
+  private fifo: IterableRequestItem[];
+  private fifoPauseLevel: number;
+  private fifoResumeLevel: number;
+
+  private promisePending: boolean;
+  private resolvePromise: iteratorPromiseResolveFunction | undefined;
+  private rejectPromise: iteratorPromiseRejectFunction | undefined;
+  private terminatorResolve: (() => void) | undefined;
+
+  // --- Constructor / Terminator ----------------------------------------------
+
+  constructor(request: Request, options?: IterableRequestOptions) {
+    this.request = request;
+    this.requestCompleted = false;
+    this.requestPaused = false;
+    this.terminating = false;
+
+    this.resultSetNo = 0;
+    this.fifo = [];
+    const fifoSize = options?.iteratorFifoSize ?? 1024;
+    this.fifoPauseLevel = fifoSize;
+    this.fifoResumeLevel = Math.floor(fifoSize / 2);
+
+    this.promisePending = false;
+
+    request.addListener('row', this.rowEventHandler);
+    request.addListener('columnMetadata', this.columnMetadataEventHandler);
+  }
+
+  public terminate(): Promise<void> {
+    this.terminating = true;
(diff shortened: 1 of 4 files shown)


arthurschreiber:
First of all, this is super exciting - thank you for working on this! ❤️ 

I haven't taken a super deep look into the implementation yet, but my idea for this was always to change the API for request execution to something like this:

```ts
// Create a new request, but without callback.
const request = new Request('select a, b, c from foobar');

// `using` so any resources are free'd to make the connection reusable in case some error happens in the user's code
using response = await connection.execSql(request);

// `.values` for a rows as arrays
// `.objects` for rows as objects
for await (const [a, b, c] of response.values()) {
  ...
}
```

Obviously, making such a profound change to the API is not simple, and there's going to be a lot of concerns around backwards compatibility and things like that.

src/iterable-request.ts
```diff
@@ -0,0 +1,254 @@
...
+    function completionCallback(error: Error | null | undefined) {
+      if (controller) {
+        controller.completionCallback(error);
+      }
+    }
+  }
+
+  [Symbol.asyncIterator](): AsyncIterator<IterableRequestItem> {
+    return this.iterator;
+  }
```
arthurschreiber, line 249: Have you thought about implementing this through using `events.on`? I haven't tested this, but it would allow you to get rid of the custom iterator implementation.

```
  constructor(sqlTextOrProcedure: string | undefined, options?: IterableRequestOptions) {
    super(sqlTextOrProcedure, completionCallback, options);
    const controller = new AbortController();
    this.doneSignal = controller.signal;

    function completionCallback(error: Error | null | undefined) {
       controller.abort(error);
    }
  }

  [Symbol.asyncIterator](): AsyncIterator<IterableRequestItem> {
    return events.on(this, 'row', { signal: this.doneSignal });
  }
```

src/iterable-request.ts
```diff
@@ -0,0 +1,254 @@
...
+import Request, { type RequestOptions } from './request';
+import { type ColumnMetadata } from './token/colmetadata-token-parser';
+
+export interface ColumnValue {
+  metadata: ColumnMetadata;
+  value: any;
+}
+
+type RowData = ColumnValue[] | Record<string, ColumnValue>;                    // type variant depending on config.options.useColumnNames
+type ColumnMetadataDef = ColumnMetadata[] | Record<string, ColumnMetadata>;    // type variant depending on config.options.useColumnNames
```
arthurschreiber, line 12: `config.options.useColumnNames` is super bad API design, and I'd prefer if we could not propagate this nonsense. 😬

The reason why it's bad API design is that the config value changes the _shape_ of the data that is being returned. Flipping this config basically from `true` to `false` or the other way around basically breaks all code that was written with the other value set.

Much better would be to have an API that allows the code to explicitly either use objects or use arrays.

chdh: I agree. But as you wrote before, this would be a profound change, beyond the scope of this PR.

src/iterable-request.ts
chdh: Thanks. I didn't know about `events.on()`. It does more or less the same as my controller and iterator classes.
I will refactor my code into using `events.on()`.

chdh: Since Tedious still has to support Node 18, I had a look at that [old implementation of `event.on()`](https://github.com/nodejs/node/blob/e578c0b1e8d3dd817e692a0c5df1b97580bc7c7f/lib/events.js#L1046).
The `close` event was not yet implemented in that version. I can't find a proper way to end the iteration so that the remaining items in the FIFO are still processed. `abortListener()` calls `errorHandler()` which calls `toError.reject(err)` which rejects the waiting `Promise`, 
Calling `iterator.return()` resolves the waiting `Promise` with `done = true`.

Should I use my implementation for old Node versions and `event.on()` for new Node versions?

src/iterable-request.ts
```diff
@@ -0,0 +1,266 @@
...
+  constructor(request: Request, options?: IterableRequestOptions) {
+    this.request = request;
+    this.requestCompleted = false;
+    this.requestPaused = false;
+    this.terminating = false;
+
+    this.resultSetNo = 0;
+    this.fifo = [];
+    const fifoSize = options?.iteratorFifoSize ?? 1024;
+    this.fifoPauseLevel = fifoSize;
```
arthurschreiber, line 74: I guess this is similar to `highWaterMark` on regular streams? I.e. once the buffer (here called `fifo`) reaches that size, no more data will be buffered and the underlying request will be paused?

src/iterable-request.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  public async return(value?: any): Promise<any> {
+    await this.controller.terminate();
+    return Promise.resolve({ value, done: true });         // eslint-disable-line @typescript-eslint/return-await
+  }
+
+  public async throw(exception?: any): Promise<any> {
+    await this.controller.terminate();
+    if (exception) {
+      return Promise.reject(exception);                    // eslint-disable-line @typescript-eslint/return-await
```
arthurschreiber, line 221: Could you `throw exception` here?

src/iterable-request.ts
```diff
@@ -0,0 +1,266 @@
...
+    } else {
+      return Promise.resolve({ done: true });              // eslint-disable-line @typescript-eslint/return-await
```
arthurschreiber, line 223: Could you `return { done: true }` here?

src/iterable-request.ts
chdh: Yes. Do you suggest to rename it to `this.fifoHighWaterMark`? And also rename the option parameter?

src/iterable-request.ts
chdh: Yes, of course. Thanks. I have changed it.

src/iterable-request.ts
chdh: Yes. Thanks.

PR #1651, opened 2024-08-16
feat: enable release attestation through GitHub Actions

MichaelSun90:


arthurschreiber:
@MichaelSun90 Thanks for preparing this! I changed the PR title to generate a feature release - this is not _really_ a new feature but given it's importance I feel like making it a bit more prominent makes sense.

PR #1654, opened 2024-08-18
refactor!: cleanup `AbortSignal` handling

arthurschreiber:
BREAKING CHANGE: `tedious` now requires Node.js version 18.17 or later.


arthurschreiber:
One note from discussing this with @MichaelSun90 - this probably requires bumping the `engine` version to `>= 18.17`, as `AbortSignal.any` was added in `18.17.0`.

PR #1668, opened 2024-10-29
Move FeatureExtensions to end of Login7 Packet + use full server name for Login7

TPVallancourt:


src/token/handler.ts
```diff
@@ -246,7 +246,7 @@ export class Login7TokenHandler extends TokenHandler {
   declare connection: Connection;
 
   declare fedAuthInfoToken: FedAuthInfoToken | undefined;
-  declare routingData: { server: string, port: number } | undefined;
+  declare routingData: { server: string, port: number, login7server: string } | undefined;
```
arthurschreiber, line 249: I'd rather store the server and just the instancename here, and then just join them together when the combination of them is required?

TPVallancourt: @arthurschreiber Agreed - switched to using instance

TPVallancourt:
Hi @arthurschreiber , are there any other things I need to do get this incorporated into the library?

src/connection.ts
```diff
@@ -2484,7 +2485,7 @@ class Connection extends EventEmitter {
     }
 
     payload.hostname = this.config.options.workstationId || os.hostname();
-    payload.serverName = this.routingData ? this.routingData.server : this.config.server;
+    payload.serverName = this.routingData ? `${this.routingData.server}\\${this.routingData.instance}` : this.config.server;
```
arthurschreiber, line 2488: Is `instance` here guaranteed to always be a non-empty string? What happens if it's empty?

kenny-statsig:
@TPVallancourt Hi, just curious if this is planned to be merged and released soon? I am currently relying on this PR for Fabric Warehouse connections

jasonhiggs (approved):

TPVallancourt:
Hi @arthurschreiber ! Finally found some time to think about this PR some more. Microsoft appears to have pushed out an update this week which has made Fabric a lot faster and more reliable, so it seemed like a good time to revisit this. 

The big change since you last reviewed is that I added an `isFabric` flag in the options which will control which version of the Login7 payload we send. This is admittedly hacky but it does let people use the library for Fabric without breaking existing integrations. Let me know what you think!

Additionally, I think it would be good to add integration testing to a Fabric warehouse - not sure what the protocol is for setting something like that up for this repo. Any ideas?

Thank you!

src/connection.ts
```diff
@@ -2442,7 +2449,8 @@ class Connection extends EventEmitter {
       clientPid: process.pid,
       connectionId: 0,
       clientTimeZone: new Date().getTimezoneOffset(),
-      clientLcid: 0x00000409
+      clientLcid: 0x00000409,
```
haggholm, line 2452: That is one hell of an opaque magic number…

src/login7-payload.ts
```diff
@@ -365,6 +378,15 @@ class Login7Payload {
       fixedData.writeUInt32LE(0, offset);
     }
 
+    if (this.isFabric) {
```
David-Engel, line 381: From just a code read, I'm trying to understand how tedious works with feature extensions before this change. It seems like the previous code is writing the feature extension data to the completely wrong place (the middle instead of the end) compared to what the spec defines. My guess is that SQL Server and Azure SQL DB endpoints are just a lot more forgiving in the name of backwards compatibility with existing clients.

Is there even any need for the isFabric logic? I would think the changes would work on both Fabric and non-Fabric endpoints.

TPVallancourt: Amazing!!! Thank you! I've added the commit to this branch

PR #1671, opened 2024-10-31
feat: add `databaseMirroringPartner` event handler

### LINKED ISSUE #1669: Error connecting to MSSQL database with clustering enabled
**Software versions**
* Tedious: 18.6.1
* SQL Server: 2019
* Node.js: 18

**Additional Libraries Used and Versions**
[kysely](https://github.com/kysely-org/kysely) v0.27.4 (query builder)
[tarn](https://github.com/Vincit/tarn.js) v3.0.2 (connection pooling)

**Table schema**
N/A

**Connection configuration**
Standard config using `default` authentication method via `username` and `password`.

**Problem description**
I am trying to connect to a MSSQL database that has clustering enabled, but am getting the following error:
```
/Users/user/Documents/service-name/node_modules/tedious/src/token/handler.ts:140
    throw new UnexpectedTokenError(this, token);
          ^
UnexpectedTokenError: Unexpected token `ENVCHANGE` in `RequestTokenHandler`
    at RequestTokenHandler.onDatabaseMirroringPartner (/Users/user/Documents/service-name/node_modules/tedious/src/token/handler.ts:140:11)
```

This error is not present when connecting to our other identical MSSQL databases where clustering is disabled.

The weirdest thing about this is we have other services that connect to the exact same database, but use Sequelize instead of Kysely, and they work just fine. They both use Tedious as the driver, so you would assume that they would also connect identically as well, but for some reason they do not.

We have also verified that we don't seem to have database mirroring enabled, but are somehow still getting sent a `DatabaseMirroringPartner` event, which is causing the unhandled exception to be thrown.

Something of note may be that it seems like it connects to the database and runs the first query, but directly after returning the result of the query it throws the unexpected token error.

**Expected behavior**
Query runs and returns values

**Actual behavior**
Query runs, returns the result then immediately crashes the process

**Error message/stack trace**
```
/Users/user/Documents/service-name/node_modules/tedious/src/token/handler.ts:140
    throw new UnexpectedTokenError(this, token);
          ^
UnexpectedTokenError: Unexpected token `ENVCHANGE` in `RequestTokenHandler`
    at RequestTokenHandler.onDatabaseMirroringPartner (/Users/user/Documents/service-name/node_modules/tedious/src/token/handler.ts:140:11)
```

**Any other details that can be helpful**
N/A

nick-w-nick:
This PR is to fix #1669, where the library throws an unhandled exception if it encounters an `ENVCHANGE` event of type `DATABASE_MIRRORING_PARTNER`.

This behavior isn't explicitly documented, but there have been a few issues brought up regarding this functionality in the past, and it has been stated that tedious does not plan to support failover partners. While that may be the case, this event being emitted by SQL Server is causing Tedious to crash completely rather than simply ignoring it as you would expect.

The changes in this PR are to create the `databaseMirroringPartner` event emitter and register the necessary handler methods in the associated token handler classes, to prevent the default behavior of throwing an `Unexpected token` error.

To clarify, this does not implement mirroring functionality or anything of the sort, it simply updates the default behavior to allow the event to pass through and be ignored, rather than unnecessarily throwing an unhandled exception.


src/token/handler.ts
```diff
@@ -355,10 +363,6 @@ export class Login7TokenHandler extends TokenHandler {
   onPacketSizeChange(token: PacketSizeEnvChangeToken) {
     this.connection.messageIo.packetSize(token.newValue);
   }
-
-  onDatabaseMirroringPartner(token: DatabaseMirroringPartnerEnvChangeToken) {
-    // Do nothing
-  }
 }
```
nick-w-nick, line 366: I moved this handler to be right below the `onDatabaseChange` event handler for better ordering and to keep the event next to all of the other `ENVCHANGE` events.

src/token/handler.ts
```diff
@@ -166,6 +166,10 @@ export class InitialSqlTokenHandler extends TokenHandler {
     this.connection.emit('databaseChange', token.newValue);
   }
 
+  onDatabaseMirroringPartner(token: DatabaseMirroringPartnerEnvChangeToken) {
+    this.connection.emit('databaseMirroringPartner', token.newValue);
+  }
+
```
arthurschreiber, line 172: Does this mean that the database mirroring partner event is sent by the server when the initial SQL setup query is sent by `tedious`?

This is... surprising. 🤔 

The MS-TDS specification states that:

> Type 13 (Database Mirroring) is sent in response to a LOGIN7 message whenever connection is requested to a database that it is being served as primary in real-time log shipping. The ENVCHANGE stream reflects the name of the partner node of the database that is being log shipped.

I'm wondering if there's some other issue in the sequencing of how we handle events and which state the handling happens in. Because based just on the specification, this event should not be send by the server in a response to the initial sql query we send.

nick-w-nick: @arthurschreiber You're actually 100% correct, I mistakenly had added the event emitters in each of them during debugging but hadn't removed them to determine if it was a specific handler or not.

I've removed the unnecessary event handler from the `InitialSqlTokenHandler` class and verified that the event is still handled correctly.

Thanks!

src/token/handler.ts
```diff
@@ -406,6 +406,10 @@ export class RequestTokenHandler extends TokenHandler {
     this.connection.emit('databaseChange', token.newValue);
   }
 
+  onDatabaseMirroringPartner(token: DatabaseMirroringPartnerEnvChangeToken) {
+    this.connection.emit('databaseMirroringPartner', token.newValue);
+  }
```
arthurschreiber, line 411: Similar to my previous question - the `ENVCHANGE` token for this is only supposed to be sent back by the server as a response to `Login7`. This adds a handler for general request response. I don't think this token should ever be received here. 🤔

nick-w-nick: @arthurschreiber Ah, I had removed that in a commit that I hadn't yet pushed up. Just pushed the commit, the event is now only being handled within the `Login7` token handler. Good catch!

nick-w-nick:
@arthurschreiber Just curious if this can be merged anytime soon?

I am still using a custom fork of this package to support my MSSQL connections, and would greatly appreciate being able to switch back to the official version once this is merged in.

Let me know if you need any changes to the PR in order to proceed.

Thanks!

PR #1679, opened 2025-02-28
fix: catch errors from identity class construction

dhensby:
see https://github.com/tediousjs/node-mssql/issues/1727

If the `@azure/identity` constructors throw an error, the error is not caught and the process crashes. This attempts to fix the problem.

**Before submitting a PR :**

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3538,34 +3538,41 @@ class Connection extends EventEmitter {
         /** Instance of the token credential to use to authenticate to the resource. */
         let credentials: TokenCredential;
 
-        switch (authentication.type) {
-          case 'token-credential':
-            credentials = authentication.options.credential;
-            break;
-          case 'azure-active-directory-password':
-            credentials = new UsernamePasswordCredential(
-              authentication.options.tenantId ?? 'common',
-              authentication.options.clientId,
-              authentication.options.userName,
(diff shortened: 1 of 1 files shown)


src/connection.ts
```diff
@@ -3538,34 +3538,41 @@ class Connection extends EventEmitter {
...
+            case 'azure-active-directory-service-principal-secret':
+              credentials = new ClientSecretCredential(
+                  authentication.options.tenantId,
+                  authentication.options.clientId,
+                  authentication.options.clientSecret
+              );
+              break;
+          }
+        } catch (err) {
+          signal.throwIfAborted();
```
arthurschreiber, line 3572: Is this required? I don't see any async processing happening between here and the previous signal check? 🤔

dhensby: I have no idea - I was just copying the same logic for the other catch incase it was important.

dhensby:
It actually seems this has been fixed by the new abort signal handling / v19.0.0. It's a problem in v18, which presumably is not taking fixes anymore?

dhensby:
OK - It's not fixed in 19.0.0, it's fixed in `master` in some unreleased refactor work. I think specifically #1663

So this is broken in 18.x and 19.x, and will be fixed when there's a new release of v19

Issue #1682, opened 2025-03-20
[FEATURE REQUEST] JSON Data Type support for Stored Procedure parameters


matthewerwin:
**Is your feature request related to a problem? If so, please give a short summary of the problem and how the feature would resolve it**

All versions of MSSQL now support JSON data type. There is no TYPE.JSON in Tedious or support to leverage this and pass JSON objects using this native type instead of as varchar/nvarchar.

**Describe the preferred solution**
Addition of JSON data type for the parameters that correctly serializes a JSON object to SQL Server when calling a stored procedure.

**Describe alternatives you've considered**
I've considered writing the PR myself but don't have any experience with Tedious code-base. The other alternatives are to continue to use varchar/nvarchar types and manually serialize and re-parse within an SP at the server...none of these are great options.

matthewerwin:
For anyone that's curious, I wrote up the below example of how support for the Json data type on stored procedures could be accomplished without waiting on Tedious team to integrate [this PR](https://github.com/tediousjs/tedious/pull/1683). This will work with the [node-mssql](https://github.com/tediousjs/node-mssql) library as an interim solution.


    const SqlJson = {
        type: {
            type: {
                id: 0xF4,
                type: 'JSON',
                name: 'Json',

                declaration: (parameter) => {
                    return 'json';
                },
                resolveLength: (parameter) => {
                    if( (parameter.value instanceof String || typeof (parameter.value) == "string") )
                        return parameter.value.length;
                    else return JSON.stringify(parameter.value).length;
                },

                generateTypeInfo: (parameter) => {
                    const buffer = Buffer.alloc(1);
                    buffer.writeUInt8(0xF4, 0);
                    return buffer;
                },

                generateParameterLength: (parameter, options) => {
                    const MAX_NULL_LENGTH = Buffer.from([0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF]);
                    const UNKNOWN_PLP_LEN = Buffer.from([0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]);
                    const value = parameter.value as Buffer | null;

                    if( value == null ) {
                        return MAX_NULL_LENGTH;
                    }
                    return UNKNOWN_PLP_LEN;
                },

                generateParameterData: function* (parameter, options) {
                    const PLP_TERMINATOR = Buffer.from([0x00, 0x00, 0x00, 0x00]);
[...]

PR #1683, opened 2025-03-21
feat: Json data type

matthewerwin:
Tested that this is working with [mssql](https://github.com/tediousjs/node-mssq) against Azure Sql Server. 

Notes:
- validates JSON if parameter value is a _string_ 
- encodes as JSON if parameter value is an _object_
- passes 'null' to the server if parameter value is null
- The inclusion of this new data type does not create any breaking changes to the codebase

I used the following logic to deduce the implementation:

1. Referenced 0xF4 and PLP based on [SqlClient](https://github.com/dotnet/SqlClient)
2. That no parameter collation bytes are sent because [Microsoft](https://learn.microsoft.com/en-us/sql/t-sql/data-types/json-data-type?view=azuresqldb-current) is very clear that JSON collation is fixed to Latin1_General_100_BIN2_UTF8 to match JSON Spec
3. That the protocol only needed 0xf4 id prefix and no length bytes since its a newer data type which doesn't have a legacy <= 8000 implementation like varchar did (and the 'json' type has no length setting in SQL)
4. That due to the fixed collation that UTF-8 encoding in validate() is sufficient

I did not use the iconv-lite library b/c it seemed like unnecessary overhead vs the direct Buffer.from( ) support in Node. 

This resolves this [Feature Request](https://github.com/tediousjs/tedious/issues/1682) I created earlier before deciding to address with a PR after a little digging.


matthewerwin:
@arthurschreiber @dhensby -- is there a plan to roll any updates out to Tedious in the near future? Noticed there has been no movement on the repo in 6+ months. 

I had a couple in-flight items including this PR and would like to do what I can to help if needed. Another one was 

https://github.com/tediousjs/node-mssql/issues/1727

and it seems a deprecation on Microsoft's side related to that as well:
https://github.com/tediousjs/tedious/issues/1684


Generally it seems some maintenance things are accumulating without any regular interval releases from the Tedious team. Happy to help organize and move some things through for a release including reviewing other PR's if it's a matter of limited time to organize/test/review.

matthewerwin:
@arthurschreiber you bet -- I'll make a pass through the open PR's & tickets over the next week and see what stands out.

peterbud:
Is there a plan to merge this PR? It would be highly appreciated my many users I believe

src/data-types/json.ts
```diff
@@ -0,0 +1,59 @@
...
+const Json: DataType = {
+  id: 0xF4,
+  type: 'JSON',
+  name: 'Json',
+
+  declaration: (parameter) => {
+    return 'json';
+  },
+  resolveLength: (parameter) => {
+    if ((parameter.value instanceof String || typeof (parameter.value) == 'string'))
```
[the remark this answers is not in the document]
matthewerwin: No -- this eliminates an edge-case where anytime you're dealing with an actual String object the conditional fails.

src/data-types/json.ts
```diff
@@ -0,0 +1,59 @@
...
+    yield value;
+    yield PLP_TERMINATOR;
+  },
+
+  validate: (value, collation): Buffer | null => {
+    if (value == null) {
+      return null;
+    }
+
+    if (!(value instanceof String || typeof (value) == 'string'))
```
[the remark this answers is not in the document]
matthewerwin: Again this creates a bug on an edge case. For the library to robustly handle a string of any type this additional check must be present.

src/data-types/json.ts
```diff
@@ -0,0 +1,59 @@
...
+  type: 'JSON',
+  name: 'Json',
+
+  declaration: (parameter) => {
+    return 'json';
+  },
+  resolveLength: (parameter) => {
+    if ((parameter.value instanceof String || typeof (parameter.value) == 'string'))
+      return parameter.value.length;
+    else return JSON.stringify(parameter.value).length;
```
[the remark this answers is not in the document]
matthewerwin: This also isn't accurate. To my knowledge this is sent by Tedious as nvarchar at serialization time -- the correct length is as indicated. @arthurschreiber thoughts?

arthurschreiber:
Thanks for the work on this @matthewerwin, and apologies again for the long silence — I've gone through this in depth now. The core wire encoding for *sending* a `json` parameter looks right and matches what SqlClient does (single-byte `0xF4` TYPE_INFO, no collation, PLP-encoded UTF-8, `0xFF…FF` for PLP NULL). A few things need to happen before this can ship, and since I have push access to your branch I'm happy to push the fixes myself rather than round-trip through review comments — just say if you'd rather take any of them yourself.

Issues found, roughly by severity:

1. **No `JSONSUPPORT` feature-extension negotiation.** Per MS-TDS, JSON support is enabled via a LOGIN7 FeatureExt handshake (feature id `0x0D`) that the server acknowledges in FEATUREEXTACK — that's how SqlClient and mssql-jdbc gate the type. As written we send `0xF4` unconditionally: against SQL Server ≤ 2022 that surfaces as a cryptic msg 8016 (`Data type 0xF4 is unknown`) protocol error, and against SQL 2025/Azure it works only because the server currently tolerates unnegotiated writes. I'd like to add the FeatureExt (the structure from #1718 makes this straightforward) and fail client-side with a clear error when the server doesn't ack it.

2. **No read path.** `metadata-parser.ts` / `value-parser.ts` don't handle `0xF4`, so registering `Json` in the `TYPE` map only changes the error message when a server sends a json column (`Unrecognised type Json` instead of `Unrecognised data type 0xF4`). Today that's masked because an un-negotiating client gets json columns downcast to varchar — but the moment we negotiate `JSONSUPPORT` (point 1), the server will send `0xF4` in result metadata and every `SELECT` of a json column would kill the token stream. So 1 and 2 have to land together. The values are plain PLP UTF-8, so the read side is close to the existing max-varchar/`Xml` path.

3. **Malformed PLP stream for empty values.** `generateParameterData` returns without yielding anything for a zero-length value, but `generateParameterLength` has already emitted the unknown-length PLP header, which obligates the `00 00 00 00` terminator. It's currently unreachable (empty string fails `JSON.parse` in `validate`), but it's a latent protocol bug — the terminator should always be yielded, matching the nvarchar implementation.

4. **`validate()` edge cases.** A `Buffer` value silently takes the `JSON.stringify` branch and inserts `{"type":"Buffer","data":[…]}`; a function/symbol makes `JSON.stringify` return `undefined` and produces a confusing `Buffer.from(undefined)` TypeError. These should be explicit `TypeError`s like our other types throw. Related: `resolveLength` runs *after* `validate()` has replaced the value with a Buffer, so it returns `JSON.stringify(buffer).length` — nothing consumes it for this type, so I'd just remove it rather than keep a wrong implementation.

5. **Tests.** We'll need unit coverage in `test/unit/data-type.ts` (buffer-level assertions for `generateTypeInfo` / `generateParameterLength` / `generateParameterData`, null and multi-byte UTF-8 cases) plus integration coverage once the negotiation and read path exist.

Plus some minor style alignment (braced conditionals, `===`, dropping the boxed-`String` acceptance — no other type accepts `new String(…)`).

I'll start pushing to this branch shortly. Thanks again for doing the protocol digging — the SqlClient/MS-TDS references in the description made verifying this much easier.

arthurschreiber:
All of the above is now resolved on this branch:

- 5793724 — merged latest `master` (clean, no conflicts; picks up the Node ≥ 22 requirement, the TypeScript test conversion, and the FeatureExt rework from #1718).
- 0630b26 — issues 1 + 2: `JSONSUPPORT` is requested at login, the FEATUREEXTACK is tracked on the connection, and a `TYPES.JSON` parameter sent to a server that didn't ack now fails client-side with a descriptive `RequestError` (`EJSONNOTSUPPORTED`). The read path handles `json` COLMETADATA (no length/collation bytes, verified against SqlClient's implementation) and PLP UTF-8 values in ROW, NBCROW and RETURNVALUE tokens, returned as strings.
- 127dd31 — issue 3: the PLP terminator is now always emitted for non-null values.
- d53f039 — issue 4: explicit `TypeError`s for Buffers and non-serializable values; `resolveLength` removed. The style nits are folded into these commits as well.
- 527cea2 — issue 5: unit coverage for the wire encoding, validation, ack parsing and token parsing (incl. multi-byte UTF-8), plus a new `test/integration/json-test.ts` that exercises both server capabilities — round-trips on servers that ack `JSONSUPPORT`, and the `EJSONNOTSUPPORTED` error on servers that don't.
- 091ed6f — one thing not in the list above: the type is exposed as `TYPES.JSON` rather than `TYPES.Json`, matching how JavaScript itself names it (the global `JSON` object). Since the type has never been released, this isn't a breaking change.

CI status: lint, CodeQL and the unit suite pass. The `claude-review` check fails for infrastructure reasons only (fork PRs get no OIDC token / API key), and the Azure SQL jobs skip on fork PRs for the same secrets reason.

One remaining caveat: the integration tests lint and typecheck, but I haven't been able to run them against a real SQL Server 2025 / Azure SQL instance. @matthewerwin since you have an Azure SQL setup that supports `json`, it would be a great help if you could run `test/integration/json-test.ts` against it (or re-test your original mssql scenario on this branch) to confirm the negotiation works end-to-end.

matthewerwin:
@arthurschreiber thanks for the teamwork effort on this! I’ll have a closer look (and run all tests) when I return from Miami (Tues 7/21) but looks solid based on your write up.

arthurschreiber:
A few more things landed on this branch since the last summary, all verified against SQL Server 2025 RTM-CU7 (17.0.4065.4):

- 2ccb056 — **`json` columns in bulk loads.** The server rejects the `json` data type (`0xF4`) in bulk load column metadata (`Invalid column type from bcp client`), even when `JSONSUPPORT` was negotiated — on RC1 and RTM-CU7 alike. Instead, `varchar(max)` type info is substituted in the COLMETADATA and the server converts the PLP-encoded UTF-8 values based on the `json` column type in the `insert bulk` statement. This matches SqlClient's `SqlBulkCopy`, which performs the same substitution (and whose issue tracker notes that SQL Server 2025 releases before CU5 have server-side bugs in this conversion path — dropped rows, broken NULL handling: dotnet/SqlClient#3000, dotnet/SqlClient#3737).
- 38fb742 — **`json` columns in TVPs.** Sending `0xF4` in TVP column metadata is worse than in bulk loads: the server kills the session. The same `varchar(max)` substitution is applied here, with one important difference: the collation matters. With a zeroed collation the server decodes the column data using its default codepage, garbling multi-byte UTF-8. The substituted type info therefore carries `Latin1_General_100_BIN2_UTF8` — the collation the `json` type is fixed to. (SqlClient has no json-in-TVP support at all; its TVP type info writer has no case for it.)
- c4aea72 — the bulk load substitution now sends the same collation instead of zeroed bytes, keeping the two substitutions identical. The server ignores the collation there (verified in a database with a non-UTF-8 default collation), but sending the truthful value keeps things consistent. This is a deliberate, documented deviation from SqlClient, which zeroes these bytes.
- 8f19878 — **stricter `JSONSUPPORT` ack validation.** A FEATUREEXTACK with a version other than 1 (or a malformed data length) previously degraded silently to "server does not support json". That's unsound: the server considers the negotiation complete and will use the acknowledged version's wire format, which this client would misparse as v1 UTF-8 text. The raw ack data is now carried on the token (like `fedAuth`) and the `Login7TokenHandler` fails the login with a `ConnectionError` for anything other than a clean version-1 ack — matching SqlClient.
- f715b4b — the `EJSONNOTSUPPORTED` error now distinguishes its two root causes: a server that didn't ack `JSONSUPPORT` vs. a configured `tdsVersion` below `7_4` (in which case the negotiation was never attempted and the message points at the config rather than blaming the server).
- 2a8b989, 937e6c9 — unit coverage for the RETURNVALUE read path (the last uncovered json path, incl. json output parameters against a live server) and style alignment of `json.ts` with the other data type files.

Test status: the full unit suite, lint and `tsc` pass; the integration suite in `test/integration/json-test.ts` now covers result sets, parameters, output parameters, bulk loads and TVPs, and passes against SQL Server 2025 RTM-CU7 on TDS 7.4 (plus the `EJSONNOTSUPPORTED` path on TDS 7.3).

saurabh500:
@matthewerwin Thanks for this change. We are hoping to consume the json datatype support as well.

@arthurschreiber any timeline for releasing this change?

matthewerwin:
@arthurschreiber I tested & pushed up a change to package-lock.json and package.json to address security vulnerabilities. I looked at rebasing against master but it appears you've committed some refactoring changes to how the interface looks on DataType (resolveLength, and writeValue).

Rather than fumbling through it and possibly messing something up in how you have that working I'll leave the final rebase + merge to you. Let me know if you need a re-test.

matthewerwin:
@arthurschreiber I tested against Azure (including all integration tests), migrated the entire branch to latest from/master, & pushed up a change to package-lock.json and package.json to address security vulnerabilities.

Everything looks tight. Since this included a re-base and roll forward to the new methods on DataType to match master can you give it a solid test on your side just to confirm? I passed everything through Opus-5.5 Max as a double-check, but saw you were using Fable so let me know if anything comes back on your review.

Thanks a ton for your work getting this out!

src/token/handler.ts
```diff
@@ -342,13 +342,25 @@ export class Login7TokenHandler extends TokenHandler {
...
 
+    if (token.jsonSupport !== undefined) {
+      if (token.jsonSupport.length === 1 && token.jsonSupport[0] === 0x01) {
+        this.connection.serverSupportsJson = true;
+      } else {
+        // The server acknowledged a JSON version this client did not offer.
+        // It will use that version's wire format for `json` values, which
+        // this client can not safely parse - fail the login instead of
+        // continuing on a connection whose data can not be trusted.
+        this.loginError = new ConnectionError('Received invalid JSON support acknowledgement');
```
arthurschreiber: 🔴 **This error never fails the login.** Per MS-TDS, `FEATUREEXTACK` comes *after* `LOGINACK`, so `loginAckReceived` is already `true` when this runs. All three `performSentLogin7With*` paths in `connection.ts` check `if (handler.loginAckReceived) return handler.routingData;` **before** `else if (handler.loginError)` (e.g. `connection.ts:3558-3561`, and the same pattern around `3594` and `3646`). So `loginError` is silently dropped.

We reproduced this with a mock TDS server that sends LOGINACK(7.4) → FEATUREEXTACK{0x0D, data} → DONE, using data `[0x02]`, `[]`, `[0x01, 0x01]` and `[0x00]`. In every case `connect()` succeeds with no error, and `serverSupportsJson === false`.

Outgoing JSON params are still blocked by `EJSONNOTSUPPORTED`. But the server believes it negotiated JSON vN and may send `0xF4` values in a vN wire format, which this client then decodes as v1. That is exactly what this commit meant to prevent.

The new `handler-test.ts` case only checks `handler.loginError`, not what `connect()` returns, so it doesn't catch this.

Suggested fix: check `handler.loginError` before `handler.loginAckReceived` in the three login paths, or clear `loginAckReceived` when an error is set here. Then add a connection-level test with a mock server; the pattern in `test/unit/tds71-deprecation-test.ts` works for this. (The fedAuth ack errors in this same function have the same ordering problem on master today.)

---

src/data-types/json.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (typeof value === 'string') {
+      JSON.parse(value);
+      return Buffer.from(value, 'utf8');
+    }
+
+    if (Buffer.isBuffer(value)) {
+      throw new TypeError('Invalid JSON value.');
+    }
+
+    const serialized = JSON.stringify(value);
```
arthurschreiber: 🔴 **Many non-plain-object values are silently turned into the wrong JSON here.**

**Streams / async iterables.** Since #1777 the other max types (`VarChar`, `NVarChar`) accept async iterables and stream them via `writePlpStream`. This type has no `isAsyncIterable` branch, so these values reach `JSON.stringify`:
- `(async function*(){ yield '{"a":1}' })()` is sent as `{}`
- `Readable.from(['{}'])` is sent as `{"_events":{},"_readableState":{...` (the stream's internal state)

A user switching from `NVarChar(max)` streaming to `TYPES.JSON` would store garbage and get no error.

**Other values that are silently changed:**

| Input | Sent as |
|---|---|
| `Uint8Array` | `{"0":1,"1":2}` |
| `Map`, `Set` | `{}` |
| `{ a: undefined }` | `{}` |
| `NaN`, `Infinity` | JSON `null`, not SQL NULL |
| `new String('x')` | `"x"`, while the primitive `'x'` is rejected because it is parsed as JSON text |

Suggestion: either support async iterables the way `varchar.ts` does (`writePlpStream(buffer, v, s => Buffer.from(s, 'utf8'))`), or reject them. Reject at least streams, typed arrays and `ArrayBuffer`, and consider rejecting non-finite numbers. Also document the conversion rules in the data type table:
- strings are JSON *text*, not JSON string values
- objects go through `JSON.stringify`
- values read back are always strings

Two smaller notes on this function:
- `JSON.parse` on every string builds and throws away the whole object tree. On a 61 MB string it took ~560 ms. The server validates anyway, so consider dropping the check or making it optional.
- V8's `SyntaxError` messages quote short inputs, e.g. `"secret-token-abc" is not valid JSON`, and that text ends up in the `EPARAM` error and the debug log. A fixed message with the original error as `cause` avoids leaking parameter values.

---

test/integration/json-test.ts
```diff
@@ -0,0 +1,280 @@
...
+      connection.on('end', done);
+      connection.close();
+    } else {
+      done();
+    }
+  });
+
+  describe('on servers that support the json data type', function() {
+    beforeEach(function() {
+      if (!connection.serverSupportsJson) {
```
arthurschreiber: 🔴 **These tests skip based on the flag they are testing.** Both this suite and the negative suite at line 255 decide whether to run from `connection.serverSupportsJson`, which is the value under test.

If JSONSUPPORT negotiation breaks, the flag is never set. Then every positive test skips, and the negative `EJSONNOTSUPPORTED` test runs and passes because the check is client-side. That happens even on the SQL Server 2025 CI job, so a real regression shows up green.

Suggestion: decide whether the server supports JSON independently of the flag, then assert that `serverSupportsJson` matches and skip only based on that independent signal. Options:
- `SERVERPROPERTY('ProductMajorVersion') >= 17`
- `SERVERPROPERTY('EngineEdition')` for Azure
- a probe like `SELECT CAST('{}' AS json)`

Related: the SELECT tests (e.g. lines 53, 67, 83) only assert inside the `row` handler and never check that a row arrived, so they would pass with zero rows. Counting rows and asserting in the completion callback would close that, like the output-parameter tests already do with `returnValueReceived`.

---

package.json
arthurschreiber: 🟡 **Unrelated dependency churn.** Commit d446467 ("fix: npm audit fix") bumps **mocha ^11.7.5 → ^12.0.1**, a major version, which a plain `npm audit fix` wouldn't do without `--force`. It also churns ~525 lines of `package-lock.json`: chokidar 5, glob 13, workerpool 10, yargs removed, and so on. Other details:
- mocha 12 now pairs with `@types/mocha` ^10
- mocha 12 requires Node `^20.19.0 || >=22.12.0`

It's all devDependencies and unrelated to JSON support. Could it move to a separate `chore(deps-dev)` PR so this one stays focused? The same goes for the message-io flaky-test fix in 7de0cc3, which is a real fix but is also unrelated.

---

arthurschreiber (commented):
Thanks for pushing this forward! The TDS side looks solid. The JSONSUPPORT request (`0x0D`, v1, TDS 7.4 only), COLMETADATA (type byte only), PLP reads (null, empty, and UTF-8 split across chunks, verified with a mock server), RPC encoding, and the varchar(max) + `Latin1_General_100_BIN2_UTF8` substitution for bulk load and TVP all check out. Lint, typecheck, build and the unit tests pass locally (560 passing). Everything from the July review is addressed.

**To fix before merge** (details inline):
1. **An invalid JSON feature ack doesn't fail the login.** `loginAckReceived` is checked before `loginError` in the login paths, so the error from `onFeatureExtAck` is dropped.
2. **`validate()` silently mis-serializes values.** Async iterables and streams, typed arrays, `Map`/`Set`, `NaN` and so on all go through `JSON.stringify` without error. Streaming, which the sibling max types support, becomes `{}`.
3. **The integration tests skip based on `serverSupportsJson` itself,** so a broken negotiation would show up as all-skipped rather than failing.
4. **Unrelated changes** (mocha 12 major bump plus lockfile churn, and the message-io test fix) would be better as separate PRs.

**Should fix / smaller items:**
- The PR description says `null` sends JSON `'null'`, but the code (correctly) sends SQL NULL. Worth updating, and documenting in the data-type table that:
  - strings are treated as JSON *text*
  - objects are `JSON.stringify`'d
  - values read back are always strings
- **Missing unit tests:**
  - multi-chunk, unknown-length PLP reads with a multibyte character split across chunks and across `parseTokens` buffers
  - the two `EJSONNOTSUPPORTED` branches in `makeRequest` (only an integration test covers them today)
  - the exact JSON feature bytes in the login packet (`0d 01000000 01`)
  - `validate` edge cases: primitives, `''`, `BigInt`, `Date`
- `connection.prepare()` isn't covered by the `EJSONNOTSUPPORTED` check, so it fails server-side with "Cannot find data type json".
- The `EJSONNOTSUPPORTED` message calls the *negotiated* `tdsVersion` the "configured" one, because `onLoginAck` overwrites `config.options.tdsVersion`.
- Lone surrogates in strings are silently replaced with U+FFFD by `Buffer.from(s, 'utf8')`.
- `serverSupportsJson` is marked `@private` but emitted as public in the `.d.ts`, and the tests depend on it.

Also: Node CI and CodeQL haven't run on `7de0cc3` yet (they need maintainer approval because this is a fork PR). Nobody has re-run the integration tests since the port to the `writeTypeInfo`/`writeValue` contract.

---

matthewerwin:
@arthurschreiber great feedback. I pushed the changes.

I disagree with Claude's recommendations on separating the Mocha update + moving the improved test to another PR. 

Mocha is a test platform and dev dependency and second, migrating code into /master that has known high-vulnerabilities and could end up released isn't great. While, Mocha 12 does mean a Node 22 lower limit to 22.12 (I added >=22.12 since Mocha raised the minimum because 20.19 and 22.12 are the first releases with the require(esm) behavior Mocha 12 relies on) that still meaningfully supports Node v22.

Fixing the split-packet issue on a test doesn't affect production code and is meaningful to this branch (and master) not having intermittent test failures over network behavior.

All other requested changes were implemented.

src/token/handler.ts
matthewerwin: Done.

src/data-types/json.ts
matthewerwin: Done.

package.json
matthewerwin: Mocha is a test platform and dev dependency and second, migrating code into /master that has known high-vulnerabilities and could end up released isn't great. While, Mocha 12 does mean a Node 22 lower limit to 22.12 (I added >=22.12 since Mocha raised the minimum because 20.19 and 22.12 are the first releases with the require(esm) behavior Mocha 12 relies on) that still meaningfully supports Node v22.

test/integration/json-test.ts
matthewerwin: Done.

PR #1695, opened 2025-10-21
fix: login7 against Fabric

David-Engel:
Follow on to #1668 to try to address test failures while still supporting Fabric connections.

I believe there were several issues causing problems connecting to Fabric and making fixes for Fabric difficult;

1. Fabric wants (wanted? I think Fabric may have relaxed this now) TraceID in PreLogin.
2. Fabric wasn't forgiving of a technically incorrect Login7 payload
3. The Login7 payload should never have feature extensions when negotiating TDS < 7.4
4. Some of the Login7 payload tests were just wrong.
    a. Incorrectly expecting feature ext data for TDS 7.2
    b. Not accounting for UTF-8 feature ext data in TDS 7.4

### DIFF
--- a/src/login7-payload.ts
+++ b/src/login7-payload.ts
@@ -72,7 +72,6 @@ interface Options {
   connectionId: number;
   clientTimeZone: number;
   clientLcid: number;
-  isFabric: boolean;
 }
 
 /*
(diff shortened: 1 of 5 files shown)


src/login7-payload.ts
```diff
@@ -412,16 +417,14 @@ class Login7Payload {
       }
     }
 
-    if (this.tdsVersion >= versions['7_4']) {
```
David-Engel: This condition is moved outside of the buildFeatureExt method because the entire thing should only be called for TDS >= 7.4.

Issue #1686, opened 2025-05-21
[CHORE] Update package-lock.json in /benchmarks


RyanJohannesBland:
Package-lock.json in /benchmarks is more than 2 years out of date. This causes it to still have @azure/identity installed at "^2.0.4", which has know vulnerabilities. These vulnerabilities are caught by services such as AWS Inspector. To upgrade, simply recreate package-lock.json based on the current directory.

I attempted to create a PR with these changes, but I do not have access to push a new branch.

sjgutta:
I would like to second this issue. This package-lock file has been sitting stale for a long time and trips up vulnerabilities in a bunch of different vulnerability scanning services when there isn't actually a relevant vulnerability. It would be great to be able to just install tedious without the benchmarks files included but at the very least maintenance of the package lock file would be a huge improvement. It has been a major time suck for us every time it trips up a vulnerability alert that isn't relevant

Thank you!

PGodward:
I am starting to think that there is no one maintaining Tedious anymore...

ZsoltPath:
Definitely feels like that. There was no release since last August and 51 PRs are waiting. :/

PR #1697, opened 2025-10-24
chore: add benchmarks to npmignore

TPVallancourt:
https://github.com/tediousjs/tedious/issues/1686

Benchmarks (like tests & examples) do not need to be published with the package. Add benchmarks to `.npmignore`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,3 +5,4 @@ scripts/
 todo
 .*
 /scratch
+benchmarks/


arthurschreiber:
I changed this from `fix:` to `chore:` because I don't think this warrants publishing a new release on it's own.

This is a good change, thank you! ❤️

PR #1713, opened 2025-12-19
test: convert unit and token parser tests to TypeScript

arthurschreiber:
Convert JavaScript test files to TypeScript:
- All unit test files (26 files)
- Token parser test files (12 files)
- Tracking-buffer test file
- Performance test file

Changes include:
- Replace require() with ES module imports
- Add type annotations where appropriate
- Use proper TypeScript patterns for chai assertions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "@types/chai": "^4.3.20",
         "@types/depd": "^1.1.37",
         "@types/lru-cache": "^5.1.1",
+        "@types/mitm": "^1.3.8",
         "@types/mocha": "^10.0.10",
         "@types/sinon": "^21.0.0",
         "@types/sprintf-js": "^1.1.4",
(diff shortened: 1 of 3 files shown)


test/unit/debug-test.ts
```diff
@@ -18,12 +19,12 @@ describe('Packet Tests', function() {
 
     const debug = new Debug({ packet: true });
 
-    debug.on('debug', function(text) {
+    debug.on('debug', function(text: string) {
```
arthurschreiber: @claude Do we need the explicit type annotation here and in other places? If not, I think we should remove them.

test/integration/parameterised-statements-test.ts
```diff
@@ -23,31 +21,28 @@ function getConfig() {
...
+
+function execSql(
+  done: Mocha.Done,
+  type: DataType,
+  value: unknown,
+  tdsVersion?: TdsVersion,
+  options?: import('../../src/request').ParameterOptions | null,
+  expectedValue?: unknown,
+  cast?: boolean,
+  connectionOptions?: import('../../src/connection').ConnectionOptions
```
arthurschreiber: @claude Can you change these inline imports to be top level imports instead?

PR #1738, opened 2026-06-11
fix: keep message stream aligned when canceling with a disabled cancel timeout

arthurschreiber:
This PR fixes two related defects in how a connection determines the outcome of a canceled request. Both stem from the cancellation result being inferred indirectly from unrelated state, rather than tracked explicitly.

## Canceling with `cancelTimeout: 0` corrupts the message stream

When a request is canceled after its request message was fully sent, an attention message is sent to the server, and the server responds with two messages: the (possibly cut short) response to the canceled request, followed by a separate message containing the `DONE` token that acknowledges the attention message.

Whether the connection expected that acknowledgement message was decided by checking if a cancel timer exists — using the timer as a proxy for "an attention message was sent". With `cancelTimeout: 0`, no cancel timer is ever created, so after a cancellation:

- the response to the canceled request was processed as a regular response, invoking the request's completion callback **without an error**, and
- the attention acknowledgement message was left unread in the message stream, so the **next** request consumed the leftover acknowledgement as its own response, leaving the message stream permanently misaligned from that point on.

The connection now tracks explicitly whether an attention message was sent for the active request, and the cancel timer goes back to being purely a watchdog.

## `ECANCEL` was threaded through the payload stream teardown

When a request is canceled *before* its request message was fully sent (the message is then terminated with the `IGNORE` bit set), the `ECANCEL` error was only set indirectly: the payload stream was destroyed with the error, and the payload stream's `error` event handler then stored it on the request. That indirection is unreliable:

- Destroying a payload stream that has already ended (and was auto-destroyed) is a no-op that emits no `error` event — the canceled request would then complete reporting **success**.
- The `error` event is deferred until the payload's async iterator settles, so for a payload waiting on further data (e.g. a bulk load with an idle row stream), it fires arbitrarily late.
- The error never actually reached the underlying payload stream: a stream wrapped as a payload is torn down by the async iterator machinery with its own `AbortError`, whether the outer stream is destroyed with an error or not.

The cancel handler now sets the `ECANCEL` error directly and destroys the payload stream without an error; the payload stream's `error` handler is solely responsible for genuine payload errors. Request timeouts are unaffected (`ETIMEOUT` is assigned unconditionally right after the cancellation is initiated and overrides `ECANCEL` as before).

## Testing

- New unit tests against a mock server that follows the protocol's request/response model (one response message per client request, attention acknowledgement as a standalone message):
  - canceling with `cancelTimeout: 0` completes the request with `ECANCEL` *and* the follow-up request receives its own response, proving the stream stays aligned (fails on `master`);
  - canceling before the request message was fully sent completes with `ECANCEL` and sends no attention message (pins behavior that previously depended on stream teardown internals).
- Verified against a real SQL Server 2022 instance: the full integration suite passes, including the request cancellation and bulk load tests.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3766,6 +3766,14 @@ Connection.prototype.STATE = {
         }
 
         const onCancel = () => {
+          // If the request was canceled before the request message was
+          // fully sent, the message was terminated with the `IGNORE` bit
+          // set and no attention message was sent. The server's response
+          // to the ignored message is handled like a regular response.
+          if (!this.attentionSent) {
+            return;
+          }
+
           tokenStreamParser.removeListener('end', onEndOfMessage);
 
           if (this.request instanceof Request && this.request.paused) {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1036,6 +1036,14 @@ class Connection extends EventEmitter {
   declare requestTimer: undefined | NodeJS.Timeout;
 
   /**
+   * Whether an attention message was sent to the server to cancel the
+   * currently active request.
+   *
+   * @private
+   */
+  declare attentionSent: boolean;
+
+  /**
    * @private
    */
   declare _cancelAfterRequestSent: () => void;
@@ -1780,8 +1788,11 @@ class Connection extends EventEmitter {
 
     this.state = this.STATE.INITIALIZED;
 
+    this.attentionSent = false;
+
     this._cancelAfterRequestSent = () => {
       this.messageIo.sendMessage(TYPE.ATTENTION);
+      this.attentionSent = true;
       this.createCancelTimer();
     };
 
@@ -3176,6 +3187,7 @@ class Connection extends EventEmitter {
       }
 
       this.request = request;
+      this.attentionSent = false;
       request.connection! = this;
       request.rowCount! = 0;
       request.rows! = [];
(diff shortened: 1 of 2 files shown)


arthurschreiber:
All four review points are addressed:

**2. Unused `chunks` collections / 3. swallowed server errors / 4. timing comment** — done in 667e65b: the drain loops now use a `drainMessage` helper, mock server errors are passed to `done(err)` instead of being logged, and the `setTimeout(..., 50)` cancel now has a comment explaining that it ensures the cancellation happens via an attention message rather than the `IGNORE` bit.

**1. The missing `attentionSent` check in the `SentClientRequest` cancel handler** — this turned out to be more than a defensive guard: it's a reproducible hang. If a request is canceled while its request message is still being sent *and* the server has already started responding (most easily reached with a bulk load, where the message stays in flight while rows stream and the server can respond early with an error), the handler transitioned to `SentAttention` even though no attention message was sent — and since no attention was sent, no cancel timer was started either, so the connection hung indefinitely.

Fixed in e8b0ea7, with one deviation from the suggested shape: guarding only the `transitionTo` call would still hang, because the handler removes the token stream parser's `end` listener before the transition, so the request could never complete through the regular path either. Instead, the handler returns early when no attention message was sent, leaving the regular response handling fully intact — the response to the ignored message then completes the request with the `ECANCEL` error.

The new test pins this with a bulk load whose row stream never completes (so the message is guaranteed to be in flight) and an explicit client→server synchronization point instead of timing assumptions.

---

arthurschreiber:
Latest review round addressed in ac7ea8e:

**1. Timing-sensitive tests** — both wall-clock delays are gone. The `cancelTimeout: 0` test now cancels when the server signals (via a shared promise) that it received the full request message — which guarantees the message `finish` already fired on the client, so the attention path is always exercised. The bulk load test now cancels from a `columnMetadata` listener, i.e. exactly when the client has *processed* the first response packet, which is the precise precondition for the race the test pins (this is stronger than signaling when the server *wrote* the packet, since the race requires client-side receipt). I also verified the reworked test still hangs with the guard reverted, so the event-driven trigger didn't lose the scenario.

**2. `attentionSent` reset on error paths** — done, but in `cleanupConnection` rather than only `SENT_ATTENTION`'s `socketError` handler: that's the single place every teardown path (including `SENT_CLIENT_REQUEST` socket errors and explicit `close()`) converges and where the in-flight request is already failed, so the invariant now holds unconditionally.

**3. LOGIN7 mock missing trailing `DONE`** — added; all three mock login responses now send `LoginAck + DONE_FINAL`.

**4. `declare attentionSent` uninitialized before `connect()`** — small correction: the flag is initialized to `false` in the constructor (next to the `_cancelAfterRequestSent` definition), not in `connect()`, so it's never `undefined` at runtime.

---

PR #1739, opened 2026-06-14
fix: clear the cancel timer when cleaning up a connection

arthurschreiber:
## Problem

When a request is canceled after its request message was fully sent, an attention message is sent to the server and a cancel timer is armed as a watchdog (`_cancelAfterRequestSent` → `createCancelTimer`). That timer is only ever cleared on the happy path — when the attention acknowledgement is received (`SentAttention`), or when it is replaced by a newer timer.

`cleanupConnection` cleared the request timer but **not** the cancel timer. So if the connection was torn down while a cancel was still in flight — via `close()`, a socket error, or the socket closing — the armed cancel timer was leaked.

It would later fire in the `Final` state, where `dispatchEvent('socketError')` has no handler and instead emits an `error` event (and calls `close()` again) — long after the connection had already emitted `end`. With no remaining `error` listener, this crashes the process with an unhandled `error` event.

## Fix

Clear the cancel timer in `cleanupConnection`, alongside the request timer, so it can never outlive the connection:

```js
cleanupConnection() {
  if (!this.closed) {
    this.clearRequestTimer();
    this.clearCancelTimer();   // ← added
    this.closeConnection();
    ...
```

The legitimate cancel watchdog path is unaffected: there the timer fires (and tears the connection down) *before* cleanup runs, so clearing it during cleanup is a no-op in that case.

## Testing

Adds `test/unit/connection-cancel-timer-test.ts`, a mock-server regression test that logs in, issues a request, cancels it (arming the cancel timer via the attention message), then `close()`s the connection while the cancel is in flight.

Detection is timing-based: tearing down a connection that is waiting on a response surfaces an immediate "connection lost" error, whereas the orphaned cancel timer only fires ~`cancelTimeout` later — so an error observed well after `close()` can only be the leaked timer.

- **Without the fix:** fails (`a spurious error was emitted ~150ms after the connection was closed`)
- **With the fix:** passes
- Full unit suite remains green.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2140,6 +2140,7 @@ class Connection extends EventEmitter {
   cleanupConnection() {
     if (!this.closed) {
       this.clearRequestTimer();
+      this.clearCancelTimer();
       this.closeConnection();
 
       process.nextTick(() => {
--- a/test/unit/connection-cancel-timer-test.ts
+++ b/test/unit/connection-cancel-timer-test.ts
@@ -28,11 +28,27 @@ function buildLoginAckToken(): Buffer {
 
 // Consume and discard a message's data so the iterator can advance to the
 // next incoming message.
+//
+// This drives the async iterator explicitly rather than using `for await...of`
+// because the project's eslint config (`no-unused-vars` + `no-void`) rejects an
+// unused loop binding.
 async function drainMessage(message: AsyncIterable<Buffer>): Promise<void> {
   const iterator = message[Symbol.asyncIterator]();
   while (!(await iterator.next()).done) {
     // Discard each chunk.
   }
+}
+
+// Read the next message the client sends, assert its TDS packet type, and drain
+// its payload. Guards against the message stream ending unexpectedly, so a
+// premature end surfaces as a clear assertion failure rather than a `TypeError`
+// on `message.type`.
+async function expectMessage(iterator: AsyncIterableIterator<Message>, expectedType: number): Promise<void> {
+  const { value: message, done } = await iterator.next();
+  assert.isNotOk(done, 'expected another message from the client, but the stream ended');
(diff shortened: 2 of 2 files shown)


arthurschreiber:
Addressed the test feedback in 99c5da13:

1. **Dead `chunks` variables** — replaced with a `drainMessage` helper that consumes each message via its async iterator without retaining the data.
2. **Timing sensitivity** — bumped `cancelTimeout` to 300ms, so the threshold (150ms) now sits ~150ms above any plausible teardown noise *and* ~150ms below when the leaked timer fires, doubling the margin on both sides.
3. **Request-callback assertion** — the test now asserts the in-flight request completes with an `ECLOSE` error.
4. **`_connections` naming** — renamed the local to `serverConnections`.

Verified: still fails 3/3 without the fix and passes 3/3 with it; full unit suite green; lint clean.

arthurschreiber:
Addressed in 501799ca:

1. **Undefined-message guard** — extracted an `expectMessage(iterator, type)` helper that asserts the stream hasn't ended (`assert.isNotOk(done)` + `assert.isDefined(message)`) before checking `message.type` and draining. This also collapses the six near-identical read/assert/drain blocks.
4. **`drainMessage` style** — switched to `for await (const _ of message) {}` with a targeted `// eslint-disable-next-line @typescript-eslint/no-unused-vars` for the unused loop binding (02857adf).

On the others:
2. **Test duration** — leaving `cancelTimeout` at 300ms. It was deliberately widened from 150ms in the previous review round to increase the separation between teardown noise and the leaked-timer signal; halving it would walk that back.
3 & 5. No change needed (`requestError` works as-is, and the `ECONNRESET` swallow is already explained by the adjacent comment).

Still fails 3/3 without the fix, passes 3/3 with it; full unit suite green (390 passing); lint clean.

PR #1757, opened 2026-07-17
docs: align trustServerCertificate default value in docstring

AxBolduc:
Small change to the JSDoc comment on trustServerCertificate. The default value of this used to be true, but was changed to false in #1464. The comment was never updated and led to some confusion internally so I thought I'd align the docs.


arthurschreiber (approved):

arthurschreiber:
Thank you! ❤️

PR #1764, opened 2026-08-26
fix: cap TLS max send fragment size for encrypt: "strict" connections

### LINKED ISSUE #1182: Azure AD with accessToken Connection lost - read ECONNRESET
I am working on an App that is authenticating user using Azure AD, extracting his `accessToken` and then using this token to connect to the Azure SQL server using below setting.

But unfortunately, I am getting `ESOCKET "Connection lost - read ECONNRESET"` right away,

```
const config = {
          server: 'db-server-name.database.windows.net',
          authentication: {
            type: 'azure-active-directory-access-token',
            options: {
              token: cloudAccessToken
            }
          },
          options: {
            debug: {
              packet: true,
              data: true,
              payload: true,
              token: false,
              log: true
            },
            database: 'DBNAME',
            encrypt: true,
            packetSize: 8192,
            keepAlive:true,
            requestTimeout: 300000,
            connectionTimeout: 32000,
          }
        };

        const connection = new Connection(config);

        connection.on('connect', function (err) {
          if (err) {
            console.log(err);
          }
          executeStatement();
        });

        connection.on('debug', function (text) {
          console.log(text);
        });

        connection.on('error', function (err) {
          console.error(err); // --> this gets trigger with error ESOCKET right away
        });;
```

zinvob:
## Summary

Fixes #1182.

`wrapWithTls` (used only for `encrypt: "strict"`, i.e. TDS 8.0) never calls `setMaxSendFragment` on the resulting TLS socket, unlike the classic `MessageIO#startTls` path (`message-io.ts`), which caps it to the TDS packet size for exactly this reason.

Without that cap, Node's TLS layer is free to coalesce multiple back-to-back TDS packet writes into a single, larger TLS record. A LOGIN7 message that spans more than one TDS packet — most notably any FedAuth login carrying an Entra ID access token, which routinely exceeds one 4KB packet — gets sent as one oversized TLS record. Azure SQL's TDS 8.0 gateway rejects this by resetting the connection (`ECONNRESET`), which is exactly the symptom reported in #1182.

Plain SQL-auth logins are small enough to fit in a single TDS packet, which is why this reads as "FedAuth-specific" until you look at packet counts — it isn't; it's a general multi-packet-under-strict-mode issue that happens to be triggered almost exclusively by FedAuth token payloads in practice.

## Root cause, confirmed

Reproduced directly against a real Azure SQL Database instance:
- `azure-active-directory-access-token` and `azure-active-directory-default` auth under `encrypt: "strict"`: 100% failure (`ECONNRESET`) across tedious 18.6.2, 19.2.1, and 20.0.5.
- Plain SQL auth under `encrypt: "strict"`: works reliably.
- Microsoft's own `Microsoft.Data.SqlClient` (.NET) connecting with the same Entra ID identity + strict encryption against the same server: works reliably — ruling out a server-side/gateway-only limitation.
- Root-caused to the missing `setMaxSendFragment` call by comparing the strict-mode `wrapWithTls` path against the classic `startTls` path, which already sets it (`message-io.ts:107`) for the identical reason.

## Fix

One-line change: call `encryptsocket.setMaxSendFragment(this.config.options.packetSize)` in `wrapWithTls`'s `secureConnect` handler, mirroring `MessageIO#startTls`.

## Verification

- Against real Azure SQL: `azure-active-directory-access-token` and `azure-active-directory-default` under `encrypt: "strict"` go from 100% `ECONNRESET` to 100% successful connection (multiple repeated runs).
- Plain SQL auth: unaffected, still works.
- Added a unit test (`test/unit/connection-test.ts`) that spins up a real TLS server using the existing `localhost.crt`/`.key` fixtures, negotiates the `tds/8.0` ALPN protocol, and asserts (via a `sinon` spy) that the client caps `setMaxSendFragment` to the configured `packetSize` right after the handshake. Verified the test fails without the fix and passes with it.
- Full unit suite: 431/431 passing, no regressions.
- `eslint`/`tsc`: clean.

## Test plan

- [x] `npm test` — 431 passing, 1 pending (pre-existing), 0 failing
- [x] `npx eslint src test` — clean
- [x] Manual verification against a real Azure SQL Database with `encrypt: "strict"` + `azure-active-directory-access-token` / `azure-active-directory-default`

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -2237,7 +2237,22 @@ class Connection extends EventEmitter {
 
       try {
         const onError = reject;
-        const onConnect = () => { resolve(encryptsocket); };
+        const onConnect = () => {
+          // Without this, Node's TLS layer is free to coalesce multiple back-to-back
+          // TDS packet writes into a single, larger TLS record. The non-strict `startTls`
+          // path (message-io.ts) caps max send fragment to the TDS packet size for the
+          // same reason; `wrapWithTls` (TDS 8.0 / encrypt: "strict") must match it.
+          //
+          // Without this fix, a LOGIN7 spanning multiple TDS packets (e.g. any FedAuth
+          // login carrying an Entra ID access token, which routinely exceeds one 4KB
+          // packet) reliably fails with ECONNRESET against Azure SQL once TLS is
+          // negotiated but before LOGINACK — see
+          // https://github.com/tediousjs/tedious/issues/1182. Single-packet logins
+          // (e.g. plain SQL auth with short credentials) are unaffected either way,
+          // which is why this bug reads as "FedAuth-specific" until you look closer.
+          encryptsocket.setMaxSendFragment(this.config.options.packetSize);
+          resolve(encryptsocket);
+        };
 
         encryptsocket.once('error', onError);
         encryptsocket.once('secureConnect', onConnect);
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -173,6 +173,11 @@ const DEFAULT_CONNECT_RETRY_INTERVAL = 500;
  */
 const DEFAULT_PACKET_SIZE = 4 * 1024;
 /**
+ * Node's `tls.TLSSocket#setMaxSendFragment` silently ignores values outside this range.
+ * @private
+ */
+const MAX_TLS_SEND_FRAGMENT_SIZE = 16384;
+/**
  * @private
  */
 const DEFAULT_TEXTSIZE = 2147483647;
@@ -2250,7 +2255,11 @@ class Connection extends EventEmitter {
           // https://github.com/tediousjs/tedious/issues/1182. Single-packet logins
           // (e.g. plain SQL auth with short credentials) are unaffected either way,
           // which is why this bug reads as "FedAuth-specific" until you look closer.
-          encryptsocket.setMaxSendFragment(this.config.options.packetSize);
+          //
+          // `setMaxSendFragment` silently no-ops for values outside 512-16384, so a
+          // `packetSize` above that (e.g. 32767, commonly used for bulk load) has to be
+          // clamped or it would leave the cap unset and reintroduce the bug it fixes.
+          encryptsocket.setMaxSendFragment(Math.min(this.config.options.packetSize, MAX_TLS_SEND_FRAGMENT_SIZE));
           resolve(encryptsocket);
         };
 
@@ -2261,7 +2270,7 @@ class Connection extends EventEmitter {
           return await promise;
         } finally {
           encryptsocket.removeListener('error', onError);
-          encryptsocket.removeListener('connect', onConnect);
+          encryptsocket.removeListener('secureConnect', onConnect);
         }
       } finally {
         signal.removeEventListener('abort', onAbort);
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -173,11 +173,6 @@ const DEFAULT_CONNECT_RETRY_INTERVAL = 500;
  */
 const DEFAULT_PACKET_SIZE = 4 * 1024;
 /**
- * Node's `tls.TLSSocket#setMaxSendFragment` silently ignores values outside this range.
- * @private
- */
-const MAX_TLS_SEND_FRAGMENT_SIZE = 16384;
-/**
  * @private
  */
 const DEFAULT_TEXTSIZE = 2147483647;
@@ -2259,7 +2254,7 @@ class Connection extends EventEmitter {
           // `setMaxSendFragment` silently no-ops for values outside 512-16384, so a
           // `packetSize` above that (e.g. 32767, commonly used for bulk load) has to be
           // clamped or it would leave the cap unset and reintroduce the bug it fixes.
-          encryptsocket.setMaxSendFragment(Math.min(this.config.options.packetSize, MAX_TLS_SEND_FRAGMENT_SIZE));
+          encryptsocket.setMaxSendFragment(Math.min(this.config.options.packetSize, MessageIO.MAX_TLS_SEND_FRAGMENT_SIZE));
           resolve(encryptsocket);
         };
 
--- a/test/unit/connection-test.ts
+++ b/test/unit/connection-test.ts
@@ -1,4 +1,7 @@
 import * as net from 'net';
+import * as tls from 'tls';
+import { readFileSync } from 'fs';
+import * as sinon from 'sinon';
 import { Connection, ConnectionError } from '../../src/tedious';
 import { assert } from 'chai';
 
@@ -73,5 +76,70 @@ describe('Using `strict` encryption', function() {
     connection.on('end', () => {
       done();
     });
+  });
+
+  it('caps the TLS max send fragment size to the configured packet size', function(done) {
+    // Regression test for https://github.com/tediousjs/tedious/issues/1182.
+    //
+    // `wrapWithTls` (used only for `encrypt: "strict"`, i.e. TDS 8.0) must cap the TLS
+    // max send fragment size to the TDS packet size, the same way the classic
+    // `MessageIO#startTls` path already does. Without it, Node's TLS layer is free to
+    // coalesce multiple back-to-back TDS packet writes into a single, larger TLS
+    // record. A LOGIN7 message that spans more than one TDS packet (e.g. any FedAuth
+    // login carrying an Entra ID access token, which routinely exceeds one 4KB packet)
+    // would then be sent as one oversized TLS record, which Azure SQL's TDS 8.0
+    // gateway rejects by resetting the connection.
+    const setMaxSendFragmentSpy = sinon.spy(tls.TLSSocket.prototype, 'setMaxSendFragment');
+    const serverSockets: net.Socket[] = [];
+
+    server.on('connection', (rawSocket) => {
+      serverSockets.push(rawSocket);
+
+      // A minimal TDS 8.0 "server": complete a real TLS handshake advertising the
+      // `tds/8.0` ALPN protocol (as `wrapWithTls` requires), then go silent. We only
+      // care about what the client does immediately after the handshake completes.
+      const tlsSocket = new tls.TLSSocket(rawSocket, {
+        isServer: true,
+        key: readFileSync('./test/fixtures/localhost.key'),
+        cert: readFileSync('./test/fixtures/localhost.crt'),
+        ALPNProtocols: ['tds/8.0']
+      });
+
+      // Ignore any post-handshake errors (the fake server never responds to PRELOGIN).
+      tlsSocket.on('error', () => {});
+    });
+
+    const addressInfo = server.address() as net.AddressInfo;
+
+    const connection = new Connection({
+      server: addressInfo.address,
+      options: {
+        port: addressInfo.port,
+        encrypt: 'strict',
+        trustServerCertificate: true,
+        packetSize: 2048,
+        connectTimeout: 3000
+      }
+    });
+
+    // The fake server never completes PRELOGIN/LOGIN7, so the connection will time out
+    // eventually — that's fine, we only need to observe what happens right after the
+    // TLS handshake, which happens well before the timeout fires.
+    connection.connect(() => {});
+
+    setTimeout(() => {
+      setMaxSendFragmentSpy.restore();
+      connection.close();
+      for (const socket of serverSockets) {
+        socket.destroy();
+      }
+
+      assert.isTrue(
+        setMaxSendFragmentSpy.calledWith(2048),
+        `expected setMaxSendFragment to be called with 2048, got calls: ${JSON.stringify(setMaxSendFragmentSpy.args)}`
+      );
+
+      done();
+    }, 500);
   });
 });
--- a/test/unit/connection-test.ts
+++ b/test/unit/connection-test.ts
@@ -142,24 +142,41 @@ describe('Using `strict` encryption', function() {
       }
     });
 
-    // The fake server never completes PRELOGIN/LOGIN7, so the connection will time out
-    // eventually — that's fine, we only need to observe what happens right after the
-    // TLS handshake, which is what `setMaxSendFragment` is called for.
-    connection.connect(() => {});
+    let finished = false;
+    const finish = (err?: Error) => {
+      if (finished) {
+        return;
+      }
+      finished = true;
 
-    onCalled.then(() => {
       setMaxSendFragmentStub.restore();
       connection.close();
       for (const socket of serverSockets) {
         socket.destroy();
       }
 
-      assert.deepEqual(
-        calls, [2048],
-        `expected setMaxSendFragment to be called with 2048, got calls: ${JSON.stringify(calls)}`
-      );
+      done(err);
+    };
 
-      done();
+    // The fake server never completes PRELOGIN/LOGIN7, so the connection would time out
+    // eventually — that's fine, we only need to observe what happens right after the
+    // TLS handshake, which is what `setMaxSendFragment` is called for. If the handshake
+    // itself breaks, the connect callback reports the error, failing fast instead of
+    // hanging until the mocha timeout. (`finish` ignores the late timeout error that
+    // this callback reports after a successful run tears the connection down.)
+    connection.connect(finish);
+
+    onCalled.then(() => {
+      try {
+        assert.deepEqual(
+          calls, [2048],
+          `expected setMaxSendFragment to be called with 2048, got calls: ${JSON.stringify(calls)}`
+        );
+      } catch (err: any) {
+        return finish(err);
+      }
+
+      finish();
     });
   });
 });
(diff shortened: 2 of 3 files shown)


test/unit/connection-test.ts
```diff
@@ -74,4 +77,69 @@ describe('Using `strict` encryption', function() {
...
+    server.on('connection', (rawSocket) => {
+      serverSockets.push(rawSocket);
+
+      // A minimal TDS 8.0 "server": complete a real TLS handshake advertising the
+      // `tds/8.0` ALPN protocol (as `wrapWithTls` requires), then go silent. We only
+      // care about what the client does immediately after the handshake completes.
+      const tlsSocket = new tls.TLSSocket(rawSocket, {
+        isServer: true,
+        key: readFileSync('./test/fixtures/localhost.key'),
+        cert: readFileSync('./test/fixtures/localhost.crt'),
```
arthurschreiber: Running this test on a clean checkout fails deterministically, even with the fix applied:

```
AssertionError: expected setMaxSendFragment to be called with 2048, got calls: []
```

The client-side TLS handshake never completes, because `test/fixtures/localhost.crt` expired on 2022-08-05, and its SAN only covers `DNS:localhost` while the test connects to `127.0.0.1`. Since `secureConnect` never fires on the client, `setMaxSendFragment` is never reached and the spy stays empty.

Rather than regenerating the fixtures (which just re-arms the same time bomb), I'd suggest generating a throwaway CA + leaf certificate at runtime in the test, then:
- passing the CA to the connection via `cryptoCredentialsDetails: { ca: [...] }` — this flows into `tls.createSecureContext()` in `wrapWithTls`, so validation succeeds legitimately, and
- giving the leaf cert an `IP:127.0.0.1` SAN (or issuing it for `localhost` and setting `options.serverName: 'localhost'`), since `wrapWithTls` maps an IP `server` to an empty SNI servername.

---

test/unit/connection-test.ts
```diff
@@ -74,4 +77,69 @@ describe('Using `strict` encryption', function() {
...
+    });
+
+    const addressInfo = server.address() as net.AddressInfo;
+
+    const connection = new Connection({
+      server: addressInfo.address,
+      options: {
+        port: addressInfo.port,
+        encrypt: 'strict',
+        trustServerCertificate: true,
```
arthurschreiber: This option has no effect here and is misleading: under `encrypt: 'strict'`, `trustServerCertificate` is deliberately ignored — matching TDS 8.0 semantics across the Microsoft drivers (with `Encrypt=Strict`, `TrustServerCertificate` is treated as false and certificate validation is mandatory). `wrapWithTls` accordingly never passes `rejectUnauthorized` to `tls.connect`. Trust has to come from a real CA (see comment above), so this line should be dropped.

---

test/unit/connection-test.ts
```diff
@@ -74,4 +77,69 @@ describe('Using `strict` encryption', function() {
...
+        connectTimeout: 3000
+      }
+    });
+
+    // The fake server never completes PRELOGIN/LOGIN7, so the connection will time out
+    // eventually — that's fine, we only need to observe what happens right after the
+    // TLS handshake, which happens well before the timeout fires.
+    connection.connect(() => {});
+
+    setTimeout(() => {
```
arthurschreiber: Asserting after a fixed 500 ms wall-clock delay has two problems: it adds a flat 500 ms to every suite run, and it's racy on loaded CI hosts — if TCP connect + TLS handshake take longer than 500 ms, the assertion runs before `setMaxSendFragment` is called and fails spuriously.

Better to react to the handshake instead: listen for `secureConnect` on the server-side `tlsSocket` (plus a `setImmediate` so the client-side callback has run), or resolve when the spy is first called.

---

src/connection.ts
```diff
@@ -2237,7 +2237,22 @@ class Connection extends EventEmitter {
...
+          // same reason; `wrapWithTls` (TDS 8.0 / encrypt: "strict") must match it.
+          //
+          // Without this fix, a LOGIN7 spanning multiple TDS packets (e.g. any FedAuth
+          // login carrying an Entra ID access token, which routinely exceeds one 4KB
+          // packet) reliably fails with ECONNRESET against Azure SQL once TLS is
+          // negotiated but before LOGINACK — see
+          // https://github.com/tediousjs/tedious/issues/1182. Single-packet logins
+          // (e.g. plain SQL auth with short credentials) are unaffected either way,
+          // which is why this bug reads as "FedAuth-specific" until you look closer.
+          encryptsocket.setMaxSendFragment(this.config.options.packetSize);
```
arthurschreiber: Two things worth considering on the fix itself:

1. `setMaxSendFragment` only accepts 512–16384 and returns `false` (silently doing nothing) outside that range. tedious doesn't enforce an upper bound on `packetSize`, and values like `32767` are commonly recommended for bulk loads — exactly the configurations most likely to send large multi-packet messages, and exactly where this cap would silently not apply. Consider clamping: `setMaxSendFragment(Math.min(this.config.options.packetSize, 16384))`. (The same latent issue exists on the classic path at message-io.ts:107 — fine to fix separately.)

2. This caps the fragment size once, at handshake time. If the server later renegotiates the packet size down via ENVCHANGE during login, the strict path never re-applies the cap — `MessageIO#packetSize()` only re-caps through `this.securePair`, which is undefined for strict-mode connections. The classic path re-applies it on every packet-size change. Fine as a follow-up issue, but worth tracking so the same ECONNRESET class doesn't resurface post-login.

---

arthurschreiber (changes requested):
Thanks for digging into this — the production change is directionally right. `wrapWithTls` should cap the TLS max send fragment size the same way the classic `MessageIO#startTls` path does (message-io.ts:107), and the analysis of why this only bites multi-packet LOGIN7 messages (FedAuth tokens) is convincing.

However, I checked out the branch and ran the new test, and it fails even with the fix applied — the TLS handshake in the test never completes, so the spy records no calls. Details in the inline comments, along with two smaller points on the production change.

One pre-existing bug adjacent to the changed lines that's worth fixing while you're in here: in the `finally` block just below, the handler registered with `once('secureConnect', onConnect)` is removed with `removeListener('connect', onConnect)` — mismatched event name, so the listener is never actually detached on the error/abort path.

---

zinvob:
Applied this fix locally and also worked through the review feedback above.

 - Fragment size clamping: setMaxSendFragment silently no-ops outside 512–16384 bytes. Since packetSize is user-configurable (commonly 32767 for bulk operations), I clamped the value passed in — otherwise a large packetSize would leave the cap unset and reintroduce the exact ECONNRESET bug this PR fixes.
 - Renegotiation gap: wrapWithTls's cap was only applied once, at the initial handshake. If the server later shrinks the packet size via an ENVCHANGE token post-login, strict-mode connections never re-applied the cap (the classic path already handled this via securePair). Fixed by re-checking instanceof tls.TLSSocket in MessageIO#packetSize.
 - Listener cleanup bug: wrapWithTls's finally block called removeListener('connect', onConnect), but the handler was registered on 'secureConnect' — the mismatch meant it never actually detached on non-happy paths.
 - Test certificate: confirmed test/fixtures/localhost.crt is expired (2022) and lacks an IP SAN, and that wrapWithTls indeed ignores trustServerCertificate entirely (only the classic startTls path reads it). Reworked the test to use the existing test/fixtures/loopback-ip.crt fixture (valid, has an IP Address:127.0.0.1 SAN) and trust it via cryptoCredentialsDetails.ca, so the handshake succeeds through genuine cert validation.
 - Timing: replaced the fixed 500ms delay with reacting to the first setMaxSendFragment call instead.

Worth calling out for anyone else surprised the original test passed despite the expired cert: it only worked because NODE_TLS_REJECT_UNAUTHORIZED=0 was set in the test environment, which disables certificate validation globally and silently papered over both the expired cert and the ignored trustServerCertificate option. With that unset, the original test failed the handshake outright — which is how this got caught. Verified the new test both fails (times out) with the fix reverted and passes with it applied, in both cases with NODE_TLS_REJECT_UNAUTHORIZED unset.

arthurschreiber (approved):
Thanks for the thorough follow-up — this addresses everything from the previous review, and I've verified it locally: the regression test now passes (the 100-year loopback-ip fixture with IP SANs works nicely in place of runtime generation), the full unit suite is green, and lint/tsc are clean. The ENVCHANGE re-cap via the `instanceof tls.TLSSocket` check is exactly the right wiring for the strict path.

Two take-them-or-leave-them nits, neither blocking: `MAX_TLS_SEND_FRAGMENT_SIZE` is now defined in both `connection.ts` and `message-io.ts` and could be shared; and the test would fail faster/cleaner with a rejection path (e.g. also resolving/rejecting on a client-side `error`) instead of relying on the mocha timeout if the handshake ever regresses.

We'll leave this open for a few days in case you'd like to pick up those two nits; after that we'll go ahead and merge as is.

---

arthurschreiber:
Since maintainer edits aren't available to us on this branch, here are the two optional nits from the review as ready-made patches, validated locally on top of `823b788` (full unit suite passing, `eslint`/`tsc` clean). Apply with `git am` (save each block to a file, or pipe both in order), or ignore them — they remain non-blocking either way.

Note on the first patch: a plain named export from `message-io.ts` wouldn't work here, because the `module.exports = MessageIO` assignment at the bottom of that file clobbers named exports for CJS consumers — hence the static class property.

<details>
<summary><b>Patch 1/2</b> — refactor: share the TLS max send fragment size constant</summary>

```patch
From ee5112950dbde3a25ca5dccdb952b29ec8b7260b Mon Sep 17 00:00:00 2001
From: Claude <noreply@anthropic.com>
Date: Sat, 29 Aug 2026 10:20:11 +0000
Subject: [PATCH 1/2] refactor: share the TLS max send fragment size constant
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

MAX_TLS_SEND_FRAGMENT_SIZE was defined identically in connection.ts and
message-io.ts. Move it to a static property on MessageIO — a named export
would be clobbered by the module.exports assignment at the bottom of
message-io.ts for CJS consumers.

Co-Authored-By: Claude <noreply@anthropic.com>
---
 src/connection.ts |  7 +------
 src/message-io.ts | 12 +++++++-----
 2 files changed, 8 insertions(+), 11 deletions(-)

diff --git a/src/connection.ts b/src/connection.ts
index eb91f39..11bdc92 100644
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -172,11 +172,6 @@ const DEFAULT_CONNECT_RETRY_INTERVAL = 500;
  * @private
  */
 const DEFAULT_PACKET_SIZE = 4 * 1024;
-/**
- * Node's `tls.TLSSocket#setMaxSendFragment` silently ignores values outside this range.
- * @private
- */
-const MAX_TLS_SEND_FRAGMENT_SIZE = 16384;
 /**
  * @private
  */
@@ -2259,7 +2254,7 @@ class Connection extends EventEmitter {
           // `setMaxSendFragment` silently no-ops for values outside 512-16384, so a
           // `packetSize` above that (e.g. 32767, commonly used for bulk load) has to be
           // clamped or it would leave the cap unset and reintroduce the bug it fixes.
-          encryptsocket.setMaxSendFragment(Math.min(this.config.options.packetSize, MAX_TLS_SEND_FRAGMENT_SIZE));
+          encryptsocket.setMaxSendFragment(Math.min(this.config.options.packetSize, MessageIO.MAX_TLS_SEND_FRAGMENT_SIZE));
           resolve(encryptsocket);
         };
 
diff --git a/src/message-io.ts b/src/message-io.ts
index 2e462b5..a44414a 100644
--- a/src/message-io.ts
+++ b/src/message-io.ts
@@ -13,10 +13,12 @@ import { TYPE } from './packet';
 import IncomingMessageStream from './incoming-message-stream';
 import OutgoingMessageStream from './outgoing-message-stream';
 
-// Node's `tls.TLSSocket#setMaxSendFragment` silently ignores values outside this range.
-const MAX_TLS_SEND_FRAGMENT_SIZE = 16384;
-
 class MessageIO extends EventEmitter {
+  // Node's `tls.TLSSocket#setMaxSendFragment` silently ignores values outside this range.
+  // (A static class property rather than a named export, as the `module.exports`
+  // assignment at the bottom of this file would clobber named exports for CJS consumers.)
+  static readonly MAX_TLS_SEND_FRAGMENT_SIZE = 16384;
+
   declare socket: Socket;
   declare debug: Debug;
 
@@ -56,7 +58,7 @@ class MessageIO extends EventEmitter {
       this.outgoingMessageStream.packetSize = packetSize;
     }
 
-    const maxSendFragment = Math.min(this.outgoingMessageStream.packetSize, MAX_TLS_SEND_FRAGMENT_SIZE);
+    const maxSendFragment = Math.min(this.outgoingMessageStream.packetSize, MessageIO.MAX_TLS_SEND_FRAGMENT_SIZE);
 
     if (this.securePair) {
       // Classic `encrypt: true` path: TLS is layered over a `DuplexPair`.
@@ -115,7 +117,7 @@ class MessageIO extends EventEmitter {
 
         this.emit('secure', securePair.cleartext);
 
-        securePair.cleartext.setMaxSendFragment(Math.min(this.outgoingMessageStream.packetSize, MAX_TLS_SEND_FRAGMENT_SIZE));
+        securePair.cleartext.setMaxSendFragment(Math.min(this.outgoingMessageStream.packetSize, MessageIO.MAX_TLS_SEND_FRAGMENT_SIZE));
 
         this.outgoingMessageStream.unpipe(this.socket);
         this.socket.unpipe(this.incomingMessageStream);
-- 
2.43.0
```

</details>

<details>
<summary><b>Patch 2/2</b> — test: fail fast if the TLS handshake breaks in the fragment cap test</summary>

```patch
From ad3a15e2a0b1e9d82775827852fee7f02aaf2685 Mon Sep 17 00:00:00 2001
From: Claude <noreply@anthropic.com>
Date: Sat, 29 Aug 2026 10:20:11 +0000
Subject: [PATCH 2/2] test: fail fast if the TLS handshake breaks in the
 fragment cap test

Route the connect callback's error into the test result instead of
ignoring it: a broken handshake now fails at the connect timeout with the
underlying error, rather than hanging until the mocha timeout. Assertion
failures are reported through done(err) as well, instead of surfacing as
an unhandled rejection.

Co-Authored-By: Claude <noreply@anthropic.com>
---
 test/unit/connection-test.ts | 37 ++++++++++++++++++++++++++----------
 1 file changed, 27 insertions(+), 10 deletions(-)

diff --git a/test/unit/connection-test.ts b/test/unit/connection-test.ts
index 30ddcb6..03b01d1 100644
--- a/test/unit/connection-test.ts
+++ b/test/unit/connection-test.ts
@@ -142,24 +142,41 @@ describe('Using `strict` encryption', function() {
       }
     });
 
-    // The fake server never completes PRELOGIN/LOGIN7, so the connection will time out
-    // eventually — that's fine, we only need to observe what happens right after the
-    // TLS handshake, which is what `setMaxSendFragment` is called for.
-    connection.connect(() => {});
+    let finished = false;
+    const finish = (err?: Error) => {
+      if (finished) {
+        return;
+      }
+      finished = true;
 
-    onCalled.then(() => {
       setMaxSendFragmentStub.restore();
       connection.close();
       for (const socket of serverSockets) {
         socket.destroy();
       }
 
-      assert.deepEqual(
-        calls, [2048],
-        `expected setMaxSendFragment to be called with 2048, got calls: ${JSON.stringify(calls)}`
-      );
+      done(err);
+    };
 
-      done();
+    // The fake server never completes PRELOGIN/LOGIN7, so the connection would time out
+    // eventually — that's fine, we only need to observe what happens right after the
+    // TLS handshake, which is what `setMaxSendFragment` is called for. If the handshake
+    // itself breaks, the connect callback reports the error, failing fast instead of
+    // hanging until the mocha timeout. (`finish` ignores the late timeout error that
+    // this callback reports after a successful run tears the connection down.)
+    connection.connect(finish);
+
+    onCalled.then(() => {
+      try {
+        assert.deepEqual(
+          calls, [2048],
+          `expected setMaxSendFragment to be called with 2048, got calls: ${JSON.stringify(calls)}`
+        );
+      } catch (err: any) {
+        return finish(err);
+      }
+
+      finish();
    });
   });
 });
-- 
2.43.0
```

</details>

---

arthurschreiber:
Correction to the previous comment: GitHub's comment rendering mangled the inline patches (it stripped the author email from the `From:` lines, and a context line lost a leading space), so `git am` would reject them as posted. Please disregard them.

Instead, the same two commits are now available on the `claude/tedious-issue-1182-8b5q8y` branch in this repository, on top of this PR's `823b788`:

- `ee51129` — refactor: share the TLS max send fragment size constant
- `ad3a15e` — test: fail fast if the TLS handshake breaks in the fragment cap test

To pull them into this PR:

```bash
git remote add upstream https://github.com/tediousjs/tedious.git  # if not already added
git fetch upstream claude/tedious-issue-1182-8b5q8y
git cherry-pick ee5112950dbde3a25ca5dccdb952b29ec8b7260b ad3a15e2a0b1e9d82775827852fee7f02aaf2685
git push
```

They apply cleanly on the current PR head (no conflicts), and with them the full unit suite passes and `eslint`/`tsc` are clean. Still optional/non-blocking, as noted in the review.

---

zinvob:
Thanks. I added your adjustments. Always good with some cleanup.

PR #1775, opened 2026-09-02
fix: surface token parse failures as request errors

james-bg:
Split out of #1765 per the sequencing discussed there — this is the part accepted as a pure bugfix.

The token parser's internal stream (`Readable.from(StreamParser.parseTokens(...))`) has no `'error'` listener, so a parse failure inside a response surfaces as an unhandled `'error'` event and crashes the process instead of failing the request.

This PR re-emits stream errors from `Parser`, and routes them in `SentClientRequest` through the existing socket-error path: the active request fails with the parse error and the connection is closed — a parse failure leaves the connection at an undefined position in the TDS stream, so it cannot be recovered at the request level. The listener is removed at end-of-message; during a canceled response's drain it stays attached, so a parse failure there is still surfaced.

Complements #1759, which fixed the same unhandled-`'error'` hole for the login path.

Test: `connection-parse-error-test.ts` — a scripted server answers a SQL batch with an invalid token type (`0x00`). On current master the test dies with an uncaught `Unknown type: 0` stream error; with this fix the request errors and the process stays alive. Full unit suite and lint pass.

The inactivity-timer half of #1765 will follow separately, targeted at v21 on top of the `AbortSignal` work, per the maintainer's sequencing.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -1821,6 +1821,13 @@ class Connection extends EventEmitter {
     };
 
     this._onSocketError = (error) => {
+      // Nothing left to route once the connection is closed: `Final` has no
+      // `socketError` handler, and the error would only surface as noise
+      // after `close()`. (Mirrors `socketEnd()`.)
+      if (this.state === this.STATE.FINAL) {
+        return;
+      }
+
       this.dispatchEvent('socketError', error);
       process.nextTick(() => {
         this.emit('error', this.wrapSocketError(error));
@@ -3494,8 +3501,14 @@ class Connection extends EventEmitter {
 
       const handler = new Login7TokenHandler(this);
       const tokenStreamParser = this.createTokenStreamParser(message, handler);
+      // If the abort wins this race, the pending `once()` is left
+      // unobserved, and a parse error landing afterwards would reject it
+      // with nobody listening. Observe it so that cannot become an
+      // unhandled rejection (same idiom as `withAbortRace`).
+      const endOfMessage = once(tokenStreamParser, 'end');
+      endOfMessage.catch(() => {});
       await Promise.race([
-        once(tokenStreamParser, 'end'),
+        endOfMessage,
         signalAborted
       ]);
 
@@ -3524,8 +3537,14 @@ class Connection extends EventEmitter {
 
         const handler = new Login7TokenHandler(this);
         const tokenStreamParser = this.createTokenStreamParser(message, handler);
+        // If the abort wins this race, the pending `once()` is left
+        // unobserved, and a parse error landing afterwards would reject it
+        // with nobody listening. Observe it so that cannot become an
+        // unhandled rejection (same idiom as `withAbortRace`).
+        const endOfMessage = once(tokenStreamParser, 'end');
+        endOfMessage.catch(() => {});
         await Promise.race([
-          once(tokenStreamParser, 'end'),
+          endOfMessage,
           signalAborted
         ]);
 
@@ -3570,8 +3589,14 @@ class Connection extends EventEmitter {
 
       const handler = new Login7TokenHandler(this);
       const tokenStreamParser = this.createTokenStreamParser(message, handler);
+      // If the abort wins this race, the pending `once()` is left
+      // unobserved, and a parse error landing afterwards would reject it
+      // with nobody listening. Observe it so that cannot become an
+      // unhandled rejection (same idiom as `withAbortRace`).
+      const endOfMessage = once(tokenStreamParser, 'end');
+      endOfMessage.catch(() => {});
       await Promise.race([
-        once(tokenStreamParser, 'end'),
+        endOfMessage,
         signalAborted
       ]);
 
@@ -3668,8 +3693,14 @@ class Connection extends EventEmitter {
       ]);
 
       const tokenStreamParser = this.createTokenStreamParser(message, new InitialSqlTokenHandler(this));
+      // If the abort wins this race, the pending `once()` is left
+      // unobserved, and a parse error landing afterwards would reject it
+      // with nobody listening. Observe it so that cannot become an
+      // unhandled rejection (same idiom as `withAbortRace`).
+      const endOfMessage = once(tokenStreamParser, 'end');
+      endOfMessage.catch(() => {});
       await Promise.race([
-        once(tokenStreamParser, 'end'),
+        endOfMessage,
         signalAborted
       ]);
     });
@@ -3744,10 +3775,7 @@ Connection.prototype.STATE = {
         try {
           message = await this.messageIo.readMessage();
         } catch (err: any) {
-          this.dispatchEvent('socketError', err);
-          process.nextTick(() => {
-            this.emit('error', this.wrapSocketError(err));
-          });
+          this._onSocketError(err);
           return;
         }
         // request timer is stopped on first data package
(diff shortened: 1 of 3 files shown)


src/connection.ts
```diff
@@ -3755,6 +3755,19 @@ Connection.prototype.STATE = {
...
         const tokenStreamParser = this.createTokenStreamParser(message, new RequestTokenHandler(this, this.request!));
 
+        // A token parse failure leaves the connection at an undefined
+        // position in the TDS stream, so it cannot be recovered at the
+        // request level — treat it like a socket error. Without a listener
+        // here, a parse failure would surface as an unhandled `'error'`
+        // event on the parser's internal stream and crash the process.
+        const onParserError = (err: Error) => {
+          this.dispatchEvent('socketError', err);
+          process.nextTick(() => {
```
arthurschreiber: This dispatches `socketError` unconditionally, but the parser can still surface an error after the connection has left `SentClientRequest`. If user code calls `close()` from an `infoMessage` / `errorMessage` / `row` handler while malformed bytes are still buffered, the error arrives while `state` is already `Final`, and `Final` has no `socketError` handler. `dispatchEvent` then emits `Error("No event 'socketError' in state 'Final'")`, followed by the wrapped parse error, both after the connection was closed.

I reproduced this locally with the PR applied: a scripted server answers a batch with `[valid INFO token][0x00]`, and the client does `connection.on('infoMessage', () => connection.close())`. After `close()` the connection emits:

```
Error: No event 'socketError' in state 'Final'
ConnectionError: Connection lost - Unknown type: 0
```

Anyone who removes their `'error'` listener after `close()` gets an unhandled `'error'` event and a crash, which is the failure class this PR is meant to remove.

Suggested fix: guard the same way `socketEnd()` does:

```ts
const onParserError = (err: Error) => {
  if (this.state === this.STATE.FINAL) {
    return;
  }
  this.dispatchEvent('socketError', err);
  process.nextTick(() => {
    this.emit('error', this.wrapSocketError(err));
  });
};
```

Alternatively, remove the listener in the state's `exit` hook so it can't fire after the transition. As a smaller point, this closure is now the fourth copy of the `dispatchEvent('socketError')` + `nextTick` emit sequence in this file (socket `'error'` handler, `socketEnd()`, the `readMessage()` catch just above, and this). Reusing `this._onSocketError` and adding the `Final` guard there would keep the routing in one place.

---

src/token/token-stream-parser.ts
```diff
@@ -30,6 +30,10 @@ export class Parser extends EventEmitter {
     this.parser.on('end', () => {
       this.emit('end');
     });
+
+    this.parser.on('error', (error: Error) => {
```
arthurschreiber, line 34: Re-emitting here is the right call, and it means the login and initial-SQL paths now fail the connect callback instead of crashing, which is a nice side effect. One gap it opens: those paths wait with

```ts
await Promise.race([
  once(tokenStreamParser, 'end'),
  signalAborted
]);
```

When the abort wins (connect timeout, or `close()` during login), the `once()` promise is left unobserved. A parse error that lands afterwards now rejects it with nobody listening, so it becomes an `unhandledRejection`. Before this PR the same scenario crashed via the unhandled stream `'error'`, so severity is unchanged, but since the emit path is new it would be worth attaching a no-op `.catch(() => {})` (or a throwaway `'error'` listener) at those race sites so the abort path is fully safe.

---

src/connection.ts
```diff
@@ -3809,6 +3822,8 @@ Connection.prototype.STATE = {
         };
 
         const onEndOfMessage = () => {
+          tokenStreamParser.removeListener('error', onParserError);
```
arthurschreiber, line 3859: When the parser errors, `onEndOfMessage` never runs, so the request keeps its `'cancel'` (`_cancelAfterRequestSent`, `onCancel`), `'pause'` and `'resume'` listeners after the callback has already been invoked with the parse error. `Request.cancel()` only checks `canceled`, not completion, so a late `request.cancel()` (a user-side timeout racing the failure, for example) calls `_cancelAfterRequestSent`, which writes an ATTENTION packet to the destroyed socket, sets `attentionSent` and starts a cancel timer on a closed connection. The resulting socket error then hits the same `Final` state hole as above.

The mid-response socket error path has the same shape today, so this isn't strictly new, but this PR adds a more reachable route into it. Removing the request listeners in `onParserError` (or in the `socketError` handler's teardown) would close it.

---

arthurschreiber (changes requested):
Thanks for splitting this out. The core change is correct and the test covers it: with the `src` changes reverted the new test dies with an uncaught `Unknown type: 0`, and with them it passes. Full unit suite is green on my side too.

Three things I'd like addressed before merging, in the inline comments:

1. `onParserError` can fire after the connection is already in `Final` (reproduced) and then emits a spurious "No event 'socketError' in state 'Final'" plus a second error after `close()`. Needs a `Final` guard or listener removal on state exit.
2. The `once(tokenStreamParser, 'end')` / abort races in the login paths leave an unobserved promise that the new `'error'` emit can reject.
3. Request listeners are left attached on the failed request.

One more that is outside the diff but is the same bug one state over: `SentAttention` (around `connection.ts:3880`) does `await once(tokenStreamParser, 'end')` on its own parser, and the surrounding IIFE's `.catch` rethrows via `process.nextTick`. With `Parser` now re-emitting `'error'`, a malformed token in the attention-ack response rejects that `once()` and crashes the process, and the request callback is never invoked. Since this PR's stated goal is "surface token parse failures as request errors", I think that path should get the same `socketError` routing, either here or in a follow-up you're happy to own.

Not blocking: the copied `buildLoginAckToken` / `drainMessage` helpers in the test match what the sibling tests do, so no objection there.

---

james-bg:
Thanks for the thorough review — all three inline items and the `SentAttention` case are addressed in 7ad2b18.

- The `Final` guard now lives in `_onSocketError`, as you suggested, and the two adjacent `readMessage()` catches plus both parser-error handlers route through it, so the dispatch/emit sequence is in one place.
- `SentClientRequest` detaches the request's `cancel`/`pause`/`resume` listeners before failing it. Your read was exactly right: without that, a late `cancel()` armed a cancel timer on the closed connection and its expiry hit the same `Final` hole.
- The four login/initial-SQL race sites observe their `once()` promise (same idiom as `withAbortRace`). No test for this one — the abort tears the socket down before a late parse error can arrive, so I couldn't construct it deterministically; happy to add one if you see a way.
- `SentAttention` wraps its `once()` and routes a rejection the same way, so a malformed attention acknowledgement fails the request instead of crashing.

New tests cover the close-from-handler, late-cancel and malformed-ack cases; each fails on the previous commit in the way you described (the first reproduces your `No event 'socketError' in state 'Final'` verbatim).

arthurschreiber (approved):
Thanks, this addresses everything. I re-verified on `7ad2b18`:

- The three new tests each fail against the previous commit's `connection.ts` in the way described, and pass on this head.
- Lint and `tsc` are clean, the unit suite is green here, and it stays green on a trial merge with current master (the branch is behind #1773, but merges cleanly).
- I looked for a case where the `Final` guard in `_onSocketError` could hide a socket error that some path still needs and did not find one: every earlier state has its own `socketError` handler, and by `Final` the connection is already torn down.

On the untested abort-then-parse-error race: I agree it isn't deterministically constructible, since the abort destroys the socket before a late message can arrive. The no-op catch is the right defensive measure without a test.

LGTM.

---

